PR #3, opened 2025-06-26
update repo link

pablo-garay:


ko3n1g (approved):

chtruong814:
We'll also want to fix it here
https://github.com/NVIDIA-NeMo/Eval/blob/main/nemo_eval/package_info.py#L30

PR #7, opened 2025-06-27
Add PAT GHA CI

pablo-garay:


.github/actions/test-template/action.yml
```diff
@@ -54,6 +54,9 @@ inputs:
     description: "Has Azure credentials"
     required: false
     default: "false"
+  PAT:
+    description: "GitHub Personal Access Token"
+    required: true
```
chtruong814: I don't think you need this here for now

chtruong814: Instead, you'll want to add it to this template
https://github.com/NVIDIA-NeMo/Eval/blob/main/.github/workflows/_build_container.yml#L36

.github/workflows/cicd-main.yml
```diff
@@ -80,6 +81,7 @@ jobs:
           azure-client-id: ${{ secrets.AZURE_CLIENT_ID }}
           azure-tenant-id: ${{ secrets.AZURE_TENANT_ID }}
           azure-subscription-id: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
+          PAT: ${{ secrets.PAT }}
```
chtruong814: I don't think we need this here now

.github/workflows/cicd-main.yml
```diff
@@ -108,6 +110,7 @@ jobs:
...
```
chtruong814: I don't think we need this here now

.github/actions/test-template/action.yml
pablo-garay: I did based on what you originally suggested. After our conversation just now, updated based on new agreement

.github/workflows/cicd-main.yml
pablo-garay: I did based on what you originally suggested. After our conversation just now, updated based on new agreement

.github/workflows/cicd-main.yml
pablo-garay: I did based on what you originally suggested. After our conversation just now, updated based on new agreement

PR #11, opened 2025-07-04
Refactor deployment and add tests

marta-sd:


src/nemo_eval/api.py
```diff
@@ -30,10 +30,10 @@ def deploy(
     nemo_checkpoint: Optional[AnyPath] = None,
     serving_backend: str = "pytriton",
     model_name: str = "megatron_model",
-    server_port: int = 8000,
```
marta-sd, line 32: @athitten I have modified how args are called and used so that user will end up with the same url for querying regardless of the backend. Please let me know if you're ok with such change.

src/nemo_eval/api.py
```diff
@@ -231,8 +231,7 @@ def evaluate(
             f"as it is required to run {eval_cfg.type} evaluation"
         )
 
-    base_url, _ = target_cfg.api_endpoint.url.split("/v1")
-    server_ready = wait_for_fastapi_server(base_url=base_url, model_name=target_cfg.api_endpoint.model_id)
+    server_ready = check_endpoint(target_cfg.api_endpoint.url, "completions", target_cfg.api_endpoint.model_id)
```
marta-sd: @athitten I have modified the check if the server is ready to work both with triton and ray. Also, now it will work with any endpoint (ie ones not deployed with nemo).

src/nemo_eval/utils/ray_deploy.py
```diff
@@ -107,11 +105,6 @@ def deploy_with_ray(
         num_cpus=get_available_cpus(),
         num_gpus=total_gpus,
         include_dashboard=include_dashboard,
-        runtime_env={
```
marta-sd, line 109: @athitten @pthombre
Is it required to set visible devices here? I have removed it because it was very confusing for me - it took me couple of hours to debug why ray doesn't work for me (and it was because this var didn't match my setup).

I think it's less error-prone if we leave it to the user to set the visible devices. Let me know what you think.

src/nemo_eval/api.py
```diff
@@ -45,10 +45,9 @@ def deploy(
     enable_flash_decode: bool = True,
     enable_cuda_graphs: bool = True,
     # Ray deployment specific args
-    num_replicas: Optional[int] = None,
+    num_replicas: int = 1,
```
marta-sd, line 47: @athitten I have modified the default to match other parameters (`num_gpus=1` etc). Please let me know if it's ok

athitten: This looks okay to me, thanks

src/nemo_eval/utils/api.py
```diff
@@ -56,12 +51,6 @@ class ConfigParams(BaseModel):
     limit_samples: Optional[Union[int, float]] = Field(
         description="Limit evaluation to `limit` samples. Default: use all samples", default=None
     )
-    num_fewshot: Optional[int] = Field(
-        description="Number of examples in few-shot context. Default: None.",
-        default=None,
-        deprecated="This parameter is deprecated and not used for evaluations with NVIDIA Evals Factory. "
-        "In order to specify few-shot please use extra.num_fewshot.",
```
athitten, line 63: Do we have any examples in the docs that shows how to specify num_fewshot ? I think I have seen it, but just wanted to confirm

src/nemo_eval/api.py
athitten: Thank you @marta-sd , should we fetch `endpoint_type` from the url and not hardcode as `completions` in L234 ?

athitten: Also do you think its sufficient to just have the check_endpoint here and not do a health check ?

src/nemo_eval/utils/base.py
```diff
@@ -15,11 +15,58 @@
...
+            logger.info("Server is not ready")
+            time.sleep(retry_interval)
+    return False
+
+
+def check_endpoint(
+    endpoint_url: str, endpoint_type: str, model_name: str, max_retries: int = 600, retry_interval: int = 2
+) -> bool:
+    """
+    Check the health of the FastAPI server.
```
athitten: Minor comment on the docstring, want to add for both PyTriton(via FAstAPI) and Ray

src/nemo_eval/api.py
athitten: This is okay with me, as long as we clearly mention that in the args docstrings

src/nemo_eval/utils/ray_deploy.py
pthombre: I think this should be okay, as long as this environment variable is defined by the user. This parameter just defines the list of GPU Ids that should be made visible to your Ray cluster.

athitten: I think it will be an additional step for the user right to explicitly set export CUDA_VISIBLE_DEVICES and users might end up not seeing the speed up or face some issues if they miss to set this. Instead shall we just set it to the default for 1 gpu in deploy method like we have done for other args and mention in the docs to change it along with num replicas etc., if they need more replicas ?

athitten: I think the problem arose because the [default](https://github.com/NVIDIA-NeMo/Eval/blob/main/src/nemo_eval/api.py#L51) was empty here. Setting it to "0" along with num_replicas=1 should be fine I believe. Lmk what you think @marta-sd

src/nemo_eval/api.py
marta-sd: Good catch, endpoint type should be taken from the config. Let me fix that.

Regarding health check: I think it would be nice to have it, but currently triton and ray have different urls. So we either need to pass the health url to this function or modify the deployment code to unify them.

src/nemo_eval/utils/ray_deploy.py
marta-sd: Yes, the empty default was the root cause, but I found it confusing that we silently set the variable inside the function.

If I don't set it at all, the code just works fine for me (I guess all GPUs are visible?)
I think this is a good default unless the code won't work if e.g. the user have 8 GPUs available and requested 4 replicas. @pthombre / @athitten would such mismatch cause an error? In such case I agree that "0" is a better default.

src/nemo_eval/utils/api.py
marta-sd: I checked and we actually don't. Let me add it to the docs PR.

ko3n1g:
/ok to test [6b82035](https://github.com/NVIDIA-NeMo/Eval/pull/11/commits/6b82035df77977965ac96b07716556f31a229a40)

ko3n1g:
/ok to test [3aa1beb](https://github.com/NVIDIA-NeMo/Eval/pull/11/commits/3aa1bebd480ab19fe8b5a7e98c8c8fcbd1081bfc)

ko3n1g:
/ok to test [176c4c4](https://github.com/NVIDIA-NeMo/Eval/pull/11/commits/176c4c4eb10089c85fb9904e0afd7483822a1ce2)

ko3n1g:
/ok to test [39ec62f](https://github.com/NVIDIA-NeMo/Eval/pull/11/commits/39ec62f0f2b09e731d74122bc075a35fe2d4e4d3)

src/nemo_eval/api.py
athitten: Yeah health check is not the same url. Although we can do a health check on both for some num of retries and check, its not a clean solution. We can skip it for now and see if the need arises

ko3n1g:
/ok to test [e45c662](https://github.com/NVIDIA-NeMo/Eval/pull/11/commits/e45c66265122bc646de4cebff2f40b01e1cfd7a1)

src/nemo_eval/utils/ray_deploy.py
athitten: @marta-sd so I just tried these cases with your branch on my local WS that has 2 GPUs:
1) Set num_gpus=2, num_replicas=2, dont set any CUDA_VISIBLE_DEVICES. It used all GPUs available.
2) Set num_gpus=1, num_replicas=1, dont set any CUDA_VISIBLE_DEVICES. It just used GPU 0. 

So I think we should be safe to remove `cuda_visible_devices` from the `deploy_with_ray` API.
@pthombre please let us if you have any comments or if we have missed any cases ? Thanks!

marta-sd: Ok, thanks for checking, then I won't revert this change

src/nemo_eval/utils/base.py
marta-sd: Fixed in 8783acd

src/nemo_eval/api.py
marta-sd: Fixed in f3bd499

ko3n1g:
/ok to test [c689f9b](https://github.com/NVIDIA-NeMo/Eval/pull/11/commits/c689f9b1a1f2f0e78a68599e52a1359757ff8b12)

tests/functional_tests/test_deployment.py
```diff
@@ -79,13 +80,13 @@ def test_deployment(serving_backend, completions_request, logprobs_request, chat
...
+                /home/TestData/nemo2_ckpt/llama3-1b-lingua,
                 "--max_batch_size",
-                str(max_batch_size),
+                str(4),
                 "--port",
-                str(port),
+                str(8886),
                 "--serving_backend",
-                serving_backend,
+                pytriton,
```
athitten: Are these changes intentional @ko3n1g ? Or typo

ko3n1g:
/ok to test [60ac191](https://github.com/NVIDIA-NeMo/Eval/pull/11/commits/60ac191a5f30309744fa3de43255b1e3a28920f6)

PR #17, opened 2025-07-10
ci: Switch to dockerhub/cuda

ko3n1g:
Dockerhub images are cached by our infra, which makes network-based activity more reliable.

Since we cannot push to dockerhub, we're building the image for each test-job. While this sounds like an overhead, we're actually faster than before (on average) due to the pull-through cache of dockerhub.


docker/common/install.sh
```diff
@@ -1,9 +1,67 @@
 #!/bin/bash
```
chtruong814, line 15: Let's add the copyright here?

PR #18, opened 2025-07-10
ci: Add pip-install test

ko3n1g:
This adds a CI check that enforces that all modules are importable just with basic dependencies. Right now, TransformerEngine is the only dependency that is optional. If we were to require a TransformerEngine import in any module, after adding this CI check, we would need to guard that import with a `try/except` wrapper.

Example:

```python

try:
   import transformer_engine

   HAVE_TE = True

except ImportError:
   HAVE_TE = False

...

def my_super_duper_function_that_needs_te():
  if not HAVE_TE:
    raise ImportError("TransformerEngine is an optional dependency and not installed. Please install with `pip install nemo-eval[te]`")
```

This is just meant as an example, the actual error message can differ of course.

We recommend this so that users can navigate across the project without being forced to install everything. Dependencies that cannot be installed in non-CUDA environments are prohibitive for entry-level users.

### DIFF
--- a/.github/workflows/install-test.yml
+++ b/.github/workflows/install-test.yml
@@ -112,10 +112,6 @@ jobs:
             PIP_ARGS=(--no-build-isolation)
           fi
 
-          if [[ $EXTRA == *"trtllm"* ]]; then
-            apt-get install -y libopenmpi-dev
-          fi
-
           pip install ${PIP_ARGS[@]} .$EXTRA
 
           echo -e "machine github.com\n  login token\n  password ${{ secrets.PAT }}" > ~/.netrc


pyproject.toml
```diff
@@ -66,7 +66,12 @@ classifiers = [
     "Topic :: Software Development :: Libraries",
     "Topic :: Utilities",
 ]
-dependencies = ["nvidia-lm-eval", "uvicorn", "nemo-export-deploy"]
+dependencies = [
+    "nvidia-lm-eval",
+    "uvicorn",
+    "flask",
+    # "nemo-export-deploy"
```
marta-sd, line 73: What is our long-term plan for export-deploy? Do we want to have it in a basic version (i.e. without TE)?

.github/workflows/install-test.yml
```diff
@@ -0,0 +1,246 @@
...
+          pip install --upgrade pip
+
+          PIP_ARGS=()
+
+          if [[ $EXTRA == *"te"* ]]; then
+            pip install torch setuptools pybind11 wheel_stub
+            PIP_ARGS=(--no-build-isolation)
+          fi
+
+          if [[ $EXTRA == *"trtllm"* ]]; then
```
marta-sd: I think we don't need trtllm for now - we can't evaluate models deployed this way

pyproject.toml
athitten: Atleast for the container, export-deploy with just pytriton and ray (wo trtllm, vllm) would be installed for in-fw evaluation. Are we following the same here @ko3n1g  in pyproject.toml for basic dependencies

athitten: Clarification: Would guarding apply to trtllm, vllm too in the future when we enable evaluations with them ?

ko3n1g: nemo-export-deploy without TE/TRTLLM/vLLM should be a basic dependency imho since it's very lightweight to install and since we're heavily dependent on it.

TE/TRTLLM/vLLM we should track as optional dependencies and guard those if you ask me. We should also bump those dependencies independently of export-deploy for higher velocity and higher test coverage during dependency bump testing

PR #19, opened 2025-07-10
docs: Update contrib guide

ko3n1g:


CONTRIBUTING.md
```diff
@@ -1,5 +1,80 @@
 # Contributing To NeMo Eval
 
+Thanks for your interest in contributing to NeMo Eval!
+
+## 🛠️ Setting Up Your Environment
+
+### Local workstation
+
+NeMo Export-Deploy uses [uv](https://docs.astral.sh/uv/) for package management.
```
athitten: Its probably NeMo Eval instead of NeMo Export-Deploy ?

ko3n1g: fixed

CONTRIBUTING.md
```diff
@@ -1,5 +1,80 @@
...
+
+You can configure uv with the following commands:
+
+```bash
+uv sync --only-group build  # Installs build dependencies required by TransformerEngine
```
athitten, line 14: Clarification for my understanding, does it install only dependencies required by TE or all dependencies of NeMo-Eval ? I though this command in the next line `uv sync --extra te` installs TE and its dependencies.

ko3n1g: `uv sync --only-group build` installs basic dependencies + build dependencies of TE

the next command is then because of `--only-group build` able to install TE

PR #20, opened 2025-07-10
ci: Skip on docs only

ko3n1g:


.github/workflows/cicd-main.yml
```diff
@@ -29,9 +29,75 @@ permissions:
...
+
+          echo "main=$force_run_all" | tee -a "$GITHUB_OUTPUT"
+
   cicd-wait-in-queue:
     runs-on: ubuntu-latest
+    needs: [pre-flight]
     environment: test
+    if: |
+      !(needs.pre-flight.outputs.is_ci_workload == 'true'
+      || needs.pre-flight.outputs.docs_only == 'true')
```
thomasdhc, line 100: This should be false right or negation?

ko3n1g: It’s a negation

thomasdhc: Ah! my bad didn't read the brackets correctly

PR #25, opened 2025-07-13
Add nemo-run script

athitten:
- Adds nemo-run script for evaluation
- Updates the original [script](https://github.com/NVIDIA/NeMo/blob/main/scripts/llm/evaluation.py) to work with the modified deploy API with ray

### DIFF
--- a/scripts/evaluation_with_nemo_run.py
+++ b/scripts/evaluation_with_nemo_run.py
@@ -1,4 +1,3 @@
-
 # Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -26,7 +25,6 @@ import nemo_run as run
 
 from nemo_eval.api import deploy, evaluate
 from nemo_eval.utils.api import ApiEndpoint, ConfigParams, EvaluationConfig, EvaluationTarget
-
 
 ENDPOINT_TYPES = {"chat": "chat/completions/", "completions": "completions/"}
 


scripts/evaluation_with_nemo_run.py
```diff
@@ -0,0 +1,322 @@
...
+
+import nemo_run as run
+
+from nemo_eval.api import deploy, evaluate
+from nemo_eval.utils.api import ApiEndpoint, ConfigParams, EvaluationConfig, EvaluationTarget
+
+
+ENDPOINT_TYPES = {"chat": "chat/completions/", "completions": "completions/"}
+
+COMPLETIONS_TASKS = (
```
marta-sd: I think we should remove the list of benchmarks now. First, we enabled log-probs ones, so we don't need to restrict users. Also, I expect this list to grow with every release and it will become difficult to maintain.

What do you think?

scripts/evaluation_with_nemo_run.py
```diff
@@ -0,0 +1,322 @@
...
+        type=int,
+        default=2,
+        help="Batch size for deployment and evaluation",
+    )
+    parser.add_argument(
+        "--eval_task",
+        type=str,
+        default="lm-evaluation-harness.mmlu",
+        help="Evaluation benchmark to run.",
+        choices=EVAL_TASKS,
```
marta-sd: Remember to remove this line if we decide to remove `EVAL_TASKS` var

scripts/evaluation_with_nemo_run.py
```diff
@@ -0,0 +1,322 @@
...
+        action="store_true",
+        help="Run on slurm using run.SlurmExecutor",
+        default=False,
+    )
+    parser.add_argument('--nodes', type=int, default=1, help="Num nodes for the executor")
+    parser.add_argument('--devices', type=int, default=8, help="Num devices per node for the executor")
+    parser.add_argument(
+        '--container_image',
+        type=str,
+        default="nvcr.io/nvidia/nemo:dev",
```
marta-sd: I'm not familiar with this tag, so double checking: is this a good default?

scripts/evaluation_with_nemo_run.py
athitten: Yes makes sense, we can have the users refer to the docs page to see the tasks available. Let me remove it

scripts/evaluation_with_nemo_run.py
athitten: So every week we release dev containers but they might not be fully tested yet until the final release. I think it safe to add the latest release container here. Let me add 25.07 here, thanks for pointing out.

scripts/evaluation_with_nemo_run.py
athitten: Addressed in 62d2914ba14cfe1ffd6d57e64668a5cfba9f7f2e

scripts/evaluation_with_nemo_run.py
athitten: Removed 62d2914ba14cfe1ffd6d57e64668a5cfba9f7f2e

scripts/evaluation_with_nemo_run.py
athitten: added in a7fd6f7ec7064eb39ee514fd12bfda80375711b3

PR #30, opened 2025-07-17
build: Align release dependencies

ko3n1g:


pyproject.toml
```diff
@@ -101,10 +104,9 @@ nemo-export-deploy-pre = ["nemo-export-deploy"]
 
 [tool.uv.sources]
 # Remove this
-megatron-bridge = { git = "https://github.com/NVIDIA-NeMo/Megatron-Bridge.git", rev = "958fb9b7da4eb1506142a122965fa5424cfdc8a2" }
+megatron-bridge = { git = "https://github.com/NVIDIA-NeMo/Megatron-Hub.git", rev = "9e8c78c64f0396cf921c9d10d0f87517fa51566d" }
```
marta-sd: I think it should be `https://github.com/NVIDIA-NeMo/Megatron-Bridge.git`

pyproject.toml
```diff
@@ -70,11 +70,14 @@ dependencies = [
     "nvidia-lm-eval",
     "uvicorn",
     "flask",
-    # "nemo-export-deploy"
+    "megatron-core>=0.13.0a0,<0.14.0",
+    "nvidia-modelopt[torch,onnx]>=0.31.0a0,<0.32.0; sys_platform != 'darwin'",
```
marta-sd, line 74: why do we need these two:
* `nvidia-modelopt[torch,onnx]`
* `nvidia-resiliency-ext`
? Are they coming from export-deploy?

ko3n1g: NVRx comes from Mhub/Mcore and modelopt from export-deploy. Since we're installing those anyway, I'd like to track them as direct dependencies so that we can track dependency bumps and breakages faster

PR #33, opened 2025-07-22
Revise README

athitten:


README.md
```diff
@@ -94,20 +95,20 @@ print(results)
...
 
 ## 🏗️ Architecture
 
 ### Core Components
 
 #### 1. Deployment Layer
 
-- **PyTriton Backend**: High-performance inference using NVIDIA Triton Inference Server and OpenAI API compatibility via FastAPI Interface with model parallelism across single and multi node. Does not support multi instance evaluation.
-- **Ray Backend**: Single node model parallel multi instance evaluation using Ray Serve with OpenAI API compatibility. Multi node support coming soon.
+- **PyTriton Backend**: Delivers high-performance inference via NVIDIA Triton Inference Server, with OpenAI API compatibility through a FastAPI interface. Supports model parallelism across both single- and multi-node setups. Note: Multi-instance evaluation is not supported.
```
athitten, line 105: @jgerh do we need hyphen after single in `single-` ?

jgerh: It is correct. We're using a suspended compound adjective, where “single-” and “multi-” both modify “node.” The hyphen after "single" is needed to avoid confusion and make it clear that "node" is shared by both.

athitten: Got it, thanks for clarifying

PR #63, opened 2025-08-12
Pin nvidia-lm-eval version

athitten:
Pins nvidia-lm-eval to 25.6.1 as the latest 25.7.1 version leads to error. 

`ModuleNotFoundError: No module named 'core_evals.lm_evaluation_harness.input'`

We should pin the version until 25.7.1 nvidia-lm-eval is fixed by the CompEval team.
cc: @marta-sd 

Also this is an issue with latest 25.07 nemo fw container since it has nvidia-lm-eval 25.7.1. So all future nemo fw container builds should use nvidia-lm-eval 25.6.1 until latest nvidia-lm-eval is fixed

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -67,7 +67,7 @@ classifiers = [
     "Topic :: Utilities",
 ]
 dependencies = [
-    "nvidia-lm-eval",
+    "nvidia-lm-eval==25.6.1",
     "uvicorn",
     "flask",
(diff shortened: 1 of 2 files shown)


ko3n1g (commented):
Thanks for adding this. Please also update the lock file by running `uv lock` in the workdir

athitten:
> Thanks for adding this. Please also update the lock file by running `uv lock` in the workdir

Thank you @ko3n1g added in the latest commit.

PR #126, opened 2025-09-16
docs(README): remove misleading links and statements

piojanu:


piojanu:
This README is outdated and now it points to the new docs.

PR #140, opened 2025-09-16
feat: Enable auto-export for Slurm runs submitted from a local machine

ka00ri:
- Enables auto-export for Slurm runs submitted from a local machine, using `export_config.yml` under `artifacts/` alongside `run_config.yml`.
- Updates docs on triggering auto-export by ensuring the launcher binary is on PATH via `execution.env_vars.evaluation.PATH`.
- Interim approach to remote auto-export: artifacts-based with low boilerplate; keeps run config separate from export configurations; doesn’t require `execDB` on remote machine.


agronskiy (approved):

agronskiy (approved):
LG, thanks a ton Meriem!

PR #193, opened 2025-09-18
fix(config): default schema + relaxing overrides in config

agronskiy:
- adjusts the `default` schema
- fixes the ipynb example
- adds the possibility for `dict_overrides` in API to _extend_ the config by temporarily dropping `struct` mode


packages/nemo-evaluator-launcher/examples/notebooks/nemo-evaluator-launcher-api.ipynb
```diff
@@ -16,7 +16,16 @@
...
-   "outputs": [],
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "/home/agronskiy/code/nvidia-nemo-eval/packages/nemo-evaluator-launcher/.venv/lib/python3.12/site-packages/tqdm/auto.py:21: TqdmWarning: IProgress not found. Please update jupyter and ipywidgets. See https://ipywidgets.readthedocs.io/en/stable/user_install.html\n",
+      "  from .autonotebook import tqdm as notebook_tqdm\n"
+     ]
+    }
```
AWarno: question: is it ok? probably yes

agronskiy: Thank, this is benign but I 'll drop that output though, thanks

PR #195, opened 2025-09-22
feat(lazy-loading): Use lazy import for Lepton to speed up the CLI

fgalko-oss:


agronskiy:
I assume a network flake in CI

PR #220, opened 2025-09-23
Add output_dir param to EvaluationConfig in docs

athitten:
Add output_dir param to EvaluationConfig in docs


marta-sd (approved):

marta-sd (changes requested):
I think we need the same fix in `docs/evaluation-with-ray.md`

athitten:
> I think we need the same fix in `docs/evaluation-with-ray.md`

Thanks, added

PR #229, opened 2025-09-25
fix(executors): migrate to eval-factory cmd

piojanu:
It unblocks us to use new Eval Factory containers in the launcher — they don't have `nv-eval`/`nv_eval` alias anymore.


packages/nemo-evaluator-launcher/tests/unit_tests/test_slurm_executor.py
```diff
@@ -105,9 +105,7 @@ def mock_dependencies(self):
             }
             mock_get_health.return_value = "http://localhost:8000/health"
             mock_get_endpoint.return_value = "http://localhost:8000/v1"
-            mock_get_eval_command.return_value = (
-                "nemo_evaluator_launcher run_eval --test"
```
agronskiy, line 109: I wonder how that worked before lol

PR #248, opened 2025-10-01
bug(docs): fix self-service docs

marta-sd:


prokotg (approved):
LGTM, thank you for updating!

PR #260, opened 2025-10-02
fix(local-exec): avoid creation of exec id for dry-run

agronskiy:
* this avoids creating job ids and invocation ids for eval db when dry-run is requested
* fixes number of bad or useless tests


packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/execdb.py
```diff
@@ -148,41 +148,135 @@ def write_job(self, job: JobData) -> None:
             )
             raise
 
+    def _resolve_invocation_id(self, short_id: str) -> Optional[str]:
```
fgalko-oss, line 151: looks great, could you please add some tests for resolving prefixes for invocations/jobs?

PR #264, opened 2025-10-03
fix(cli): change entrypoint name from eval-factory to nemo-evaluator

prokotg:
This MR changes `nemo-evaluator` entrypoint from `eval-factory` to `nemo-evaluator`. 

To facilitate backward compatibility we take the following measures:
1. We keep `eval-factory` entrypoint with deprecation warning in `nemo-evaluator`
2. We use either entrypoint in `nemo-evaluator-launcher`, with preference (first-matching) for `nemo-evaluator`


prokotg:
/ok to test [d1dba13](https://github.com/NVIDIA-NeMo/Evaluator/pull/264/commits/d1dba13baf5e172a89acc258284cb9bf1fafe214)

wprazuch (approved):
LGTM!

PR #266, opened 2025-10-03
Fix for Ray deployment in nemo-run script

athitten:


scripts/evaluation_with_nemo_run.py
```diff
@@ -41,8 +41,8 @@
   --server_port {server_port} \
   --num_gpus {devices} \
   --num_nodes {nodes} \
-  --tensor_parallelism_size {tensor_model_parallel_size} \
-  --pipeline_parallelism_size {pipeline_model_parallel_size} \
+  --tensor_model_parallel_size {tensor_model_parallel_size} \
+  --pipeline_model_parallel_size {pipeline_model_parallel_size} \
```
athitten, line 45: Use the same name between Pytriton and Ray script for TP, PP. PR in ED for this: [https://github.com/NVIDIA-NeMo/Export-Deploy/pull/501](https://github.com/NVIDIA-NeMo/Export-Deploy/pull/501)

scripts/evaluation_with_nemo_run.py
```diff
@@ -288,8 +289,8 @@ def main():
         "max_input_len": args.max_input_len,
         "tensor_model_parallel_size": args.tensor_parallelism_size,
         "pipeline_model_parallel_size": args.pipeline_parallelism_size,
-        "max_batch_size": args.batch_size,
-        "devices": args.devices,
+        "max_batch_size": args.batch_size, #TODO check in llama 405B run
+        "num_gpus": args.devices if args.serving_backend == "pytriton" else args.devices * args.nodes,
```
athitten: Ray requires `num_gpus` to be total num of gpus in case of multi node.

scripts/evaluation_with_nemo_run.py
```diff
@@ -308,17 +311,23 @@ def main():
...
             **commons_args, additional_args=additional_args
         )
+        deploy_run_script = run.Script(
+            inline=deploy_script, metadata={"use_with_ray_cluster": True}
+        )
     else:
         raise ValueError(f"Invalid serving backend: {args.serving_backend}")
+
     print(deploy_script)
-    # [snippet-config-start]
```
marta-sd, line 320: This comment is used in docs to load the snippet, we need to add it back. However, we now have the `deploy_run_script` created inside if-else, so I'm not sure where it should be placed

scripts/evaluation_with_nemo_run.py
```diff
@@ -293,7 +294,9 @@ def main():
         "tensor_model_parallel_size": args.tensor_parallelism_size,
         "pipeline_model_parallel_size": args.pipeline_parallelism_size,
         "max_batch_size": args.batch_size,
-        "devices": args.devices,
+        "devices": args.devices
```
marta-sd, line 297: Do you think it's feasible to unify this arg between the two scripts? I think it might be confusing for the users.

Also adding a note to myself to make sure it's well documented in #432

scripts/evaluation_with_nemo_run.py
athitten: Oh my bad, let me add it back.

scripts/evaluation_with_nemo_run.py
athitten: @marta-sd both the scripts use `num_gpus` itself for devices. Do you mean you don't want to have devices as `args.devices` in case of pytriton and `args.devices * args.nodes` in case of Ray ? Lmk, thanks!

scripts/evaluation_with_nemo_run.py
athitten: As of now, I have added it back to where it was.

scripts/evaluation_with_nemo_run.py
marta-sd: Sorry, let me clarify: I think the arguments in the nemo-run script are defined in the optimal way.

I was thinking about unifying the meaning of "devices" argument in the `deploy_ray_inframework.py` and `deploy_inframework_triton.py` scripts - until reviewing this PR I didn't realize that ray requires the total number of GPUs and PyTriton a per-node number

athitten: Okay, yeah it can be changed we should add `--num_nodes` as well to `deploy_ray_inframework.py` and compute total GPUs inside the script and allow only GPUs per node for `--num_gpus` arg. Also this might diverge from other scripts like `deploy_ray_hf` and `deploy_ray_trtllm` and I haven't verified large scale runs with it. Is it okay to unify them a in a follow-up when I can make sure this holds good for `deploy_ray_hf` and `deploy_ray_trtllm` as well ? Its on my list try these.

marta-sd: Sounds good, thanks!

PR #275, opened 2025-10-06
fix(launcher) Set `--network=host` for the local executor and fix eval factory command

marta-sd:


marta-sd:
/ok to test [8065e9d](https://github.com/NVIDIA-NeMo/Evaluator/pull/275/commits/8065e9d41a2bb7a2f0bc6820f72c6d3c02643769)

PR #277, opened 2025-10-06
feat(core): Multiple endpoint types on benchmark and endpoint end

prokotg:
This MR introduces multiple type definition on both benchmark and endpoint ends. With these changes, we can define a list of required types on benchmark end. For example: `[chat, vlm]`. On the endpoint side, we define capabilites which must be a superset of required typed. More information is provided in the documentation

### DIFF
--- a/packages/nemo-evaluator/src/nemo_evaluator/core/input.py
+++ b/packages/nemo-evaluator/src/nemo_evaluator/core/input.py
@@ -401,7 +401,11 @@ def check_type_compatibility(evaluation: Evaluation):
                 f"The benchmark '{evaluation.config.type}' does not support any of the model types '{evaluation.target.api_endpoint.type}'. \n"
                 f"The benchmark supports: {_get_benchmark_types_readable(evaluation.config.supported_endpoint_types)}"
             )
-
+        # unpack types back. Listifying was for checking types only
+        evaluation.config.supported_endpoint_types = (
+            evaluation.config.supported_endpoint_types[0]
+        )
+        evaluation.target.api_endpoint.type = evaluation.target.api_endpoint.type[0]
     if evaluation.target.api_endpoint.type:
         # Check this only if the model is really required (to accomodate for non-model evals)
         if evaluation.target.api_endpoint.url is None:


prokotg:
/ok to test [cc7936d](https://github.com/NVIDIA-NeMo/Evaluator/pull/277/commits/cc7936d2441319dab13246379d0d5e5ad28b2c43)

packages/nemo-evaluator/src/nemo_evaluator/core/input.py
```diff
@@ -348,20 +348,38 @@ def get_evaluation(
...
+    # Model endpoint types must be checked against benchmark required capabilities.
+    # All benchmark required capabilities must be present in model endpoint types.
+
+    # If the evaluation does not specify particular endpoint types,
+    # we treat it as 'any's
+
+    # We have to be carefull in terms of types. We might run into turning a stringable
+    # dataclass into a set
+    if evaluation.config.supported_endpoint_types is not None:
+        if not isinstance(evaluation.target.api_endpoint.type, list):
```
wprazuch, line 378: Wondering about the check in line 376. What if type is actually `None`?
Would that not lead to `[None]` being set here?

packages/nemo-evaluator/src/nemo_evaluator/core/input.py
```diff
@@ -348,20 +348,38 @@ def get_evaluation(
...
+                evaluation.config.supported_endpoint_types
+            ]
+        model_types = set(evaluation.target.api_endpoint.type)
+        is_target_compatible = False
+        for benchmark_type_combination in evaluation.config.supported_endpoint_types:
+            if not isinstance(benchmark_type_combination, list):
+                benchmark_type_combination = [benchmark_type_combination]
+
+            if model_types.issuperset(set(benchmark_type_combination)):
+                is_target_compatible = True
```
wprazuch: We only keep information about last successful check in this variable - is this not a potential issue? To be completely honest with you, I have a hard time understanding this part, could you elaborate on that? :D

packages/nemo-evaluator/src/nemo_evaluator/core/input.py
```diff
@@ -348,20 +348,38 @@ def get_evaluation(
...
+
         if evaluation.target.api_endpoint.type is None:
```
wprazuch: If I got this right, this will never be true? Since it's going to be wrapped in a list, if it's None, in line 359

packages/nemo-evaluator/src/nemo_evaluator/core/input.py
prokotg: Benchmark requirements are strict. From a benchmark perspective, you define a list of sets that are accepted. For example, let's image that your benchmark is a VLM benchmark but accepts both base and chat models. Then, on benchmark side, you would define `evaluation.config.supported_endpoint_types` as  a list of :

```yaml
config:
  supported_endpoint_types:
    - [vlm, chat]
    - [vlm, completions]
```
This means, that your model must have all capabilities in at least one of these sets (those sets are implemented as lists but that does not matter here IMO)

So if we define on model endpoint capabilities it must be a superset of at least one element from supported_endpoint_types list. 

Let's have a look at examples.

1. Model `[chat]` - netiher a superset of `[vlm, chat]` nor `[vlm, completions]`
2. Model `[chat, vlm]` - superset of `[vlm, chat]` but not `[vlm, completions]`
3. Model `[chat, vlm, completions]` - superset of both `[vlm, chat]` and `[vlm, completions]`
4. Model `[chat, vlm, completions, audio]` -  superset of both `[vlm, chat]` and `[vlm, completions]`

It does not matter which one is accepted - it only matters that at least one is compatible. So we keep the last one

prokotg: But I do see a crash where we have multiple combinations compatible - which one is passed down to the evaluation harness?

packages/nemo-evaluator/src/nemo_evaluator/core/input.py
prokotg: you're right, thanks!

packages/nemo-evaluator/src/nemo_evaluator/core/input.py
prokotg: I'm raising an error on multiple compatible

packages/nemo-evaluator/src/nemo_evaluator/core/input.py
prokotg: fixed by not going into this case where we have a "None" type on endpoint side. Thanks!

prokotg:
/ok to test [f781e54](https://github.com/NVIDIA-NeMo/Evaluator/pull/277/commits/f781e54d7d2294c5c7c174a39941f8831641bd19)

prokotg:
Closing as new ideas for model types - tags emerges

PR #282, opened 2025-10-07
feat(exporters): add config overrides, fix auto-export,  fix logged configs, adapts tests and docs

ka00ri:
PR includes the following changes:

- Support CLI overrides for exporters: -o export.<dest>.key=value
- Add --output-dir/-out flag for LocalExporter, instead of previous -o now reserved for overrides
- Fix Slurm auto-export: expand env var placeholders at runtime on host machine avoids no secrets in logs
- Move export configs to top-level export.<dest> (see examples/local_auto_export_llama_3_1_8b_instruct.yaml)
- Separate export env_vars from evaluation
- Fix logging user config (config.yml) to MLflow/W&B instead of run_config
- Adapt tests to new config upload and artifact path behavior
- Support logging run logs along artifacts (results) and user config to MLFlow/WandB
- Update exporters documentation

### DIFF
--- a/docs/nemo-evaluator-launcher/configuration/execution/index.md
+++ b/docs/nemo-evaluator-launcher/configuration/execution/index.md
@@ -15,12 +15,11 @@ execution:
   type: local  # or slurm, lepton
   output_dir: /path/to/results  # Common: where to save results (mounted as /results in evaluation container, must be creatable on the given executor)
   auto_export:                   # Common: automatic result export
-    destinations: ["wandb", "mlflow", "gsheets", "s3"]
+    destinations: ["wandb", "mlflow", "gsheets"]
     configs:
       wandb: {}
       mlflow: {}
       gsheets: {}
-      s3: {}
   # Executor-specific parameters (see individual executor docs):
 ```
 
@@ -31,7 +30,6 @@ On the executor level, we can define results export methods. The `auto_export` f
 - **W&B**: Weights & Biases experiment tracking
 - **MLFlow**: ML experiment management
 - **GSheets**: Google Sheets integration
-- **S3**: AWS S3 storage
 
 For detailed configuration options, see the {doc}`Exporters Documentation <../../exporters/overview>`.
 
--- a/docs/nemo-evaluator-launcher/configuration/execution/local.md
+++ b/docs/nemo-evaluator-launcher/configuration/execution/local.md
@@ -9,6 +9,7 @@ See the complete configuration structure in the [Local Config File](../../../../
 ## Key Settings
 
 - **`output_dir`**: Directory where evaluation results will be saved (required)
+- **`extra_docker_args`**: Additional arguments to pass to the `docker run` command (optional). This flag allows advanced users to customize their setup (see [Advanced configuration](#advanced-configuration)).
 
 Tips:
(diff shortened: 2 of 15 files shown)


packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/cli/export.py
```diff
@@ -67,20 +75,29 @@ class ExportCmd:
...
+
         from nemo_evaluator_launcher.api.functional import export_results
 
+        # Validation: ensure IDs are provided
+        if not self.invocation_ids:
+            print("Error: No IDs provided. Specify one or more invocation or job IDs.")
+            print(
+                "Usage: nemo-evaluator-launcher export <id> [<id>...] --dest <destination>"
+            )
+            return
```
AWarno, line 97: Do we want to print it or raise Error?

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/cli/export.py
```diff
@@ -95,20 +112,90 @@ def execute(self) -> None:
...
         )
 
         result = export_results(self.invocation_ids, self.dest, config)
 
-        if not result["success"]:
-            print(f"Export failed: {result.get('error', 'Unknown error')}")
+        if not result.get("success", False):
+            err = result.get("error", "Unknown error")
+            print(f"\nExport failed: {err}")
+            # Provide actionable guidance for common configuration issues
```
AWarno, line 176: maybe error should be raised?

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/functional.py
```diff
@@ -456,22 +456,47 @@ def export_results(
...
+                                "storage_type": "remote_local",
+                            },
                             config=cfg_yaml,
                         )
+                        # DEBUG: print what we loaded
+                        print(f"DEBUG: cfg_yaml keys: {list(cfg_yaml.keys())}")
+                        if "evaluation" in cfg_yaml:
+                            print(
+                                f"DEBUG: evaluation.tasks: {cfg_yaml.get('evaluation', {}).get('tasks')}"
+                            )
```
AWarno, line 499: maybe [our logger](https://github.com/NVIDIA-NeMo/Evaluator/blob/9367ef0026ec1e00ca8ff71b09acc6e46c1fe1dc/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/logging_utils.py)  for the debug?

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/cli/export.py
ka00ri: This is to not block overall workflow and users can always retry the export. Exporting failure shouldn't affect overall job/evaluation status.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/cli/export.py
ka00ri: I am going through code and taking notes for the error-handling unification task.

PR #291, opened 2025-10-09
fix(missing-env-var): fix missing env var

AWarno:
The gated GPQA now uses ifeval (previously there was a missing environment variable error). We could fix it by adding this environment variable, but we already have a GPQA example in other places.

### DIFF
--- a/packages/nemo-evaluator-launcher/examples/local_auto_export_llama_3_1_8b_instruct.yaml
+++ b/packages/nemo-evaluator-launcher/examples/local_auto_export_llama_3_1_8b_instruct.yaml
@@ -92,4 +92,4 @@ deployment:
 
 evaluation:
   tasks:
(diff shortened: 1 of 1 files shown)


packages/nemo-evaluator-launcher/examples/local_auto_export_llama_3_1_8b_instruct.yaml
```diff
@@ -92,4 +92,4 @@ deployment:
 
 evaluation:
   tasks:
-    - name: simple_evals.gpqa_diamond
+    - name: ifeval
```
ka00ri, line 95: ifeval does't follow the harness.task format?

AWarno: no, it is unique

PR #300, opened 2025-10-09
feat(interceptors): remove params from payload recursively

piojanu:


packages/nemo-evaluator/src/nemo_evaluator/adapters/interceptors/payload_modifier_interceptor.py
```diff
@@ -109,10 +109,29 @@ def intercept_request(
         new_data = original_data.copy()
 
         # Remove specified parameters
+        def _remove_param_recursively(data: Any, param: str, _parent_key: str = ""):
```
prokotg, line 112: Overall looks good, but would be great to have a unit test covering the use case you have in mind

piojanu: Good point, DONE.

PR #307, opened 2025-10-13
feat: Import-checker

pablo-garay:


.github/workflows/install-test.yml
```diff
@@ -18,82 +18,255 @@
...
+      - name: Verify installation
+        run: |
+          python -c "import nemo_evaluator_launcher; print(f'nemo-evaluator-launcher version: {nemo_evaluator_launcher.__version__}')"
+          which nemo-evaluator-launcher
+          which nv-eval
+          nemo-evaluator-launcher --help
+
+      - name: Run unit tests
+        working-directory: packages/nemo-evaluator-launcher
+        run: |
```
ko3n1g: We discussed offline which tests we would run here. Unit and functional tests are not required here. Please update the workflows to make sure the correct install checkers are ran.

.github/workflows/install-test.yml
```diff
@@ -18,82 +18,255 @@
...
-    container:
-      image: ubuntu:24.04
-    environment: nemo-ci
+    needs: [test-nemo-evaluator, test-nemo-evaluator-launcher]
     strategy:
-      fail-fast: false
       matrix:
-        python-version: ["3.10", "3.11", "3.12"]
-        package: ["nemo-evaluator"]
+        python-version: ["3.10", "3.13"]
```
agronskiy: `question`: we're sure that .11 and .12 are not catching anything, e.g. in the locked deps?

.github/workflows/install-test.yml
```diff
@@ -18,82 +18,255 @@
...
         with:
           repository: NVIDIA-NeMo/FW-CI-templates
-          ref: v0.39.0
-          path: FW-CI-templates
+          path: ./FW-CI-templates
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.10"
```
agronskiy: `question`: for my understanding, here we only check the 3.10 right?

.github/workflows/install-test.yml
```diff
@@ -18,82 +18,255 @@
...
         uses: ./FW-CI-templates/.github/actions/check-imports
         with:
-          package-name: nemo_evaluator
-          python-binary: /opt/venv/bin/python
+          package-name: ${{ matrix.module }}
+          python-binary: python3
 
-  install-test-summary:
-    needs: [pre-flight, pip-test]
+  lint:
```
agronskiy: `question`: we have a linting step which is (in its ruff part) an exact cunterpart of the pre-commit, whiy dfferent here? For example, we have more ruff stuff that needs to be run, and it makes more sense to run pre-commit on all files here and exit 1 if any files were changed.

.github/workflows/install-test.yml
pablo-garay: yes

.github/workflows/install-test.yml
pablo-garay: you want  .11 and .12 also ?

.github/workflows/install-test.yml
pablo-garay: So no lint here

.github/workflows/install-test.yml
pablo-garay: this was updated

pablo-garay:
Install test summaryExpected — Waiting for status to be reported
Check would go away - once merged from main. So plan is: merge this & then "Install test summaryExpected — Waiting for status to be reported" wont be appearing anymore

.github/workflows/install-test.yml
pablo-garay: Unit and functional tests removed

.github/workflows/install-test.yml
```diff
@@ -18,82 +18,130 @@
 name: Installation Test
 
 on:
+  pull_request:
+    branches:
+      - main
```
ko3n1g: This is redundant and will lead to multiple executions of the workflow. Let's remove it to keep the Checks page compact

.github/workflows/install-test.yml
```diff
@@ -18,82 +18,130 @@
...
   push:
     branches:
       - main
       - "pull-request/[0-9]+"
       - "deploy-release/*"
-env:
-  UV_PROJECT_ENVIRONMENT: "./venv"
+  workflow_dispatch:
```
ko3n1g: Do we need this? What's the reason to launch the test from a dispatch event?

.github/workflows/install-test.yml
```diff
@@ -18,82 +18,130 @@
...
-          RUN_ID: ${{ github.run_id }}
-          SKIPPING_IS_ALLOWED: ${{ needs.pre-flight.outputs.docs_only == 'true' || needs.pre-flight.outputs.is_deployment_workflow == 'true' }}
+      - name: Install uv
+        run: |
+          curl -LsSf https://astral.sh/uv/install.sh | sh
+          echo "$HOME/.local/bin" >> $GITHUB_PATH
+
+      - name: Install nemo-evaluator (if needed)
+        if: matrix.package == 'nemo-evaluator-launcher'
+        working-directory: packages/nemo-evaluator
```
ko3n1g: @marta-sd Do we want to install evaluator while import-testing launcher?

.github/workflows/install-test.yml
agronskiy: Nice, thanks. I think it's value in running the whole range of versions.

.github/workflows/install-test.yml
marta-sd: no, we don't. launcher doesn't use the evaluator directly - it's part of the docker image that launcher uses, but it's not present in launcher's py env

.github/workflows/install-test.yml
agronskiy: And why only 3.10 and not all the supported ones (3.10-13)?

.github/workflows/install-test.yml
ko3n1g: @pablo-garay lets remove this

.github/workflows/install-test.yml
pablo-garay: Up to what you want :) I am delivering/providing according to what you tell me are the requirements/wants 
You want all (3.10-13) here?

.github/workflows/install-test.yml
pablo-garay: agreed, good catch 👍

.github/workflows/install-test.yml
pablo-garay: added as per your request :)

.github/workflows/install-test.yml
```diff
@@ -23,77 +23,116 @@ on:
...
+          path: ./FW-CI-templates
 
-      - name: Check imports for nemo_evaluator
-        uses: ./FW-CI-templates/.github/actions/check-imports
+      - name: Set up Python
+        uses: actions/setup-python@v5
         with:
-          package-name: nemo_evaluator
-          python-binary: /opt/venv/bin/python
+          python-version: "3.10"
```
ko3n1g: why are we installing 3.10 if the grid spans over multiple python versions?

pablo-garay: I thought i had changed this, silly miss (sigh)
updated

ko3n1g (approved):
ty

.github/workflows/install-test.yml
```diff
@@ -23,77 +23,116 @@ on:
...
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install uv
+        run: |
+          curl -LsSf https://astral.sh/uv/install.sh | sh
+          echo "$HOME/.local/bin" >> $GITHUB_PATH
+
+      - name: Install nemo-evaluator (dependency)
```
marta-sd: let's remove this install - nemo-evaluator is not a direct dependency of launcher. sorry for pointing it now, I missed that when I did the 1st review

pablo-garay: makes sense removed

pablo-garay:
[ko3n1g](https://github.com/ko3n1g) previously approved these changes previously on behalf of https://github.com/orgs/NVIDIA-NeMo/teams/automation , so we got the code owner reviewer approval already (last commit was minimal change)

PR #313, opened 2025-10-14
chore(exporters): local run + automated local export example to MLflow

ka00ri:


packages/nemo-evaluator-launcher/examples/local_auto_export_llama_3_1_8b_instruct.yaml
```diff
@@ -49,52 +39,24 @@ target:
 
 evaluation:
   tasks:
-    - name: simple_evals.gpqa_diamond
-      overrides:
-        config.params.limit_samples: 10
+    - name: gpqa_diamond
       env_vars:
-        HF_TOKEN: HF_TOKEN
+        HF_TOKEN: HF_TOKEN_FOR_GPQA_DIAMOND
```
AWarno, line 44: maybe we can add a comment here as in other examples, how to get this token

PR #316, opened 2025-10-14
chore(repro): ensure command is logged once

agronskiy:
Following the discussion with @fgalko-oss, the gist of which is that the command we render in `results.yml` and the command we actually use are differing due to the manipulation on `evaluator` object between these two points of time, the proposed solution is to 

1. Do a non-invasive refactor that saves the precise command used for evaluations and makes use of it, rather than re-rendering
2. We drop and error-level log message about rendered  command and add a non-disruptive bash prefix `echo('Warning, the command is potentially different') && <rendered_command>` 

The reason for the above is to be non-disruptive and but visible enough.


agronskiy:
Closing for now in favour of better design after the GA cc @prokotg @fgalko-oss

PR #324, opened 2025-10-15
feat(adapters): dynamic port adapter server

prokotg:
The following MR introduces dynamic port selection for adapter server. During development I realized that we test a completely different function for adapter server spawning than during runtime so that was fixed. Summary:

1. `spawn_adapter_server` was completely removed as it only facilitated testing
2. Checks on empty and disabled interceptors/post-eval hooks were re-introduced into `AdapterServerProcess`
3. Port was provided as an argument to `AdapterServer`
4. Port selection is done inside `AdapterServerProcess`
5. [fix] `AdapterConfig` is not getting a validated form inside `AdapterServerProcess` because a) it was duplicated from`evaluate` function b) `AdapterServerProcess` must handle  empty cases for `AdapterConfig` as well. 

Possible code improvements:
1. Factoring out evaluation fixture

### DIFF
--- a/packages/nemo-evaluator/src/nemo_evaluator/adapters/server.py
+++ b/packages/nemo-evaluator/src/nemo_evaluator/adapters/server.py
@@ -636,7 +636,7 @@ class AdapterServerProcess:
         if not enabled_interceptors and not enabled_post_eval_hooks:
             return
 
-        # Get port from environment variable or use default
+        # Get host from environment variable or use default
         adapter_host = os.environ.get(
             "ADAPTER_HOST", AdapterServer.DEFAULT_ADAPTER_HOST
         )
@@ -663,10 +663,10 @@ class AdapterServerProcess:
 
     def __exit__(self, type, value, traceback):
         if not self.process:
-            return
+            return False
         self.evaluation.target.api_endpoint.url = self.original_url
         try:
-            # Get port from environment variable or use default
+            # Get host from environment variable or use default
             adapter_host = os.environ.get(
                 "ADAPTER_HOST", AdapterServer.DEFAULT_ADAPTER_HOST
             )


prokotg:
/ok to test [6a9324d](https://github.com/NVIDIA-NeMo/Evaluator/pull/324/commits/6a9324de26a52bebc03641332aa09ef113404cbd)

packages/nemo-evaluator/src/nemo_evaluator/adapters/server.py
```diff
@@ -731,3 +584,110 @@ def generate_report(self) -> None:
...
+            try:
+                s.bind((adapter_host, adapter_server_port_env))
+                s.close()
+                return adapter_server_port_env
+            except OSError:
+                s.close()
+                raise OSError(
+                    f"Adapter server was requested to start explicitly on {adapter_server_port_env} through 'ADAPTER_PORT' env-var, but the port seems to be taken. Exiting. "
+                )
+        for port in range(start_port, max_port + 1):
```
agronskiy, line 616: `question(non-blocking)`: there's a more "idiomatic" way of letting OS return any free port to you (as in "OS knows better", advantage is it usually returns normal dynamic/private ports `49152–65535`, because the ones below are in the user range, see RFC 6335) see below, just curious why you considered iterating? a) Ports might be occupied/reserved in big ranges, iteration can take a while if `start_port` is kinda there b) in unfortunate situation where `start_port` is already high one can end up having small range.
```
sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
sock.bind(('your_host', 0))
port = sock.getsockname()[1]
```

prokotg: I hoped that proximity to `DEFAULT_ADAPTER_PORT` can provide a hint where the port selection is coming from. That is, users will be familiar that around 3825, the adapter server is running. If that's not important, I can change to totally random

prokotg:
/ok to test [0545b94](https://github.com/NVIDIA-NeMo/Evaluator/pull/324/commits/0545b94f02cb6f8e01286400b21eeb3418ea020b)

packages/nemo-evaluator/src/nemo_evaluator/adapters/server.py
agronskiy: Sounds legit!

packages/nemo-evaluator/src/nemo_evaluator/adapters/server.py
```diff
@@ -731,3 +584,110 @@ def generate_report(self) -> None:
...
+            return
+        enabled_interceptors = [ic for ic in adapter_config.interceptors if ic.enabled]
+        enabled_post_eval_hooks = [
+            hook for hook in adapter_config.post_eval_hooks if hook.enabled
+        ]
+        if not enabled_interceptors and not enabled_post_eval_hooks:
+            return
+
+        # Get port from environment variable or use default
+        adapter_host = os.environ.get(
```
marta-sd, line 640: nit: `# Get host from environment variable or use default`

packages/nemo-evaluator/src/nemo_evaluator/adapters/server.py
```diff
@@ -731,3 +584,110 @@ def generate_report(self) -> None:
...
+            f"Adapter server failed to start on {adapter_host}:{self.port}"
+        )
+
+    def __exit__(self, type, value, traceback):
+        if not self.process:
+            return
+        self.evaluation.target.api_endpoint.url = self.original_url
+        try:
+            # Get port from environment variable or use default
+            adapter_host = os.environ.get(
```
marta-sd, line 670: nit: `# Get host from environment variable or use default`

packages/nemo-evaluator/src/nemo_evaluator/adapters/server.py
```diff
@@ -731,3 +584,110 @@ def generate_report(self) -> None:
...
+        max_port=65535,
+        adapter_host=AdapterServer.DEFAULT_ADAPTER_HOST,
+    ) -> int:
+        # If specific port has been requested, try only that one port
+        adapter_server_port_env = int(os.environ.get("ADAPTER_PORT", 0))
+        if adapter_server_port_env:
+            s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            try:
+                s.bind((adapter_host, adapter_server_port_env))
+                s.close()
```
wprazuch, line 609: nit(non-blocking): In theory we might get a race condition here - between bind and close there could be another process, which is going to claim the same port at the same time.

What could be done, is:
```python
def _find_and_reserve_free_port(self, adapter_host=AdapterServer.DEFAULT_ADAPTER_HOST) -> int:
    # Handle explicit port request
    adapter_server_port_env = int(os.environ.get("ADAPTER_PORT", 0))
    if adapter_server_port_env:
        s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
        try:
            s.bind((adapter_host, adapter_server_port_env))
            s.close()
            return adapter_server_port_env
        except OSError:
            s.close()
            raise OSError(
                f"Adapter server was requested to start explicitly on {adapter_server_port_env} through 'ADAPTER_PORT' env-var, but the port seems to be taken. Exiting."
            )
    
    # Let OS assign a free port
    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
    s.bind((adapter_host, 0))  # Port 0 = OS assigns free port
    port = s.getsockname()[1]
    s.close()
    return port
```
as a fallback, we let OS choose an appropriate port

packages/nemo-evaluator/src/nemo_evaluator/adapters/server.py
```diff
@@ -731,3 +584,110 @@ def generate_report(self) -> None:
...
+        logger.error(f"Adapter server failed to start on {adapter_host}:{self.port}")
+        self.process.terminate()
+        self.process.join(timeout=5)
+        raise RuntimeError(
+            f"Adapter server failed to start on {adapter_host}:{self.port}"
+        )
+
+    def __exit__(self, type, value, traceback):
+        if not self.process:
+            return
```
wprazuch: Should we not return false here?
```python
    if not self.process:
        return False
```

packages/nemo-evaluator/src/nemo_evaluator/adapters/server.py
```diff
@@ -731,3 +584,110 @@ def generate_report(self) -> None:
...
+                        f"Failed to run post-evaluation hooks: {response.status_code} - {response.text}"
+                    )
+            else:
+                logger.info(
+                    "Server not responding, post-eval hooks already run by signal handler"
+                )
+        except Exception as e:
+            logger.error(f"Failed to run post-evaluation hooks: {e}")
+        self.process.terminate()
+        self.process.join()
```
wprazuch, line 693: nit (non-blocking) we could try and safeguard the join block through:
```python
def _cleanup_process(self):
    """Safely terminate the adapter server process."""
    if not self.process:
        return
    
    if not self.process.is_alive():
        logger.info("Process already terminated")
        return
    
    logger.info("Terminating adapter server process")
    self.process.terminate()
    self.process.join(timeout=30)  # Wait up to 30 seconds
    
    if self.process.is_alive():
        logger.warning("Process didn't terminate gracefully, killing it")
        self.process.kill()  # Force kill
        self.process.join(timeout=10)  # Wait for kill to complete
        
        if self.process.is_alive():
            logger.error("Failed to kill process - may be a zombie")
```
we could then use it in `__enter__` and `__exit__` instead of terminate and join cascade
I think we might have some corner cases due to supporting 40+ harnesses, which may cause the process to hang still. If we are not imposing any timeouts, this might lead to some potential issues, such as excessive compute usage. This is out of scope for this PR, but I wanted to leave some comment so we could re-visit this topic in the next convenience :)

wprazuch (commented):
Left two nits and one comment to align :)

packages/nemo-evaluator/src/nemo_evaluator/adapters/server.py
prokotg: that's functionally the same but I will be explicit and return false, thanks!

packages/nemo-evaluator/src/nemo_evaluator/adapters/server.py
prokotg: I'm sorry, I don't see any change with the proposed solution. Could you elaborate? You still close the socket and the port is passed down to `AdapterServer` later on

packages/nemo-evaluator/src/nemo_evaluator/adapters/server.py
prokotg: fixed

packages/nemo-evaluator/src/nemo_evaluator/adapters/server.py
prokotg: fixed

packages/nemo-evaluator/src/nemo_evaluator/adapters/server.py
wprazuch: Oh right! Sorry, this is wrong 🙈 What I meant is:
```python
def _find_and_reserve_free_port(self, adapter_host=AdapterServer.DEFAULT_ADAPTER_HOST) -> int:
    # Handle explicit port request
    adapter_server_port_env = int(os.environ.get("ADAPTER_PORT", 0))
    if adapter_server_port_env:
        s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
        try:
            s.bind((adapter_host, adapter_server_port_env))
            s.close()
            return adapter_server_port_env
        except OSError:
            s.close()
            logger.warning(
                f"Adapter server was requested to start explicitly on {adapter_server_port_env} through 'ADAPTER_PORT' env-var, but the port seems to be taken. Assigning a free port."
            )
   
            # Let OS assign a free port
            s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
            s.bind((adapter_host, 0))  # Port 0 = OS assigns free port
            port = s.getsockname()[1]
            s.close()
            return port

```

So it could solve two things: port already taken, and race-condition. But it's actually a design decision tbh, and I think yours makes more sense, since we should explicitly terminate if we cannot launch on given port

prokotg:
/ok to test  [b12919f](https://github.com/NVIDIA-NeMo/Evaluator/pull/324/commits/b12919f6c9ecde108ac15e7fb167c3d3f3a8b195)

prokotg:
/ok to test [b12919f](https://github.com/NVIDIA-NeMo/Evaluator/pull/324/commits/b12919f6c9ecde108ac15e7fb167c3d3f3a8b195)

PR #337, opened 2025-10-16
feat: New changelog-build

pablo-garay:


agronskiy (commented):
`question`: Is that supposed to work in non-monorepo, where the same commit history is used for many packages with differing release versioning and cadence? We have a bit or real estate in our workflows devoted to the ability to release the two separately.

pablo-garay:
> `question`: Is that supposed to work in non-monorepo, where the same commit history is used for many packages with differing release versioning and cadence? We have a bit or real estate in our workflows devoted to the ability to release the two separately.

This is for the main general repo release as done in :
image

pablo-garay:
> > `question`: Is that supposed to work in non-monorepo, where the same commit history is used for many packages with differing release versioning and cadence? We have a bit or real estate in our workflows devoted to the ability to release the two separately.
> 
> This is for the main general repo release as done in : image

If not wanted to release, we just opt to not trigger it. We'd manually trigger for this general repo release

ko3n1g (approved):

pablo-garay:
> @pablo-garay , @agronskiy brings up a good question. I'm not sure how this works with the Evaluator repo when it's serving multiple packages. Also, they're currently continually releasing. So is Evaluator an exception in this case?


> the release in the screenshot was a one time thing. We no longer have nemo-eval and we won't be releasing it anymore. We now have two packages in the repo: nemo-evaluator and nemo-evaluator-launcher, they are separate libraries and their releases are separate: can have different versions, must have separate changelogs

PR #342, opened 2025-10-17
chore(cli):  rename debug to info

ka00ri:
- rename debug cmd to info
- include key files into displayed job infomrmation
- fix --help to not  include export flags

### DIFF
--- a/docs/libraries/nemo-evaluator-launcher/configuration/deployment/sglang.md
+++ b/docs/libraries/nemo-evaluator-launcher/configuration/deployment/sglang.md
@@ -14,14 +14,15 @@ See the complete configuration structure in `configs/deployment/sglang.yaml`.
 deployment:
   type: sglang
   image: lmsysorg/sglang:latest
-  checkpoint_path: /path/to/model  # Path to model (local or HuggingFace model ID)
+  hf_model_handle: hf-model/handle  # HuggingFace ID
+  checkpoint_path: null             # or provide a path to the stored checkpoint
   served_model_name: your-model-name
   port: 8000
 ```
 
 **Required Fields:**
 
-- `checkpoint_path`: Model path or HuggingFace model ID (e.g., `meta-llama/Llama-3.1-8B-Instruct`)
+- `checkpoint_path` or `hf_model_handle`: Model path or HuggingFace model ID (e.g., `meta-llama/Llama-3.1-8B-Instruct`)
 - `served_model_name`: Name for the served model
 
 ### Optional Settings
@@ -61,15 +62,14 @@ defaults:
   - _self_
 
 deployment:
-  checkpoint_path: meta-llama/Llama-3.1-8B-Instruct
-  served_model_name: llama-3.1-8b-instruct
-  tensor_parallel_size: 4
-  data_parallel_size: 1
+  checkpoint_path: Qwen/Qwen3-4B-Instruct-2507
+  served_model_name: qwen3-4b-instruct-2507
+  tensor_parallel_size: 8
+  data_parallel_size: 8
   extra_args: ""
-  env_vars:
-    HF_HOME: "/cache/huggingface"   # make sure you have access to GPQA-Diamond and meta-llama/Llama-3.1-8B-Instruct
 
 execution:
+  hostname: your-cluster-headnode
   account: your-account
   output_dir: /path/to/output
   walltime: 02:00:00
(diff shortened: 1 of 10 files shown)


packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/cli/info.py
```diff
@@ -109,21 +114,21 @@ def execute(self) -> None:
         elif self.artifacts:
             logger.info("Showing artifacts locations", job_count=len(jobs))
             self._show_artifacts_info(jobs)
-        elif self.copy_logs is not None:
+        elif self.copy_logs != "__NOT_PROVIDED__":
```
marta-sd: @ka00ri I don't think this works as intended. If I run:
```
nemo-evaluator-launcher info --copy_logs --copy_artifacts 440ce13cab2b7e03 bdbd0cc1a32e78ba
```
it treats the 2nd invocation ID as a dir.

Do we really nee to have these optional defaults? I think it would be less error-prone if we just required the user to pass the directories if they want to copy logs or artifacts.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/cli/info.py
```diff
@@ -35,52 +34,58 @@
...
 
     Examples:
-      nemo-evaluator-launcher debug <inv>                 # Full debug info
-      nemo-evaluator-launcher debug <inv> --config        # Show stored job config (YAML)
-      nemo-evaluator-launcher debug <inv> --artifacts     # Show artifact locations
-      nemo-evaluator-launcher debug <inv> --logs          # Show log locations
-      nemo-evaluator-launcher debug <inv> --copy-logs <path>       # Copy logs (default: current dir)
-      nemo-evaluator-launcher debug <inv> --copy-artifacts <path>   # Copy artifacts (default: current dir)
+      nemo-evaluator-launcher info <inv>                 # Full job info
+      nemo-evaluator-launcher info <inv> --config        # Show stored job config (YAML)
```
marta-sd, line 43: @ka00ri unfortunately this is not how it works - we need to pass `--config True`. This applies to all bool params:
```
(venv) martas@97d83ab-lcedt:~/nvidia-nemo/evaluator$ nemo-evaluator-launcher info --help
usage: nemo-evaluator-launcher info [-h] [-v] [--config bool] [--artifacts bool] [--logs bool] [--copy_logs [str]] [--copy_artifacts [str]] List

Info functionalities for nemo-evaluator-launcher

options:
  -h, --help            show this help message and exit
  -v, --verbose         Enable verbose logging (default: False)

InfoCmd ['info']:
  Job information functionalities for nemo-evaluator-launcher.
  

  List                  IDs to show info for (space-separated). Accepts invocation IDs or/and job IDs. (default: None)
  --config bool, --noconfig bool
                        Show job configuration (default: False)
  --artifacts bool, --noartifacts bool
                        Show artifact locations and key files (default: False)
  --logs bool, --nologs bool
                        Show log locations and key files (default: False)
  --copy_logs [str], --copy-logs [str]
                        Copy logs to local directory (current dir if not specified) (default: __NOT_PROVIDED__)
  --copy_artifacts [str], --copy-artifacts [str]
                        Copy artifacts to local directory (current dir if not specified) (default: __NOT_PROVIDED__)

```

I'm not familiar with simple parsing, but I think we should try to have these params with `action="store_true"`. If it's not feasible, let's update the examples instead.

marta-sd (changes requested):
I was testing the commands and found some issues. See my comments for details

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/cli/info.py
ka00ri: Thanks for looking into this, as I needed a second opinion on the defaults decision. Out for convenience for users with appropriate warning, this could be helpful. But I do see the limitation if using nemo-evaluator-launcher info --copy_logs --copy_artifacts 440ce13cab2b7e03 bdbd0cc1a32e78ba instead of nemo-evaluator-launcher info  440ce13cab2b7e03 bdbd0cc1a32e78ba --copy_logs --copy_artifacts.

ka00ri:
@marta-sd  in case  of nemo-evaluator-launcher info --copy-logs <JobID>,  this will return jobID missing. Examples are clear on expected CLI cmd.

marta-sd:
> @marta-sd in case of nemo-evaluator-launcher info --copy-logs , this will return jobID missing. Examples are clear on expected CLI cmd.

oooh, I see, so it's a matter of how the output of `--help` is rendered. The examples work perfectly fine indeed. Thanks for clarifying, this is something we should address in the future, but doesn't block this PR nor the release.

ka00ri:
Only issue rises if user executes for a list of Ids, then the first is passed as dir. To avoid this, export cmd will also require changes. Definitely to be addressed in the future.

PR #346, opened 2025-10-19
feat(slurm): Add exclusive flag to the Slurm executor

fgalko-oss:


marta-sd (commented):
@fgalko-oss please also add this flag to the defaults in the [config](https://github.com/NVIDIA-NeMo/Evaluator/blob/4285c1e5e3772d5c59a458ee342be013358d775e/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/configs/execution/slurm/default.yaml#L23) and list it in the [docs](https://github.com/NVIDIA-NeMo/Evaluator/blob/main/docs/libraries/nemo-evaluator-launcher/configuration/executors/slurm.md#configuration-example)

marta-sd:
As discussed offline, we'll close this for now. This change can cause a few hard to detect bugs. in short: when you do not have the exclusive flag, 2 benchmarks can be scheduled on the same node and requests can go to the wrong server

PR #347, opened 2025-10-19
chore(ux): dry-run show full config and colored structure

agronskiy:
This PR
- fixes that the dry-run output used to conceal parts of `ef_config` overriding key behind `base64` encoding. It is solved by outputting a commented-out decoded config, see screenshot
- structured and streamlines using of coloring, leading to more recognizable structure of the normal output (see screenshot)

Out of scope:
- there needs to be refactored how we pass components of command. Currently (and historically), the executors were passing a lot via the command directy _into_ `bash -c '...'`. We have ongoing ideas to wrap _everything_ into a config, but then the issue is again passing this config through the _docker_ (lepton, slurm, etc.) boundary, in such a way that it's not stored as a separate artifact (storing so is harder to debug -- external files need to be carried around, while logs are easier to look at).  so the below diagram should be somehow changed so that there's a streamlined component that distinguishes btw dry-run and real run, and still allows wrapping everything into a command, not into a file.

```
                                        ┌──────────────────────────────────┐
                                        │                                  │
┌─────────────────────────┐             │   lepton's docker command        │
│                         │             │                                  │
│  cmd generated          ┼─────────────┼──────────► goes inside           │
│                         ┼─────────┐   └──────────────────────────────────┘
│                         ┼─────┐   │                                       
└─────────────────────────┘     │   │   ┌──────────────────────────────────┐
                                │   │   │  local's docker command          │
                                │   │   │                                  │
┌─────────────────────────┐     │   │   │                                  │
│ debug decoded base64    │     │   └───┼─────────► goes inside            │
│                         │     │       └──────────────────────────────────┘
│ goes into enclosing     │     │                                           
│ scripts                 │     │        ┌─────────────────────────────────┐
│                         │     │        │ slurm's enroot command          │
│                         │     │        │                                 │
│                         │     └────────┼────────► goes inside            │
└─────────────────────────┘              │                                 │
                                         └─────────────────────────────────┘
```

image

### DIFF
--- a/packages/nemo-evaluator-launcher/tests/unit_tests/test_cli_ls_runs_extended.py
+++ b/packages/nemo-evaluator-launcher/tests/unit_tests/test_cli_ls_runs_extended.py
@@ -141,13 +141,13 @@ class TestLsRunsFiltering:
                 "invocation_id": "inv1",
                 "executor": "local",
                 "num_jobs": 5,
-                "earliest_job_ts": 1640908800.0,  # 2021-12-31 00:00:00
+                "earliest_job_ts": 1640419200.0,  # 2021-12-25
             },
             {
                 "invocation_id": "inv2",
                 "executor": "local",
                 "num_jobs": 3,
-                "earliest_job_ts": 1640995200.0,  # 2022-01-01 00:00:00
+                "earliest_job_ts": 1641542400.0,  # 2022-01-07
             },
         ]
 


agronskiy:
Great catch. I tried multiple ways of presenting it within the command (so that the `dry-run` and non-`dry-run` paths are as close to each other as possible) but it seems too dangerous at this point after your catch @marta-sd. Exploring still!

marta-sd (approved):
Thanks!

PR #351, opened 2025-10-20
chore(docs): improve examples and docs for slurm executor

ka00ri:
- document that output_dir must be absolute path
- provide all required fields in the example configs in the repo
- explain that hostname for slurm == headnode
- clarify the difference between checkpoint_path and hf_model_handle and document what should be used as values

### DIFF
--- a/docs/libraries/nemo-evaluator-launcher/configuration/deployment/sglang.md
+++ b/docs/libraries/nemo-evaluator-launcher/configuration/deployment/sglang.md
@@ -14,14 +14,15 @@ See the complete configuration structure in `configs/deployment/sglang.yaml`.
 deployment:
   type: sglang
   image: lmsysorg/sglang:latest
-  checkpoint_path: /path/to/model  # Path to model (local or HuggingFace model ID)
+  hf_model_handle: hf-model/handle  # HuggingFace ID
+  checkpoint_path: null             # or provide a path to the stored checkpoint
   served_model_name: your-model-name
   port: 8000
 ```
 
 **Required Fields:**
 
-- `checkpoint_path`: Model path or HuggingFace model ID (e.g., `meta-llama/Llama-3.1-8B-Instruct`)
+- `checkpoint_path` or `hf_model_handle`: Model path or HuggingFace model ID (e.g., `meta-llama/Llama-3.1-8B-Instruct`)
 - `served_model_name`: Name for the served model
 
 ### Optional Settings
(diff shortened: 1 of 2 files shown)


docs/libraries/nemo-evaluator-launcher/configuration/executors/slurm.md
```diff
@@ -48,4 +106,5 @@ This configuration:
 - Uses the Slurm execution backend
 - Deploys a vLLM model server on the cluster
 - Requests GPU resources (8 GPUs per node, 4-hour time limit)
-- Runs three benchmark tasks
+- Runs three benchmark tasks sequentially
```
marta-sd: They are run in parallel for slurm (each has separate deployment)

packages/nemo-evaluator-launcher/examples/slurm_llama_3_1_8b_instruct.yaml
```diff
@@ -35,11 +35,12 @@ defaults:
 
 # override default execution arguments
 execution:
-  account: ???
-  output_dir: /some/filesystem/path/available/to/slurm/node
+  hostname: ???          # SLURM headnode hostname (required)
+  username: ???          # Optional: defaults to ${oc.env:USER} from slurm/default
```
marta-sd: Let's not override the default to ??? here, it works for most users. Maybe we could add the same value as default here just for the sake of showing this param to the users?

packages/nemo-evaluator-launcher/examples/slurm_llama_3_1_8b_instruct.yaml
```diff
@@ -50,7 +51,9 @@ execution:
 
 # override default deployment arguments
 deployment:
-  checkpoint_path: /some/filesystem/path/available/to/slurm/node
+  # Prefer hf_model_handle for auto-download; if using checkpoint_path, the directory must exist and contain model files
+  # Alternatively, use checkpoint_path: null and set checkpoint_path to the path of the local model files
+  checkpoint_path: /shared/filesystem/path/available/to/slurm/node # must be an absolute path and EXISTS on the SLURM nodes
```
marta-sd: Maybe we could set this one to ??? too?

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/configs/execution/slurm/default.yaml
```diff
@@ -17,17 +17,19 @@
 
 # Executor is chosen based on this field
 type: slurm
-
 hostname: ???
 username: ${oc.env:USER}
 account: ???
+hostname: ???            # SLURM headnode (login) hostname (required)
```
marta-sd, line 18: Repeated keys- we already had hostname etc

ka00ri: Thanks for catching this one. I am still in the "merging" process, and investigating the CI failure.

packages/nemo-evaluator-launcher/examples/slurm_llama_3_1_8b_instruct.yaml
ka00ri: Sure!

PR #355, opened 2025-10-20
docs(deployment): use full example for locally-deployed model and launcher

marta-sd:


docs/get-started/quickstart/index.md
```diff
@@ -70,16 +70,57 @@ NeMo Evaluator works with any OpenAI-compatible endpoint. You have several optio
 
 If you prefer to host your own models:
 
+
+If you are deploying the model locally with Docker, you can use a .
```
agronskiy: `typo`: there's part missing?

docs/get-started/quickstart/index.md
```diff
@@ -70,16 +70,57 @@ NeMo Evaluator works with any OpenAI-compatible endpoint. You have several optio
...
+target:
+  api_endpoint:
+    model_id: microsoft/Phi-4-mini-instruct
+    url: http://my-phi-container:8000/v1/chat/completions
+    api_key_name: null
+
+evaluation:
+  tasks:
+    - name: simple_evals.mmlu_pro
+      overrides:
```
agronskiy, line 110: `warning`/`awareness`: this will needs an almost immediate follow-up after @AWarno finishes deprecating `overrides`

marta-sd: right, but if I use the `nemo_evaluator_config` it won't be valid (it's not called like that yet), and if I use `config` we'll need to have the follow-up too. let me leave the overrides, they will be easier to grep for updating

docs/get-started/quickstart/index.md
marta-sd: actually this part shouldn't be there :D I move some fragments to comments and forgot to remove the rest

PR #359, opened 2025-10-21
feat: add labeler

pablo-garay:


athitten (dismissed):
LGTM, thank you @pablo-garay !

.github/labeler.yml
```diff
@@ -0,0 +1,32 @@
...
+
+scripts:
+- scripts/**/*
+
+CI:
+- .github/**/*
+- Dockerfile
+- docker/**/*
+
+deployment:
```
marta-sd: I think these two labels (deployment and evaluation) won't be needed.

marta-sd (commented):
Thank @pablo-garay, I reviewed the set of labels and assume that someone from the automation team will review the new workflow :)

.github/labeler.yml
pablo-garay: removed as per your request/feedback TYSM

PR #386, opened 2025-10-24
docs(tutorial): local evaluation of existing endpoint

marta-sd:


docs/tutorials/local-evaluation-of-existing-endpoint.md
```diff
@@ -47,70 +57,122 @@ For a comprehensive list of supported tasks and descriptions, see {ref}`nemo-eva
...
-  overrides:  # these overrides apply to all tasks; for task-specific overrides, use the `overrides` field
-    config.params.request_timeout: 3600
+  # Optional: Global evaluation overrides - these apply to all benchmarks below
+  nemo_evaluator_config:
+    config:
+      params:
+        max_new_tokens: 32768
+        temperature: 0.6
+        top_p: 0.95
+        request_timeout: 1600
```
AWarno, line 97: For this model, these tasks, these params do not really match + setting a 32k OSL may contribute to overloading the NVCF function. Perhaps we can only leave  example overridden parameters for HumanEval.

PR #392, opened 2025-10-26
fix: Fix build docs

pablo-garay:


docs/tutorials/nemo-fw/nemo-run.md
```diff
@@ -63,8 +63,8 @@ or a Slurm one:
 :::{note}
 Please make sure to update `HF_TOKEN` with your token
 
-- in the NeMo Run script's [local_executor env_vars](https://github.com/nvidia-nemo/evaluator/blob/main/scripts/evaluation_with_nemo_run.py#l267) if using local executor
-- in the [slurm_executor's env_vars](https://github.com/nvidia-nemo/evaluator/blob/main/scripts/evaluation_with_nemo_run.py#l232) if using slurm_executor.
+- in the NeMo Run script's [local_executor env_vars](https://github.com/NVIDIA-NeMo/Evaluator/blob/main/scripts/evaluation_with_nemo_run.py#L274) if using local executor
+- in the [slurm_executor's env_vars](https://github.com/NVIDIA-NeMo/Evaluator/blob/main/scripts/evaluation_with_nemo_run.py#L237) if using slurm_executor.
```
chtruong814, line 67: It looks like this is still failng.  I think we ran into this before with NeMo and Github links.  In some cases unfortunately, the sphinx link checker is not able to verify a link if the anchor is managed through something like javascript rather than an actual html element.  

I would recommend disabling the check for anchors if we're heavily using those here.
https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-linkcheck_anchors

Additionally, at some point it may get challenging to check Github links.  We were getting severely rate limited last time we did this in other repos.  Just something to look out for.

What I have found with docs link validation is that it can be very flakey. That's why the false positives allow list exists. Of course, if we put every link in there, this test does not help. If it's useful for some of the links, then there's some value to keeping it.  

I'd recommend getting back to passing tests in as simple a way as possible. The links in the docs should reflect whatever is desired for the docs, whether it's with or without anchors. This test should not hinder that, only flag what it can.  And if it's ultimately not adding any value, we can consider disabling entirely.  Most repos have some form of link check happening at least though.

cc: @marta-sd

marta-sd:
closing in favor of #404

PR #411, opened 2025-10-30
feat(metadata): add version tracing

agronskiy:
This:
- on the launcher side, propagates the `metadata` information that contains `versioning` and `launcher_resolved_config` fields into the `ef_config.yaml` plus prints the debug information to the command, so that it's visible on the logs already
- on the evaluator side: it takes this metadata and
  - saves the launcher_config as a separate file with a comment header `# was generated with this  and this versions`
  - adds the eval factory varsion, git hash of the container and saves the full versioining information to the `results.yml` for quick debugging. 

The former makes it already useful for debug, the latter will propagate once the containers update.

All changes under tests and should be backward compatible (i.e. no cod erequires those fields)

image
image

### DIFF
--- a/packages/nemo-evaluator/src/nemo_evaluator/core/evaluate.py
+++ b/packages/nemo-evaluator/src/nemo_evaluator/core/evaluate.py
@@ -265,7 +265,7 @@ def _persist_metadata_and_build_results_block(
     # ruff.
     from nemo_evaluator import __version__ as nemo_evaluator_version
 
-    updated_versioning["nemo_evaluator_version"] = nemo_evaluator_version
+    updated_versioning["nemo_evaluator"] = nemo_evaluator_version
 
     # Construct full metadata payload to persist and return, augmenting versioning
     # with inferred fields (git-hash, nemo_evaluator_version).
--- a/packages/nemo-evaluator/tests/unit_tests/core/test_metadata.py
+++ b/packages/nemo-evaluator/tests/unit_tests/core/test_metadata.py
@@ -108,7 +108,7 @@ def test_metadata_full(monkeypatch, tmp_path):
     expected_versioning = {
         "foo": "1.2.3",
         "git-hash": "abc123",
-        "nemo_evaluator_version": nemo_evaluator_version,
+        "nemo_evaluator": nemo_evaluator_version,
     }
     expected_metadata_full = {
         "versioning": expected_versioning,
@@ -156,7 +156,7 @@ def test_env_versioning_and_resolved_only(monkeypatch, tmp_path):
     results = yaml.safe_load((tmp_path / "results.yml").read_text())
     mv = results["metadata"]["versioning"]
     assert mv["git-hash"] == "deadbeef"
-    assert mv["nemo_evaluator_version"] == nemo_evaluator_version
+    assert mv["nemo_evaluator"] == nemo_evaluator_version
 
 
 def test_only_versioning_no_env(monkeypatch, tmp_path):
@@ -180,5 +180,5 @@ def test_only_versioning_no_env(monkeypatch, tmp_path):
     ]
     assert mv["bar"] == "9.9.9"
 
-    assert mv["nemo_evaluator_version"] == nemo_evaluator_version
+    assert mv["nemo_evaluator"] == nemo_evaluator_version
     assert "git-hash" not in mv


packages/nemo-evaluator/src/nemo_evaluator/api/api_dataclasses.py
```diff
@@ -114,6 +114,23 @@ class EvaluationConfig(BaseModel):
     type: Optional[str] = Field(description="Type of the task", default=None)
 
 
+class EvaluationMetadata(BaseModel):
```
prokotg: Do we need a pre-defined structure here? If we are to pass metadata maybe it would make sense to allow end users to pass their own as well. Maybe here, it would be a model version, quantization algorithm etc. That is, I would like to keep SDK and launcher de-coupled. Wonder what you think about that.

In such case, metadata would be just of type `dict`. Please note as both these fields are optional, we are not validating anything but we do block other metadata to be easily injected

packages/nemo-evaluator/src/nemo_evaluator/core/evaluate.py
```diff
@@ -211,3 +220,44 @@ def run_evaluation_core():
...
+    def _write_with_versioning_header(filename: str, payload: dict):
+        header = "# Generated by nemo-evaluator invocation with the versions of components:\n"
+        versioning_yaml = yaml.safe_dump(versioning, sort_keys=False).rstrip("\n")
+        header += "\n".join(f"# {line}" for line in versioning_yaml.splitlines()) + "\n"
+        path = os.path.join(out_dir, filename)
+        with open(path, "w") as f:
+            f.write(header)
+            yaml.safe_dump(payload, f, sort_keys=False)
+
+    if md.launcher_resolved_config:
```
prokotg: Just to highlight what I wrote in another comment, I think in Core there should be no mention about the launcher, but I'm happy to discuss

prokotg (commented):
Overall looks great! I would love to discuss freeing metadata even more though

packages/nemo-evaluator/src/nemo_evaluator/core/evaluate.py
agronskiy: Ok, I initially meant that with free metadata but had a sensation you wanted a bit better typing. Ok, dicts it is.

packages/nemo-evaluator/src/nemo_evaluator/api/api_dataclasses.py
agronskiy: I kept the versioning field (it reflects that intranlly, core will operate it -- we populate it with core version and git hash), and I removed launcher specific ones, making it just `data` --> that's where user puts whatver they need and we don't require any structure schema.

packages/nemo-evaluator/src/nemo_evaluator/core/evaluate.py
agronskiy: Same as above, agree fully, the field is just `data` now.

packages/nemo-evaluator/src/nemo_evaluator/core/evaluate.py
```diff
@@ -211,3 +220,61 @@ def run_evaluation_core():
...
+    Writes the full metadata payload to `metadata.yaml` with a versioning header.
+    Returns "metadata.verioning" block which later will be included into results.
+    """
+    if not md:
+        return {}
+
+    # Build versioning block first so we can include it as a commented header
+    # in the persisted metadata file for better provenance/debuggability.
+    updated_versioning: dict = dict(md.versioning or {})
+    git_hash = os.getenv("CORE_EVALS_GIT_HASH")
```
prokotg, line 259: a comment that is probably outside of this MR: `CORE_EVALS_GIT_HASH` is only going to work for containers. I think we might find a way to also inject this information for wheels

packages/nemo-evaluator/src/nemo_evaluator/core/evaluate.py
```diff
@@ -211,3 +220,61 @@ def run_evaluation_core():
...
+    if git_hash:
+        updated_versioning["git-hash"] = git_hash
```
prokotg, line 261: nit: generalize underscore / dash

packages/nemo-evaluator/src/nemo_evaluator/api/api_dataclasses.py
prokotg: Sounds good! I also checked that in princple, even if the user decides to put anything into versioning, it's going to be replaced by the Core via `updated_versioning`

packages/nemo-evaluator/src/nemo_evaluator/core/evaluate.py
```diff
@@ -198,11 +202,16 @@ def run_evaluation_core():
     with open(metrics_path, "w") as f:
         json.dump(merged_metrics, f, indent=2)
 
+    metadata_block = _persist_metadata_and_build_results_block(
```
prokotg: maybe we can move it way earlier? Maybe even as one of the first steps. If something happens to this point, we will not have metadata and versoning

packages/nemo-evaluator/src/nemo_evaluator/core/evaluate.py
```diff
@@ -47,7 +49,9 @@ def parse_output(evaluation: Evaluation) -> EvaluationResult:
 
 
 def evaluate(
-    eval_cfg: EvaluationConfig, target_cfg: EvaluationTarget
+    eval_cfg: EvaluationConfig,
+    target_cfg: EvaluationTarget,
+    metadata: Optional[EvaluationMetadata] = None,
```
prokotg, line 55: seems equivalent CLI is missing. We generally try to keep parity between CLI / Python API but maybe metadata will not be a part of it (just as adapters are not directly configurable). Happy to briefly discuss, don't have a strong opinion here

packages/nemo-evaluator/tests/unit_tests/core/test_metadata.py
```diff
@@ -0,0 +1,202 @@
...
+    expected_metadata_full = {
+        "versioning": expected_versioning,
+        "data": {"b": 2},
+    }
+    assert yaml.safe_load(md_file.read_text()) == expected_metadata_full
+
+    results = yaml.safe_load(results_path.read_text())
+    assert results["git_hash"] == "abc123"
+    assert results["metadata"]["versioning"] == expected_versioning
+    assert "data" not in results["metadata"]
```
prokotg: help me out here, why would this not be present? :)

packages/nemo-evaluator/src/nemo_evaluator/core/evaluate.py
agronskiy: Thisis out of scope of this PR, this is the same logic as previously. In fact there's struct improvement, wheels we have now version.

packages/nemo-evaluator/src/nemo_evaluator/core/evaluate.py
agronskiy: I kept consistent with the existing results field git-hash

packages/nemo-evaluator/src/nemo_evaluator/api/api_dataclasses.py
agronskiy: You mean something with conflicting name?

packages/nemo-evaluator/src/nemo_evaluator/core/evaluate.py
agronskiy: Good point, will do

packages/nemo-evaluator/src/nemo_evaluator/core/evaluate.py
prokotg: make sense, thanks!

packages/nemo-evaluator/src/nemo_evaluator/core/evaluate.py
agronskiy: We discussed to keep CLI out of scope (mainly unclear how to pass complex object via flag)

packages/nemo-evaluator/tests/unit_tests/core/test_metadata.py
agronskiy: We agreed that that stuff is stored in metadata.yaml, right?

packages/nemo-evaluator/src/nemo_evaluator/core/evaluate.py
prokotg: Also, we were considering if metadata setting should be something we consider a "quick setup" and it might be not. Thanks for discussing!

packages/nemo-evaluator/tests/unit_tests/core/test_metadata.py
prokotg: Ah, I thought it's going to be stored in both places (metadata.yml and run_ocnfig) . I'm happy to have a separate file only. Thanks!

packages/nemo-evaluator/src/nemo_evaluator/api/api_dataclasses.py
prokotg: no, if they want to put their own versoning, even not by mistake. I think current design is good. No AI. Thanks!

packages/nemo-evaluator/src/nemo_evaluator/core/evaluate.py
agronskiy: Done!

agronskiy:
I tessted it with a local dev container that has this version in it, the results 
image
image

PR #414, opened 2025-10-31
docs: Fix metrics names in MLFlow examples

marta-sd:


ka00ri (approved):
LGTM. The `log_metrics` param filters metrics based on passed keys. Here you can pass either full name or a string chunk of expected metric names and it will return all that match within results.

PR #415, opened 2025-10-31
fix(lepton): completed status is success too

prokotg:


agronskiy (approved):

agronskiy:
Thanks, that was a good catch. Did we check for other statuses that lepton can return?

prokotg:
> Thanks, that was a good catch. Did we check for other statuses that lepton can return?

I just checked documentation and I don't see any list available. I'm going to merge that one and follow up with Lepton folks

PR #416, opened 2025-10-31
feat(reasoning): rename reasoning legacy field [backward-compatible]

prokotg:
Legacy's argument `use_reasoning` might wrongly hint at enabling/disabling reasoning on endpoint side. This MR renames it to `process_reasoning_traces` in a backward compatible manner.


agronskiy:
`suggestion` a good idea for deprecation is: quickly add a test (look up for overrides) that will start failing either on configured date or when > then configured version. Easy to reschedule but will nudge us regularly.

PR #423, opened 2025-11-03
feat(adapter): add optional update for progress tracking on timer

findkim:
Inference can be slow depending on the model performance, the number of GPUs the model is running on, the context length or tokens to generate or parallelism. For slow inference times, an interval based progress tracking may be a frustrating user experience to wait 30mins or longer for the first progress update.

This PR adds an optional parameter to ProgressTracking adapter to update progress on a timed interval, not enabled by default. ProgressTrackingInterceptor tracks the last status that was posted, and skips sending a request / writes to file if there has not been a change. The timed update is run as a asyncio task that is stopped on `post_eval_hook`


packages/nemo-evaluator/src/nemo_evaluator/adapters/interceptors/progress_tracking_interceptor.py
```diff
@@ -83,15 +90,27 @@ def __init__(self, params: Params):
...
         self._samples_processed = self._initialize_samples_processed()
+        self._last_updated_samples_processed = self._samples_processed
         self._lock = threading.Lock()
 
+        # Optional update on timer
+        self.progress_tracking_interval_seconds = (
+            params.progress_tracking_interval_seconds
+        )
+        self._update_on_timer_task = None
+        if self.progress_tracking_interval_seconds:
```
agronskiy, line 103: `question` / `concern`: since this is `asyncio` inside multithreaded Flask, could you clarify how this task would behave? Is the asyncio loop executed on in the main Flask thread where the object is created, and  would that keep blocking? There are suggestions on SO to have an asyncio on a separate thread to avoid issues with asyncio + threading marriage.

findkim: Thanks for the quick review! I’ve verified asyncio.create_task working with fastapi but have not with flask, this is a good callout to do extra testing

findkim: I've added a test on the adapter layer and found that `asyncio.create_task` errors with no running event loop which tracks with your series of questions about the asyncio loop. Once I reworked the background timer to run as a thread and removed asyncio.sleep, the Flask server seems to running as expected and the timer thread does not block incoming requests. Lmk what you think!

agronskiy:
I think we can  merge this since you tested this on this adapter.

PR #426, opened 2025-11-04
fix(exporters): remove score patterns and refactor parsing logic

marta-sd:
This PR introduces the following changes:

* score patterns list was removed
* parsing logic was simplified - we from now on rely on `results.yml` implementing the structure from the `EvaluationResult` dataclasses
* nested groups parsing was added
* stats associated with metrics were added
* keys were simplified - when metric and score names are the same, they are not repeated

We are removing score patterns as it's not feasible to guess all possible score names. We can on the other hand safely assume that `results.yml` has a certain structure, as it goes through a dataclass.
Also if there are too many metrics user can filter them with `log_metrics`, e.g. running the following command for ifbench job:

```bash
nemo-evaluator-launcher export f45fd1b07001ab3b --dest local --format json --log_metrics prompt-level
```

produces the following `processed_results.json`:
```json
{
  "export_timestamp": "2025-11-04T14:59:38.619455",
  "benchmarks": {
    "ifbench": {
      "models": {
        "meta-llama/Llama-3.2-1B-Instruct": [
          {
            "invocation_id": "f45fd1b07001ab3b",
            "job_id": "f45fd1b07001ab3b.0",
            "harness": "ifbench",
            "container": "nvcr.io/nvidia/eval-factory/ifbench:25.10",
            "scores": {
              "ifbench_loose:prompt-level": 0.24829931972789115,
              "ifbench_strict:prompt-level": 0.20068027210884354
            },
            "timestamp": "2025-11-04T14:59:38.619446",
            "executor": "local"
          }
        ]
      }
    }
  }
}
```


ka00ri:
LGTM. Users can still filter with log_metrics if they want only main scores.
This won't affect MLflow/WandB exporters as they treat metrics as a flat dict, so the additional stats will just be logged similar to scores.

AWarno (approved):
LGTM!

PR #428, opened 2025-11-04
feat(core): introduce pre-eval hooks and e2e tests caching logic

Glorf:
Allow pre-eval hooks
Flatten hook logic - all hooks should be treated as interceptors
Implement logic facilitating e2e testing using pre- and post-eval hooks
Facilitate debugging by finding closest cached request while using test mode
Implement pre-eval hooks for custom datasets downloads


Glorf:
To be re-iterated in a next PR (I have a better idea for this feature)

PR #448, opened 2025-11-12
feat(cli): --config <joint path> instead the two

agronskiy:
* now one can just pass one `--config <path to yaml>` instead of `--config-dir` and `--config-name`
* the previous way works but will be deprecated (depr warning added and a test will start failing in Jan)
* tests added


AWarno (approved):
lgtm

PR #474, opened 2025-11-19
docs: Extend evaluation section

marta-sd:


docs/evaluation/logprobs.md
```diff
@@ -0,0 +1,100 @@
...
+```
+
+
+## How it works
+
+When the server receives a `logprob=<int>` parameter in the request, it will return the log-probabilities of tokens.
+When combined with `echo=true`, the model will include the input in its response, along with the corresponding log-probabilities.
+
+Then the recieved response in processed on the client (benchmark) side to isolate the log-probabilities corresponding specifically to the answer portion of the input.
+For this purpose the input is tokenized, which allow to trace which log-probabilities originated from the question, and which from the answer.
```
athitten:
```suggestion
For this purpose the input is tokenized, which allows to trace which log-probabilities originated from the question, and which from the answer.
```

docs/evaluation/logprobs.md
```diff
@@ -0,0 +1,100 @@
...
+    },
+```
+
+
+## How it works
+
+When the server receives a `logprob=<int>` parameter in the request, it will return the log-probabilities of tokens.
+When combined with `echo=true`, the model will include the input in its response, along with the corresponding log-probabilities.
+
+Then the recieved response in processed on the client (benchmark) side to isolate the log-probabilities corresponding specifically to the answer portion of the input.
```
athitten:
```suggestion
Then the recieved response is processed on the client (benchmark) side to isolate the log-probabilities corresponding specifically to the answer portion of the input.
```

docs/evaluation/run-evals/logprobs.md
```diff
@@ -0,0 +1,100 @@
...
+
+
+## Prerequisites
+
+Before using a benchmark that utilizes log-probabilities, ensure:
+
+- Your model endpoint is of type "completions" and supports `logprobs` and `echo` parameters (see {ref}`compatibility-log-probs`)
+- You have access to the model's tokenizer and it is of one of the supported types: `huggingface` or `tiktoken`.
+
+
```
athitten, line 49: I think it might be good to add a line here saying we are using `piqa` as an example of logporb benchmark from lm-eval-harness. Some of the other logprob based benchmarks that are in lm-eval-harness are: arc_challenge, hellaswag etc., Basically the ones we had listed [here](https://github.com/NVIDIA-NeMo/Evaluator/blob/main/docs/nemo-fw/logprobs.md). WDYT ?

docs/evaluation/logprobs.md
```diff
@@ -0,0 +1,100 @@
...
+
+Using log-probabilities is especially useful for evaluating base (pre-trained) models, as it eliminates the need for complex instruction-following and does not require the model to adhere to a specific output format.
+
+
+## Prerequisites
+
+Before using a benchmark that utilizes log-probabilities, ensure:
+
+- Your model endpoint is of type "completions" and supports `logprobs` and `echo` parameters (see {ref}`compatibility-log-probs`)
+- You have access to the model's tokenizer and it is of one of the supported types: `huggingface` or `tiktoken`.
```
athitten: I know we had some discussion about this in the past, to refresh myself does evaluator support loading of only these two tokenizer types ? What's the guidance for other tokenizers ?

docs/get-started/_snippets/lambada.py
```diff
@@ -18,35 +18,27 @@
...
 )
 
 eval_config = EvaluationConfig(
-    type="lm-evaluation-harness.lambada_openai",
+    type="lm-evaluation-harness.polemo2",
     output_dir="/results/",
-    params=ConfigParams(
-        limit_samples=10,
-        temperature=0,
-        top_p=0,
```
athitten, line 43: Maybe its good to show these params (also few_shot) since for custom tasks the default settings are not optimal ?

athitten, line 43: Maybe its good to show these params (incl num_fewshot) to be super clear what to change and where, since the default ones are not optimal for custom tasks

docs/evaluation/_snippets/piqa_hf.py
```diff
@@ -0,0 +1,42 @@
...
+# Configure the evaluation target
+api_endpoint = ApiEndpoint(
+    url="http://0.0.0.0:8000/v1/completions/",
+    type="completions",
+    model_id="meta-llama/Llama-3.1-8B",
+)
+eval_target = EvaluationTarget(api_endpoint=api_endpoint)
+eval_params = ConfigParams(
+    extra={
+        "tokenizer": "meta-llama/Llama-3.1-8B",  # or path to locally stored checkpoint
```
athitten:
```suggestion
        "tokenizer": "meta-llama/Llama-3.1-8B",  # or path to locally stored checkpoint with tokenizer
```

docs/deployment/bring-your-own-endpoint/testing-endpoint-oai-compatibility.md
```diff
@@ -211,6 +208,27 @@ curl -X POST ${FULL_ENDPOINT_URL} \
...
+
+For evaluation with log-probabilities your `completions` endpoint must support `logprobs` and `echo` parameters:
+
+```bash
+export FULL_ENDPOINT_URL="https://your-server.com/v1/completions"
+export API_KEY="your-api-key-here"
+
+curl -X POST ${FULL_ENDPOINT_URL} \
+-H "Content-Type: application/json" \
+-H "Authorization: Bearer ${API_KEY}" \
```
athitten, line 223: Do you think its worth adding a note that `API_KEY` can be skipped if your server does not need it or its understood ?

docs/evaluation/_snippets/polemo2.py
```diff
@@ -18,35 +18,27 @@
...
+        type=EndpointType.COMPLETIONS,
+        model_id=model_name,
+        api_key="API_KEY",
     )
 )
 
 eval_config = EvaluationConfig(
-    type="lm-evaluation-harness.lambada_openai",
+    type="lm-evaluation-harness.polemo2",
     output_dir="/results/",
```
athitten, line 40: Maybe its good add `ConfigParams(` with top_p, temp, num_fewshot etc., so that its super clear to the user where they need to set it since the custom tasks don't have optimal default values for these ?

docs/evaluation/run-evals/logprobs.md
marta-sd: Good point, let me add it. Btw I forgot to mention: I replaced arc_challenge with piqa because it's present in launcher too.

docs/evaluation/logprobs.md
marta-sd: This limitation comes from the upstream lm-eval-harness ([here](https://github.com/EleutherAI/lm-evaluation-harness/blob/7ddb2b1e4bc819292ff56f334c572d1fc77dec28/lm_eval/models/api_models.py#L117C49-L117C55)). I see they have also added a "remote" tokenizer recently (i.e. support for the vllm [tokenizer endpoint](https://docs.vllm.ai/en/latest/serving/openai_compatible_server/#tokenizer-api)). I think this is really cool and something that would generalize really well.

So far we didn't get any requests to support other tokenizer types, but I think if users start to ask for such feature, we could leverage the remote approach. I think adding tokenizer endpoint to the NeMo deployment shouldn't be too difficult. Adding it on the server side would also simplify the evaluation configuration for HF and tiktoken (no need to provide path to the tokenizer, no risk that user accidentally specifies a wrong one). What do you think?

docs/evaluation/_snippets/polemo2.py
marta-sd: Actually they are configured on the task side (I mention it in the docs [here](https://github.com/NVIDIA-NeMo/Evaluator/pull/474/files#diff-f23fa1becb87b9a2564dd41040ae864647ba32d73da47f0742f20812bccb18e0R17)) and I think this is a typical situation. So maybe I can add those fields, but comment them out, just to show how they could be adjusted if needed. What do you think?

docs/get-started/_snippets/lambada.py
```diff
@@ -18,35 +18,27 @@
...
-        parallelism=1,
-        extra={
```
marta-sd, line 45: @athitten I replcaed lambada with polemo2 because I wanted to have a minimal example here. I was worried that users might be confused if they saw a tokenizer specified here and think it's always required for custom tasks

docs/deployment/bring-your-own-endpoint/testing-endpoint-oai-compatibility.md
marta-sd: We have a tip at the top of the doc (since it applies to all tests)

docs/evaluation/run-evals/logprobs.md
marta-sd: I've added a tip: [dda6ce3](https://github.com/NVIDIA-NeMo/Evaluator/pull/474/commits/dda6ce35bb36a07a40e5fbaa503a6b37dd2429b0)image

athitten: Thanks!

docs/evaluation/logprobs.md
athitten: Yeah sounds good, we can look into others if we need to as and when the request comes. Agree that tokenizer endpoint is cool and we can add it to nemo deployment.

docs/get-started/_snippets/lambada.py
athitten: Makes sense, polemo2 is okay with me.

docs/evaluation/_snippets/polemo2.py
athitten: Okay then the params configured on the task side are just not optimal correct ? 
Yes commenting them out sounds good to me.

docs/get-started/_snippets/lambada.py
athitten: Ignore duplicate comment

athitten: Ignore duplicate comment

docs/evaluation/run-evals/index.md
```diff
@@ -1,9 +1,6 @@
----
-orphan: true
----
 (eval-run)=
 
-# Run Evaluations
+# Evaluation Techniques
 
 Follow step-by-step guides for different evaluation scenarios and methodologies in NeMo Evaluation.
```
athitten:
```suggestion
Follow step-by-step guides for different evaluation scenarios and methodologies in NeMo Evaluator.
```

athitten: NeMo Evaluation may give an impression that its specific to nemo fw

docs/evaluation/run-evals/index.md
```diff
@@ -12,48 +9,13 @@ Follow step-by-step guides for different evaluation scenarios and methodologies
 Ensure you have:
 
 1. Completed the initial getting started guides for {ref}`gs-install` and {ref}`gs-quickstart`.
```
athitten, line 11: Dont we want to mention have your deployment ready ? or are you thinking it would be covered via `quickstart`

docs/evaluation/run-evals/text-gen.md
```diff
@@ -41,407 +35,88 @@ In text generation evaluation:
...
-  - Generated responses assessed against instruction criteria
-* - Mathematical Reasoning
-  - Test mathematical problem-solving and multi-step reasoning
-  - - `gsm8k`
-    - `math`
-  - Final answer extraction and numerical comparison
-* - Multilingual Evaluation
-  - Assess capabilities across different languages
-  - - `mgsm` (multilingual GSM8K)
-  - Language-specific text generation and assessment
```
athitten, line 180: I thought this categorization within generation benchmarks was nice to have. Are we having it somewhere else or removing it for a reason ?

docs/evaluation/run-evals/text-gen.md
```diff
@@ -41,407 +35,88 @@ In text generation evaluation:
...
-```
-
-:::{note}
-Task availability depends on installed frameworks. Use `nemo-evaluator-launcher ls tasks` to see the complete list for your environment.
-:::
-
-## Task Naming and Framework Specification
```
athitten, line 187: Also curious about this content, do we have it somewhere else or we don't want it ?

docs/evaluation/run-evals/text-gen.md
```diff
@@ -41,407 +35,88 @@ In text generation evaluation:
...
-This guide shows minimal configuration for getting started. For comprehensive parameter options including:
-- Framework-specific parameters (`num_fewshot`, `tokenizer`, etc.)
-- Optimization patterns for different scenarios
-- Troubleshooting common configuration issues
-- Performance tuning guidelines
-
-See {ref}`eval-parameters`.
-:::
-
-**Key Parameters for Text Generation**:
```
athitten, line 330: Same comment as above for the rest of the content below

docs/evaluation/run-evals/index.md
marta-sd: Good point, we should mention that user need to have either endpoint or checkpoint they would like to deploy. Let me add it.

docs/evaluation/run-evals/text-gen.md
marta-sd: I'm removing it because it's not fully accurate (it wasn't provided by our team) and I don't have bw to work on it now.

I agree it's nice, I also like the fact it's so concise, but it's to some extent covered by categorization we have here: https://docs.nvidia.com/nemo/evaluator/latest/evaluation/benchmarks.html#benchmark-categories so I think it's not critically needed

docs/evaluation/run-evals/text-gen.md
marta-sd: Same here: it wasn't provided by our team and I don't have time to fix it.

docs/evaluation/run-evals/text-gen.md
marta-sd: This is not the right place to have this information. This will be covered in `docs/evaluation/parameters.md` (another section, that requires review from our team, ETA: early Dec)

docs/evaluation/_snippets/polemo2.py
marta-sd: I think they are a good default, but might require tuning for particular model (e.g. if it works better with different temperature). I think the primary reason for configuring the task by the user is simply lack of parameters on the task side.

Also let me clarify that by "task side" I mean the task definition in upstream lm-eval (for polemo2 it's [here](https://github.com/EleutherAI/lm-evaluation-harness/blob/7ddb2b1e4bc819292ff56f334c572d1fc77dec28/lm_eval/tasks/polemo2/polemo2_in.yaml#L19-L20)).

I will add them commented out (using defaults from polemo2) and try to better explain it in the text.

docs/deployment/bring-your-own-endpoint/testing-endpoint-oai-compatibility.md
```diff
@@ -211,6 +208,27 @@ curl -X POST ${FULL_ENDPOINT_URL} \
...
+-d '{
+  "prompt": "3 + 3 = 6",
+  "model": "<YOUR_MODEL_NAME>",
+  "max_tokens": 1,
```
AWarno: Side note: I would also parameterize <YOUR_MODEL_NAME>. We’ve had explicit requests to parameterize items at the top, I can create a PR

docs/evaluation/custom-task.md
```diff
@@ -0,0 +1,38 @@
...
+
+### 1. Prepare the Environment
+
+Start `lm-evaluation-harness` Docker container:
+
+```bash
+docker run --rm -it nvcr.io/nvidia/eval-factory/lm-evaluation-harness:{{ docker_compose_latest }}
+```
+
+or install `nemo-evaluator` and `nvidia-lm-eval` Python package in your environment of choice:
```
AWarno, line 33: q: Do I need to install both?

docs/deployment/bring-your-own-endpoint/testing-endpoint-oai-compatibility.md
marta-sd: it's not possible here - we have payload wrapped in single quotes, that's why it's not parametrized (in all examples).

and using double quotes would require escaping a lot and make the example hard to read.

do you have any ideas how to make it parametrized and keep readable?

packages/nemo-evaluator-launcher/examples/local_vllm_llama_3_1_8b_logprobs.yaml
```diff
@@ -0,0 +1,46 @@
...
+defaults:
+  - execution: local
+  - deployment: vllm
+  - _self_
+
+execution:
+  output_dir: llama_local
+  env_vars:
+    HF_TOKEN: ${oc.env:HF_TOKEN}  # needed to access meta-llama/Llama-3.1-8B gated model
+
```
AWarno: do you think we should unify it? all over the examples we pass it differently e.g.: https://github.com/NVIDIA-NeMo/Evaluator/blob/75db6f101aa5afd4242bac829cdb693105ddd274/packages/nemo-evaluator-launcher/examples/slurm_llama_3_1_8b_instruct_hf.yaml#L46

docs/evaluation/_snippets/polemo2.py
marta-sd: Please take a look at 8e12ecb

docs/deployment/bring-your-own-endpoint/testing-endpoint-oai-compatibility.md
AWarno: I am using:

```
URL=...
API_KEY=...
MODEL=...

 curl "$URL" \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer $API_KEY" \
  -d '{
    "messages": [
      {
        "role": "user",
        "content": "Write Python code that can add a list of numbers together."
      }
    ],
    "model": "'"$MODEL"'",
    "temperature": 0.00001,
    "max_tokens": 100,
    "stream": false
  }'

```

docs/evaluation/run-evals/index.md
marta-sd: Added in 329219f

docs/evaluation/custom-task.md
marta-sd: Actually no, `nvidia-lm-eval` would pull `nemo-evaluator` as its dependency, but we list evaluator in all the docs to make it clear for the users which package they are working with.

packages/nemo-evaluator-launcher/examples/local_vllm_llama_3_1_8b_logprobs.yaml
marta-sd: Good point, let me modify my example. Btw I'll be reviewing all examples in a followup PR.

AWarno: I think yours my be better (than the one above)

docs/deployment/bring-your-own-endpoint/testing-endpoint-oai-compatibility.md
marta-sd: We discussed offline and updated snippet in 90b8037 and https://github.com/NVIDIA-NeMo/Evaluator/pull/482

docs/evaluation/run-evals/index.md
```diff
@@ -1,59 +1,21 @@
...
 ## Before You Start
 
 Ensure you have:
 
 1. Completed the initial getting started guides for {ref}`gs-install` and {ref}`gs-quickstart`.
-2. Chosen a [Model Deployment](deployment-overview) option:
-   - {ref}`launcher-orchestrated-deployment` (recommended)
-   - {ref}`bring-your-own-endpoint`
-3. Reviewed the [evaluation parameters](eval-parameters) available for optimization.
+2. Have your endpoint and API key ready or prepared the checkpoint you wish to deploy.
```
athitten:
```suggestion
2. Have your endpoint and API key ready or prepared for the checkpoint you wish to deploy.
```

docs/evaluation/_snippets/polemo2.py
athitten: Thanks! To clarify for my understanding, if its part of FDF then the default params for that task are well tuned by nemo-evaluator and users can rely on it ? if not part of FDF, the params may not be well tuned for their use case and they have look into it. Is it correct ?

docs/evaluation/run-evals/text-gen.md
athitten: We had this in the initial nemo fw eval docs right ? Atleast the part where users have to explicitly say harness-name.task-name if they have multiple harnesses installed. Do we have this info somewhere else ? If not might be good to add in a follow up.

athitten (dismissed):
Thank you @marta-sd approving. There's one suggestion I made, can be merged once thats added.

docs/evaluation/_snippets/polemo2.py
marta-sd: Yes, correct

docs/evaluation/run-evals/text-gen.md
marta-sd: Yes, you're right, it's not mentioned anywhere now. Actually I think this should be part of this page: `evaluation/nemo-fw-optional-package.html`. I'll add it there.

marta-sd: Added in [0e0b70f](https://github.com/NVIDIA-NeMo/Evaluator/pull/474/commits/0e0b70f5b151474051dc41fcb19c66f4a6541eb2)
image

PR #480, opened 2025-11-20
fix(log): make tail pick up all tasks [COML1KNX-1566]

agronskiy:
This PR:

1. For the local executor, implements streaming the logs from all the tasks simultaneously
2. Colored task name in front (this is similar how e.g. `docker compose logs -f` works to distinguish logs coming from multiple jobs)
3. Architecturally, streaming comes as an API of the BaseExecutor to allow quick extension to slurm and other logs by opening an ssh tunnel -- this is a quick follow up

Gif showing it 
example

## Root cause of the bug

The mechanics of the bug consists in the fact that at the expansion of * in tail -f /home/pruprakash/Documents/nmfw_tests/nemo_evaluator/results_100samples_v2/20251114_154736-11594d2dc70b6679/*/logs/stdout.log## -- which would contain sequentially created task folders like aime, mmlu_pro and livecodebench -- this expansion happens at the beginning, when only the first task folder is present. When the first task is finished, that hangs because the * expansion does not contain the next folder. See the screenshot that illlustrates the idea

image

### DIFF
--- a/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/functional.py
+++ b/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/functional.py
@@ -260,7 +260,9 @@ def get_status(ids_or_prefixes: list[str]) -> list[dict[str, Any]]:
     return results
 
 
-def stream_logs(ids_or_prefixes: Union[str, list[str]]) -> Iterator[Tuple[str, str, str]]:
+def stream_logs(
+    ids_or_prefixes: Union[str, list[str]],
+) -> Iterator[Tuple[str, str, str]]:
     """Stream logs from jobs or invocations by their IDs or invocation IDs.
 
     Args:
@@ -335,9 +337,7 @@ def stream_logs(ids_or_prefixes: Union[str, list[str]]) -> Iterator[Tuple[str, s
             # Single job or non-local executor
             for job_id in jobs_dict.keys():
                 try:
-                    yield from executor_cls.stream_logs(
-                        job_id, executor_name=executor
-                    )
+                    yield from executor_cls.stream_logs(job_id, executor_name=executor)
                 except NotImplementedError:
                     raise ValueError(
                         f"Log streaming is not yet implemented for executor '{executor}'"
--- a/packages/nemo-evaluator-launcher/tests/unit_tests/test_cli_logs.py
+++ b/packages/nemo-evaluator-launcher/tests/unit_tests/test_cli_logs.py
@@ -380,7 +380,9 @@ class TestFunctionalAPIStreamLogs:
             data={},
             config={
                 "execution": {"type": "local"},
-                "evaluation": {"tasks": [{"name": "test_task"}, {"name": "test_task2"}]},
+                "evaluation": {
+                    "tasks": [{"name": "test_task"}, {"name": "test_task2"}]
+                },
             },
         )
         jd2, base2 = prepare_local_job(jd2, with_required=True, with_optional=True)
@@ -491,7 +493,6 @@ class TestLogsCommandIntegration:
 
     def test_logs_command_with_multiple_ids(self, job_local, prepare_local_job):
         """Test logs command with multiple IDs."""
-        inv = job_local.invocation_id
         # Create a second job in a different invocation
         inv2 = "test9999"
         jd2 = JobData(


packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/functional.py
```diff
@@ -259,6 +259,70 @@ def get_status(ids_or_prefixes: list[str]) -> list[dict[str, Any]]:
...
+        Tuple[str, str, str]: Tuples of (job_id, task_name, log_line) for each log line.
+            Empty lines are yielded as empty strings.
+
+    Raises:
+        ValueError: If the executor doesn't support log streaming.
+    """
+    db = ExecutionDB()
+
+    # Determine if this is a job ID or invocation ID
+    if "." in id_or_prefix:
```
marta-sd: Maybe we could refactor it a little bit to use if-else for getting the executor and then have a single piece of code for both cases?

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/executor.py
```diff
@@ -627,6 +622,190 @@ def kill_job(job_id: str) -> None:
...
+                return
+            jobs = {id: job_data}
+
+        # Collect log file paths and metadata
+        log_files = []
+
+        for job_id, job_data in jobs.items():
+            output_dir = pathlib.Path(job_data.data.get("output_dir", ""))
+            if not output_dir:
+                continue
```
marta-sd, line 669: When output dir can be missing? Can we safely continue of should we raise an error here?

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/executor.py
```diff
@@ -627,6 +622,190 @@ def kill_job(job_id: str) -> None:
...
+                            log_info["file_handle"] = None
+                        except Exception:
+                            # Reset error flag if we successfully read again
+                            log_info["error_printed"] = False
+
+                # If no activity, sleep briefly to avoid busy waiting
+                if not any_activity:
+                    time.sleep(0.1)
+
+        except KeyboardInterrupt:
```
marta-sd, line 786: I'm wondering if there's an easy way to exit once all logs are printed.

Can we recognize that there's no activity because all jobs finished? Maybe we could check job status? If there are no new logs, and all jobs finished, we don't wait for ctrl+c and leave the loop.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/executor.py
```diff
@@ -627,6 +622,190 @@ def kill_job(job_id: str) -> None:
...
+
+        Args:
+            job_data: JobData object containing config.
+            job_id: Job ID for error reporting.
+
+        Returns:
+            Task name string.
+        """
+        config = job_data.config or {}
+        evaluation = config.get("evaluation", {}) or {}
```
marta-sd:
```suggestion
        evaluation = config.get("evaluation", {})
```

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/executor.py
```diff
@@ -627,6 +622,190 @@ def kill_job(job_id: str) -> None:
...
+        tasks = evaluation.get("tasks", []) or []
```
marta-sd:
```suggestion
        tasks = evaluation.get("tasks", [])
```

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/executor.py
```diff
@@ -627,6 +622,190 @@ def kill_job(job_id: str) -> None:
...
+
+        # Find the task that matches this job
+        # For job_id like "15b9f667.0", index is 0
+        try:
+            if "." in job_id:
+                index = int(job_id.split(".")[1])
+                if 0 <= index < len(tasks):
```
marta-sd: Should we raise an error here?

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/executor.py
agronskiy: Vibe coding :D

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/executor.py
agronskiy: It might. Dirs can be created by jobs that are not yet started and hence we kinda monitor.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/functional.py
agronskiy: I agree, put a TODO because I'd scope this together with similar issue in other `api` level functions (like `status`). Sounds ok to follow up with another PR?

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/executor.py
agronskiy: I thought of that but can't imagine. Skills do the same btw, it might be tricky. Timouts might solve it but again some jobs might wait for very long

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/executor.py
agronskiy: Done

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/executor.py
agronskiy: Done

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/executor.py
agronskiy: Cool, done!

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/functional.py
agronskiy: Done

marta-sd: Sounds good, thanks :+1:

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/executor.py
marta-sd: Ok, let's keep it as it is then. I'll think about it when I have free cycles

PR #506, opened 2025-11-28
feat(slurm): implement squeue+sacct hybrid for accurate job status

AdamRajfer:
Fixes: Jobs showing incorrect status during active execution

- Use squeue for real-time active job status (more reliable)
- Fall back to sacct for completed/historical jobs  
- Add comprehensive tests for new squeue parsing logic

## Summary by CodeRabbit

* **Improvements**
  * Improved SLURM job status monitoring with enhanced dual-source query capability for more reliable tracking of job states and lifecycle information.


packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
```diff
@@ -985,7 +985,102 @@ def _query_slurm_jobs_status(
...
     socket: str | None,
 ) -> Dict[str, str]:
-    """Query SLURM for job statuses using sacct command.
+    """Query SLURM for job statuses using squeue (for active jobs) and sacct (fallback).
+
+    This function first tries squeue which is more accurate for currently running jobs,
+    then falls back to sacct for completed/historical jobs that squeue doesn't show.
+
+    Args:
+        slurm_job_ids: List of SLURM job IDs to query.
```
fgalko-oss, line 1020: I suspect that most of the bugs are coming from the fact that we cannot reliably tell the slurm job id for a specific job. We are trying to read this from a file, but there can be some race conditions and manual restarts that can make the file to be out-of-sync from reality.

For the concrete case we discussed offline, will this fix the status?

gchlebus:
@coderabbitai full review

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
marta-sd: I don't think we should try to handle cases where a user does something manually, e.g. restarts the job.

Also I think the file with job IDs is the closest thing to the truth that we can get. If we tried to get the information from all user's jobs, we'd open a new can of worms - most folks run different things, not only evaluations, it's hard to predict what corner cases we'd hit.

PR #540, opened 2025-12-10
fix: add migrating instructions to the overrides error message

marta-sd:


packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/helpers.py
```diff
@@ -70,6 +70,47 @@ def _set_nested_optionally_overriding(
...
+    overrides:
+    config.params.temperature: 0.6
+    config.params.top_p: 0.95
+```
+
+with:
+
+```yaml
+evaluation:
+    nemo_evaluator_config:
```
agronskiy: Agenda like indent is off from my phone screen, if not dusregard

agronskiy (approved):

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/helpers.py
marta-sd: good catch, I had one more formatting error (un-escaped slashes)

PR #566, opened 2025-12-16
feat(vllm): add model_or_checkpoint param to vllm deployment

AdamRajfer:
## Summary by CodeRabbit

* **New Features**
  * Enhanced model selection logic for vLLM with improved fallback configuration options.

* **Chores**
  * Added deprecation notice for checkpoint configuration path; users should migrate to the `model_or_checkpoint` approach for future compatibility.

### DIFF
--- a/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/configs/deployment/vllm.yaml
+++ b/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/configs/deployment/vllm.yaml
@@ -15,7 +15,6 @@
 #
 type: vllm
 image: vllm/vllm-openai:latest
-checkpoint_path: ???
 served_model_name: ???
 port: 8000
 tensor_parallel_size: 8
@@ -30,7 +29,7 @@ endpoints:
   completions: /v1/completions
   health: /health
 
-command: vllm serve ${oc.select:deployment.hf_model_handle,/checkpoint}
+command: vllm serve ${oc.select:deployment.model_or_checkpoint,${oc.select:deployment.hf_model_handle,/checkpoint}}
   --tensor-parallel-size=${deployment.tensor_parallel_size}
   --pipeline-parallel-size=${deployment.pipeline_parallel_size}
   --data-parallel-size=${deployment.data_parallel_size}
--- a/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
+++ b/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
@@ -602,6 +602,14 @@ def _create_slurm_sbatch_script(
     deployment_is_unsafe = False
     if cfg.deployment.type != "none":
         if checkpoint_path := cfg.deployment.get("checkpoint_path"):
+            warnings.warn(
+                "cfg.deployment.checkpoint_path will be deprecated in future versions. "
+                "Please use cfg.deployment.model_or_checkpoint instead and set it to "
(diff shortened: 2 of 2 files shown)


packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
```diff
@@ -602,6 +602,14 @@ def _create_slurm_sbatch_script(
     deployment_is_unsafe = False
     if cfg.deployment.type != "none":
         if checkpoint_path := cfg.deployment.get("checkpoint_path"):
+            warnings.warn(
+                "cfg.deployment.checkpoint_path will be deprecated in future versions. "
+                "Please use cfg.deployment.model_or_checkpoint instead and set it to "
+                "the path to the checkpoint inside the container. Remember to add the checkpoint "
+                "to the mounts list in the execution.mounts.deployment section as well.",
```
AWarno, line 609: we usually are adding it to tests, e.g. https://github.com/NVIDIA-NeMo/Evaluator/actions/runs/20020951800/job/57407581252#step:3:4163

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/configs/deployment/vllm.yaml
```diff
@@ -15,7 +15,6 @@
 #
 type: vllm
 image: vllm/vllm-openai:latest
-checkpoint_path: ???
```
AWarno, line 18: Comment: This is a common argument now for all deployments, so we should probably preserve unification.

marta-sd:
@coderabbit review

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/configs/deployment/vllm.yaml
```diff
@@ -30,7 +29,7 @@ endpoints:
   completions: /v1/completions
   health: /health
 
-command: vllm serve ${oc.select:deployment.hf_model_handle,/checkpoint}
+command: vllm serve ${oc.select:deployment.model_or_checkpoint,${oc.select:deployment.hf_model_handle,/checkpoint}}
```
marta-sd, line 32: There are several problems with this change:
* I don't see the new field anywhere in the default config. How users are suppose to know what param they need to specify?
* SGlang has the identical logic for checkpoint/model handle selection and you haven't updated it
* what about trt-llm and nim? We still have `checkpoint_path` parameter there and the code in the executor is shared for all deployment methods. To make things worse, with the change you propose for slurm executor users will get a deprecation warning with no available path to update.
* why do we still have `hf_model_handle` here? shouldn't `model_or_checkpoint` fully replace it?

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
```diff
@@ -602,6 +602,14 @@ def _create_slurm_sbatch_script(
     deployment_is_unsafe = False
     if cfg.deployment.type != "none":
         if checkpoint_path := cfg.deployment.get("checkpoint_path"):
+            warnings.warn(
+                "cfg.deployment.checkpoint_path will be deprecated in future versions. "
```
marta-sd, line 606: This change is very problematic: you're deprecating a deployment parameter for only one executor. This breaks deployment-execution separation. Also, what's the reason for that? Why do we want to treat slurm differently? The mounts logic for `local` is identical, I don't see any reasons to diverge.

PR #572, opened 2025-12-17
fix(containers): load tasks from the overriden container if needed 

agronskiy:
This checks the actual set of tasks if the container is overriden.


AWarno (dismissed):
lgtm

PR #577, opened 2025-12-18
chore(containers): speed up verifier/updater (of IRs) and add multiarch info

agronskiy:
- adds to the harness and task IR the arch of the container
- uses it in the output of `ls tasks` (also from any container `--from URL`)
- uses it in the docs (for table https://docs-nemo-evaluator.nvidia.com/agronskiy/chore/containers/speed-up-verification-and-multiarch/__delim__/evaluation/benchmarks/catalog/index.html) and for task (https://docs-nemo-evaluator.nvidia.com/agronskiy/chore/containers/speed-up-verification-and-multiarch/__delim__/evaluation/benchmarks/catalog/all/harnesses/bigcode-evaluation-harness.html#bigcode-evaluation-harness-multiple-cpp)

### DIFF
--- a/packages/nemo-evaluator-launcher/tests/unit_tests/test_cli_ls_tasks.py
+++ b/packages/nemo-evaluator-launcher/tests/unit_tests/test_cli_ls_tasks.py
@@ -413,7 +413,7 @@ class TestLsTasksCommand:
             "nemo_evaluator_launcher.api.functional.get_tasks_list",
             return_value=malformed_data,
         ):
-            with pytest.raises(AssertionError):
+            with pytest.raises(ValueError):
                 cmd.execute()
 
     def test_column_width_distribution(self):


packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/container_metadata/loading.py
```diff
@@ -478,15 +479,89 @@ def find_file_matching_pattern_in_image_layers(
...
+            if os_name != "linux" or not arch:
+                continue
+
+            score = 100
+            if arch == pref:
+                score = 0
+            elif arch == "amd64":
+                score = 10
+            elif arch == "arm64":
+                score = 20
```
AWarno: question: what is it?

agronskiy: This is a strange way of sorting the manifests inside of the list (if container is multiarch, there's a json pointint to different containers inside). I'll ask Cursor to rewrite it less fancily. Total overshoot

PR #604, opened 2026-01-02
feat: Sidecar launcher

Glorf:


Glorf:
dsicontinued in favour of sandbox

Issue #605, opened 2026-01-02
Can I use Nemo evaluator without launcher?


hyunwoongko:
Hello! I am developing my own evaluation framework which contains verious evaluation toolkits including nemo evaluator to use them in my company. I found this example from https://catalog.ngc.nvidia.com/orgs/nvidia/teams/eval-factory/containers/ifbench?version=latest

I launched local vllm server and executed the following example, but got the following error.
Can I use nemo evaluator without nemo evalautor launcher?

```text
pip install -U nemo-evaluator nvidia-ifbench
```

```python
nemo-evaluator run_eval \
  --eval_type ifbench \
  --model_id Qwen/Qwen3-4B \
  --model_url http://localhost:8000/v1 \
  --model_type chat \
  --output_dir ./ifbench/
```

```text
[I 2026-01-03T02:50:10.326] Centralized logging configured (console only) console_level=INFO log_dir=none (NEMO_EVALUATOR_LOG_DIR not set) logger=nemo_evaluator.logging.utils
Traceback (most recent call last):
  File "/home/bc-user/miniconda3/envs/eval/bin/nemo-evaluator", line 7, in <module>
    sys.exit(run_eval())
             ^^^^^^^^^^
  File "/home/bc-user/miniconda3/envs/eval/lib/python3.12/site-packages/nemo_evaluator/core/entrypoint.py", line 212, in run_eval
    run(args)
  File "/home/bc-user/miniconda3/envs/eval/lib/python3.12/site-packages/nemo_evaluator/core/entrypoint.py", line 164, in run
    validate_configuration(run_config)
  File "/home/bc-user/miniconda3/envs/eval/lib/python3.12/site-packages/nemo_evaluator/core/input.py", line 508, in validate_configuration
    evaluation = get_evaluation(
                 ^^^^^^^^^^^^^^^
  File "/home/bc-user/miniconda3/envs/eval/lib/python3.12/site-packages/nemo_evaluator/core/input.py", line 422, in get_evaluation
    validate_params_in_command(command, merged_configuration)
  File "/home/bc-user/miniconda3/envs/eval/lib/python3.12/site-packages/nemo_evaluator/core/utils.py", line 183, in validate_params_in_command
    raise MisconfigurationError(
[...]

PR #614, opened 2026-01-07
fix(core): do not raise on usused params; provide warning

prokotg:
Fixes https://github.com/NVIDIA-NeMo/Evaluator/issues/605 But also provides a temporary execption de-escalation for cases that were not handled in entirety.


agronskiy (approved):
LGTM, thank you! Can you create a linear to not forget an re-enable it once the containers are ready @prokotg ?

PR #624, opened 2026-01-12
feat: add sandbox logic

Glorf:
## Summary by CodeRabbit

* **New Features**
  * Added sandbox backend supporting AWS ECS Fargate infrastructure with session management.
  * Enabled interactive command execution and file upload capabilities within sandboxes.
  * Integrated AWS credential validation and comprehensive error handling.

* **Documentation**
  * Added API documentation for the sandbox module and ECS Fargate usage.

* **Tests**
  * Added unit tests for sandbox module behavior and edge cases.

### DIFF
--- a/packages/nemo-evaluator/tests/unit_tests/sandbox/test_sandbox_module.py
+++ b/packages/nemo-evaluator/tests/unit_tests/sandbox/test_sandbox_module.py
@@ -136,9 +136,9 @@ def test_parse_exec_markers_extracts_payload_and_strips_noise(
         returncode=0,
         stdout=(
             "Starting session with SessionId: abc\n"
-            "__TB_BEGIN__\n"
+            "__NEMO_BEGIN__\n"
             "hello\n"
-            "__TB_RC__=0\n"
+            "__NEMO_RC__=0\n"
             "Exiting session with sessionId: abc\n"
         ),
         stderr="The Session Manager plugin was installed successfully. Use the AWS CLI to start a session.\n",
@@ -169,7 +169,7 @@ def test_parse_exec_markers_raises_when_rc_nonzero(monkeypatch: pytest.MonkeyPat
     cp = CompletedProcess(
         args=["aws", "ecs", "execute-command"],
         returncode=0,
-        stdout="__TB_BEGIN__\nboom\n__TB_RC__=7\n",
+        stdout="__NEMO_BEGIN__\nboom\n__NEMO_RC__=7\n",
         stderr="",
     )
 


Glorf:
/ok to test https://github.com/NVIDIA-NeMo/Evaluator/pull/624/commits/ad4fd4c33e725c5e94768806fe069b645c2043d1

Glorf:
@coderabbitai full review

Glorf:
/ok to test https://github.com/NVIDIA-NeMo/Evaluator/pull/624/commits/216cff66ce86e34907a4d24f8f6c7f37abca19f6

docs/references/api/nemo-evaluator/sandbox/index.rst
```diff
@@ -0,0 +1,55 @@
+``nemo_evaluator.sandbox``
```
agronskiy, line 1: nice! https://docs-nemo-evaluator.nvidia.com/mbien/sandbox/__delim__/references/api/nemo-evaluator/sandbox/index.html

packages/nemo-evaluator/src/nemo_evaluator/sandbox/base.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def send_command(self, command: NemoSandboxCommand) -> None: ...
+
+    def capture_pane(self, capture_entire: bool = False) -> str: ...
+
+    def is_session_alive(self) -> bool: ...
+
+    def get_incremental_output(self) -> str: ...
+
+    def get_asciinema_timestamp(self) -> float: ...
```
agronskiy, line 62: `non-blocking`: funny to see `asciinema` which I knew as a demo capturing app here as a serious application for tracking the tty of the sandbox

packages/nemo-evaluator/src/nemo_evaluator/sandbox/base.py
```diff
@@ -0,0 +1,115 @@
...
+    def spin_up(
+        cls,
+        *,
+        task_id: str,
+        trial_name: str,
+        run_id: str,
+        pre_upload_paths: Iterable[Path] | None = None,
+        upload_dest_dir: str | None = None,
+        **kwargs,
+    ) -> ContextManager["NemoEvaluatorSandbox"]:
```
agronskiy, line 91: `nit`: more a question to Cursor I guess (or to the Cursor rules one uses): in presense of `from __future__ import annotation` one does need to take the own class's name into literal quotes.
image
image

packages/nemo-evaluator/src/nemo_evaluator/sandbox/ecs_fargate.py
```diff
@@ -0,0 +1,1172 @@
...
+
+class AwsCliMissingError(RuntimeError):
+    pass
+
+
+class EcsExecError(RuntimeError):
+    pass
+
+
+def _require_aws_sdks():
```
agronskiy, line 85: `thinking loud`: since it's not in pyproject 1) I find the return to try-catch dependency management suboptimal -- it would need additional steps from the user (`pre_cmd`) to install that into a docker if one forgets? 2) alternative: to add it as another subgroup to pyproject -- it better describes what the evaluator has -- and keep lazy loading (not `importlib` but normal `import blah` in this function)

packages/nemo-evaluator/src/nemo_evaluator/sandbox/ecs_fargate.py
```diff
@@ -0,0 +1,1172 @@
...
+            _as_text(getattr(cp, "stdout", ""))
+            + "\n"
+            + _as_text(getattr(cp, "stderr", ""))
+        ).splitlines()
+
+        filtered_lines: list[str] = []
+        for line in combined_lines:
+            if line.startswith("The Session Manager plugin was installed successfully"):
+                continue
+            if line.startswith("Starting session with SessionId:"):
```
agronskiy: `question`: is it intentional that one `startswith` has capitalized `SessionId` and the one below has lowercased -- does tmux plugin emit exactly that or it's a bug?

packages/nemo-evaluator/src/nemo_evaluator/sandbox/ecs_fargate.py
```diff
@@ -0,0 +1,1172 @@
...
+        self._s = sandbox
+        self.attrs = {"Config": {"User": ""}}
+
+    def exec_run(self, cmd: list[str], user: str = ""):
+        _ = user  # ECS Exec user switching is handled by remote shell / sudo if needed.
+        out = self._s._exec_capture(cmd=cmd, timeout_sec=180.0, check=True)
+        return type("ExecResult", (), {"exit_code": 0, "output": out.encode()})
+
+
+class EcsFargateTmuxSession(NemoSandboxSession):
```
agronskiy, line 160: `suggestion`: I would really add tmux version check and either emite a huge warning if we're beyond the current tmux version + 1 to be on the safe side. Since tmux `3.0` there was a good deal of development every minor and I see below that cursor relied on parsing number of emitted things that might break any time. Minor version updates of tmux don't guarantee backward compat of the emitted logs and magic strings

agronskiy (approved):
Amzaing, some questions and nits but unblocking. General request: ask cursor to add small (here, it does not pay off to have detailed ones) docstring to each tricky function just becasue it will make it simpler for human to skim. Or even a simple ascii diagram.

packages/nemo-evaluator/src/nemo_evaluator/sandbox/ecs_fargate.py
Glorf: That's a tricky one - I can add awscli to pyproject, yet sessionmanager needs to be handled separately in a dockerfile anyways :( : https://docs.aws.amazon.com/systems-manager/latest/userguide/session-manager-working-with-install-plugin.html. For now it seemed like the best compromise we have

packages/nemo-evaluator/src/nemo_evaluator/sandbox/base.py
Glorf: +1, research mode on :D

packages/nemo-evaluator/src/nemo_evaluator/sandbox/ecs_fargate.py
Glorf: I'm not super familiar with tmux development, do you have an intuition which version should we stick to? These will also depend on what tmux version the container provides so I'm a little uneasy with forcing a specific version if this tooling is expected to be generic

packages/nemo-evaluator/src/nemo_evaluator/sandbox/ecs_fargate.py
Glorf: Thanks for finding! Fixed now (it's lowercased so we don't make it case-sensitive)

packages/nemo-evaluator/src/nemo_evaluator/sandbox/ecs_fargate.py
agronskiy: I would just research which one is working in our env and add a check on `tmux version`, issuing a warning into logs that the guaranteed one is blah and blah.

Glorf: added!

Glorf:
/ok to test https://github.com/NVIDIA-NeMo/Evaluator/pull/624/commits/1e31705d66568336d5382a949e5539a33ada057d

Glorf:
/ok to test https://github.com/NVIDIA-NeMo/Evaluator/pull/624/commits/efae04744279e1fec7f7123a0113d1297f771d74

PR #561, opened 2025-12-16
fix(nemo-evaluator): api_key -> api_key_name and content=None handling

marta-sd:
## Summary by CodeRabbit

* **New Features**
  * Introduced new `api_key_name` configuration parameter for specifying API key environment variables; `api_key` is now deprecated with automatic migration and backward compatibility warnings.

* **Bug Fixes**
  * Enhanced upstream response processing to detect and replace null message content values with empty strings; added robust error handling for JSON parsing failures.

* **Tests**
  * Added comprehensive unit test coverage for API key deprecation workflows and response content handling across various response structures.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #633, opened 2026-01-14
feat: add api_key_value field for literal API key support

gchlebus:
The api_key/api_key_name fields were designed to hold environment variable names, not literal API keys. Users passing actual API keys directly would get 401 errors because the code tried to resolve them as env var names.

This adds api_key_value to ApiEndpoint and EndpointModelConfig that accepts a literal API key string directly. The two approaches are mutually exclusive with validation to prevent setting both.

## Summary by CodeRabbit

* **New Features**
  * Added option to provide API keys directly as literal values instead of only through environment variables.
  * Enhanced API key resolution with smart fallback: uses literal keys first, then environment variables, then defaults.
  * Added validation to prevent conflicting API key configurations.

* **Tests**
  * Expanded test coverage for API key configuration scenarios.

### DIFF
--- a/packages/nemo-evaluator/src/nemo_evaluator/api/api_dataclasses.py
+++ b/packages/nemo-evaluator/src/nemo_evaluator/api/api_dataclasses.py
@@ -54,6 +54,10 @@ class ApiEndpoint(BaseModel):
         description="Name of the environment variable that stores API key for the model",
         default=None,
     )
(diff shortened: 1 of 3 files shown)


gchlebus:
@coderabbitai reivew

marta-sd:
The problem was addressed in #561 by renaming the field

PR #638, opened 2026-01-16
feat: Add optional max_walltime to prevent infinite looping in Slurm jobs

wprazuch:


packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
```diff
@@ -1336,9 +1337,77 @@ def _get_progress(
...
+
+# Record start time on first run (when no previous job ID exists)
+if [[ ! -f "$_start_time_file" ]]; then
+    date +%s > "$_start_time_file"
+    echo "Job chain started at $(date). Max total walltime: $_max_walltime"
+fi
+
+# Read start time and calculate elapsed time
+_job_chain_start_time=$(cat "$_start_time_file")
+_current_time=$(date +%s)
```
marta-sd: If I read this correctly, we're not calculating the right thing here. It doesn't calculate walltime, but a total time that has passed since the job was started, including the time that the resumed job has spent in the queue

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/configs/execution/slurm/default.yaml
```diff
@@ -24,6 +24,7 @@ num_nodes: 1
 ntasks_per_node: 1
 gres: gpu:8
 walltime: 01:00:00
+max_walltime: null       # Maximum total wall-clock time across all resumes (e.g., "24:00:00"). null = unlimited.
```
marta-sd: We should consider setting it to some high default (eg 48h). Users pointed out that the current setup is error prone.

I think that is the user forgets to increase it will cause less damage than forgetting to set it.

wprazuch: Can we agree on 72h as default? Some benchmarks can really last 48h, I would like to minimize the risk of us not delivering on time because someone forgot to increase max_walltime :(

marta-sd: 72 sounds good, I just want something < infinity :)

wprazuch:
/ok to test [c70c92c](https://github.com/NVIDIA-NeMo/Evaluator/pull/638/commits/c70c92c4eec43e0ddc2ee4ae86e4937da5eeb393)

wprazuch:
/ok to test [195a477](https://github.com/NVIDIA-NeMo/Evaluator/pull/638/commits/195a477dd02a6da65a182fe408c2a772f7315b74)

wprazuch:
/ok to test [d95b95f](https://github.com/NVIDIA-NeMo/Evaluator/pull/638/commits/d95b95f00cd7ff67a00791cee6c386118b55ca62)

wprazuch:
/ok to test [b7706d3](https://github.com/NVIDIA-NeMo/Evaluator/pull/638/commits/b7706d368bd5bbb1d73b87c28e593d328f8a79e7)

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/configs/execution/slurm/default.yaml
wprazuch: Fixed

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
wprazuch: Good catch, fixed that with sacct :)

wprazuch:
/ok to test [c9f6886](https://github.com/NVIDIA-NeMo/Evaluator/pull/638/commits/c9f6886dc186f48c1f0b7648514f0f240cb09ad3)

wprazuch:
/ok to test [a60f8c8](https://github.com/NVIDIA-NeMo/Evaluator/pull/638/commits/a60f8c855addc02250e5c7845431911d3fe32c0e)

marta-sd (dismissed):

wprazuch:
/ok to test [8e32ca7](https://github.com/NVIDIA-NeMo/Evaluator/pull/638/commits/8e32ca717820d1049b04b427b46b9a381562105c)

wprazuch:
/ok to test [32ec579](https://github.com/NVIDIA-NeMo/Evaluator/pull/638/commits/32ec57962a48e8baa76a20073fcf235e17ecaacd)

marta-sd (approved):

PR #645, opened 2026-01-19
ci: Add `release-docs` workflow

ko3n1g:
This workflow will be launched manually via the `Actions` tab - it is called `Release docs`.

When launching it, we can select the branch of the docs we want to deploy via the dropdown menu. When deploying `main`, we will upload the docs to the `latest` directory of the webserver. If it's a versioned tag, it will strip off the version prefix `v` and upload it into the version folder of the webserver (e.g. `0.2.0`).

### DIFF
--- a/.github/workflows/release-docs.yml
+++ b/.github/workflows/release-docs.yml
@@ -73,4 +73,4 @@ jobs:
           akamai-client-secret: ${{ secrets.AKAMAI_CLIENT_SECRET }}
           akamai-access-token: ${{ secrets.AKAMAI_ACCESS_TOKEN }}
           s3-target-root: ${{ secrets.S3_BUCKET_NAME }}
-          s3-target-path: nemo/megatron-bridge
+          s3-target-path: nemo/evaluator


.github/workflows/release-docs.yml
```diff
@@ -0,0 +1,76 @@
...
+          aws-region: ${{ inputs.aws-region }}
+          aws-role-to-assume: ${{ secrets.AWS_ASSUME_ROLE_ARN }}
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          akamai-host: ${{ secrets.AKAMAI_HOST }}
+          akamai-client-token: ${{ secrets.AKAMAI_CLIENT_TOKEN }}
+          akamai-client-secret: ${{ secrets.AKAMAI_CLIENT_SECRET }}
+          akamai-access-token: ${{ secrets.AKAMAI_ACCESS_TOKEN }}
+          s3-target-root: ${{ secrets.S3_BUCKET_NAME }}
+          s3-target-path: nemo/megatron-bridge
```
marta-sd: @ko3n1g I think this is copy-paste bug
```suggestion
          s3-target-path: nemo/evaluator
```

.github/workflows/release-docs.yml
```diff
@@ -0,0 +1,76 @@
...
+      requirements-file: .github/config/requirements-docs.txt
+
+  publish-docs:
+    runs-on: ubuntu-latest
+    needs: [build-docs]
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          repository: NVIDIA-NeMo/FW-CI-templates
+          ref: v0.67.2
```
agronskiy, line 50: `nit`: differs from `@0.67.0` above, I know we did have patch number matter in the past

.github/workflows/release-docs.yml
```diff
@@ -0,0 +1,76 @@
...
+        with:
+          dry-run: ${{ inputs.dry-run }}
+          artifacts-name: docs-html
+          artifacts-path: _build/html
+          emails-csv: ${{ inputs.notify-emails }}
+          overwrite-latest-on-tag: false
+          run-on-version-tag-only: ${{ github.ref_name != 'main' }}
+          request-name: megatron-bridge-publish-docs-${{ github.run_id }}
+          aws-region: ${{ inputs.aws-region }}
+          aws-role-to-assume: ${{ secrets.AWS_ASSUME_ROLE_ARN }}
```
agronskiy, line 68: `question`: is there a central place to read up the rotation info (purpose/owner) of all those keys? I assume you guys do the rotation across many repos or whatever but we could be useful to you too, proactively reaching out if anything starts failing

.github/workflows/release-docs.yml
ko3n1g: it's fine in this case, the docs-build workflow is stable and doesnt need the patch

PR #658, opened 2026-01-26
feat: Hide internal datasets in envs where not supported

Glorf:
## Summary by CodeRabbit

* **New Features**
  * Added detection of internal evaluation packages and conditional inclusion of internal evaluations based on package availability.
  * Framework evaluations now dynamically filter based on whether internal packages are installed.

### DIFF
--- a/packages/nemo-evaluator/src/nemo_evaluator/package_info.py
+++ b/packages/nemo-evaluator/src/nemo_evaluator/package_info.py
@@ -16,7 +16,7 @@
 # Below is the _next_ version that will be published, not the currently published one.
 MAJOR = 0
 MINOR = 1
-PATCH = 72
+PATCH = 73
 PRE_RELEASE = ""
 
 # Use the following formatting: (major, minor, patch, pre-release)


Glorf:
@coderabbit full review

Glorf:
/ok to test https://github.com/NVIDIA-NeMo/Evaluator/pull/658/commits/08a5f63b554c1f6de3306a9fe055c9ac74c49f66

packages/nemo-evaluator/src/nemo_evaluator/core/input.py
```diff
@@ -124,6 +131,8 @@ def get_framework_evaluations(filepath: str) -> tuple[str, dict, dict[str, Evalu
     run_config_framework_defaults["pkg_name"] = pkg_name
     evaluations = dict()
     for evaluation_dict in framework["evaluations"]:
+        if evaluation_dict.get("internal", False) and not include_internal:
```
marta-sd, line 134: double checking: we don't risk any pydantic errors because we work with raw data from `framework.yml` and we never pass `internal` to any of our dataclasses?

packages/nemo-evaluator/src/nemo_evaluator/core/entrypoint.py
```diff
@@ -136,11 +137,15 @@ def show_available_tasks() -> None:
...
 
+    include_internal = _is_internal_package_installed()
     for pkg in core_evals_pkg:
         framework_eval_mapping, *_ = _get_framework_evaluations(
-            os.path.join(pkg.module_finder.path, pkg.name, "framework.yml")
+            os.path.join(pkg.module_finder.path, pkg.name, "framework.yml"),
+            include_internal=include_internal,
         )
         for ind_pkg in framework_eval_mapping.keys():
+            if not framework_eval_mapping[ind_pkg]:
```
marta-sd, line 148: what exactly we have here? empty dict? please add a comment explaining what value we have here and that it happens for internal evals

packages/nemo-evaluator/src/nemo_evaluator/core/input.py
Glorf: We only read internal from the raw framework.yml dict to decide whether to skip that entry; we never pass it into Evaluation. The constructor receives only evaluation_dict["defaults"] merged into framework defaults, so internal is not part of the payload. That avoids any Pydantic extra="forbid" errors on Evaluation.

packages/nemo-evaluator/src/nemo_evaluator/core/entrypoint.py
Glorf: Added!

Glorf:
/ok to test https://github.com/NVIDIA-NeMo/Evaluator/pull/658/commits/e6e015884b1310c0562af318454584e4b3038c77

Glorf:
/ok to test https://github.com/NVIDIA-NeMo/Evaluator/pull/658/commits/bf806ac38bf3c04b29722592d3e63227b7497038

PR #663, opened 2026-01-27
feat: Support unlisted tasks in nel and add safeguard

gchlebus:
Add support for tasks not defined in Framework Definition Files (FDF) with a safeguard mechanism following the NEMO_EVALUATOR_TRUST_PRE_CMD pattern.

Changes:
- mapping.py: Refactor get_task_definition_for_job() to support two workflows: with explicit container (validates harness match) and without container (derives container from harness). Add is_unlisted flag and task_query to task definitions for unlisted tasks.
- helpers.py: Use full harness.task format in config.type for unlisted tasks. Add check_unlisted_tasks_safeguard() function that warns in dry-run and requires NEMO_EVALUATOR_TRUST_UNLISTED_TASKS=1 otherwise.
- All executors (slurm, local, lepton): Track unlisted tasks during task iteration and call the safeguard check.
- Tests: Add comprehensive tests for new functionality.
- endpoint_type field in the task specification, which allows user to select which endpoint should be used for unlisted task or for tasks with multiple supported types

### DIFF
--- a/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/mapping.py
+++ b/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/mapping.py
@@ -236,18 +236,24 @@ def get_task_from_mapping(query: str, mapping: dict[Any, Any]) -> dict[Any, Any]
         )
 
 
-def _minimal_task_definition(task_query: str, *, container: str) -> dict[str, Any]:
+def _minimal_task_definition(
+    task_query: str,
+    *,
+    container: str,
+    endpoint_type: str | None = None,
+) -> dict[str, Any]:
     """Create a minimal task definition when task is not known in any mapping.
 
     Args:
         task_query: The original query string (e.g., "lm-evaluation-harness.polemo2").
         container: The container image to use.
+        endpoint_type: The endpoint type to use (defaults to 'chat' if not provided).
 
     Returns:
         A minimal task definition dict with:
         - 'task': Task name
         - 'harness': Harness name
-        - 'endpoint_type': 'chat' (default)
+        - 'endpoint_type': user-provided endpoint type or default 'chat'
         - 'container': Container image
         - 'is_unlisted': True (task not in FDF)
     """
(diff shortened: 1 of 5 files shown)


packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/mapping.py
```diff
@@ -249,6 +259,8 @@ def _minimal_task_definition(task_query: str, *, container: str) -> dict[str, An
         "harness": harness,
         "endpoint_type": "chat",
         "container": container,
+        "is_unlisted": True,
+        "task_query": task_query,  # Preserve original query for EF command
```
marta-sd: I don't think this is needed - we use `task_type = f"{harness}.{task}"` when building the command.

marta-sd:
I verified it works with the following config:

```
defaults:
  - execution: local
  - deployment: vllm
  - _self_

execution:
  output_dir: llama_local
  env_vars:
    deployment:
      HF_TOKEN: ${oc.env:HF_TOKEN}  # needed to access meta-llama/Llama-3.1-8B gated model

# specify deployment arguments
deployment:
  checkpoint_path: null
  hf_model_handle: meta-llama/Llama-3.2-1B
  served_model_name: meta-llama/Llama-3.2-1B
  tensor_parallel_size: 1
  data_parallel_size: 1
  extra_args: "--max-model-len 32768"

# specify the benchmarks to evaluate
evaluation:
  # global config settings that apply to all tasks, unless overridden by task-specific config
  nemo_evaluator_config:
    config:
      params:
        request_timeout: 3600  # timeout for API request in seconds
        parallelism: 1  # 1 parallel request to avoid overloading the server
        limit_samples: 10 # uncomment to limit number of samples for quick testing
        extra:  # for log-probability tasks like piqa, you need to specify the tokenizer
          tokenizer: meta-llama/Llama-3.2-1B  # or use a path to locally stored checkpoint
          tokenizer_backend: huggingface      # or "tiktoken"
  env_vars:
      HF_TOKEN: HF_TOKEN  # needed to access the tokenizer on the client side
  tasks:
    - name: lm-evaluation-harness.polemo2
      endpoint_type: completions
 ```

PR #671, opened 2026-01-28
fix(slurm): node_array undefined

agronskiy:
## Summary

When running the launcher on Slurm with `deployment.type: none`, the generated sbatch script could fail at runtime with:

- `line N: nodes_array[0]: unbound variable`

This was triggered by `set -u` (nounset) and an unconditional `--nodelist ${nodes_array[0]}` in the evaluation client `srun`.

## Impact

- **Configs affected**: any Slurm run with `deployment.type=none` (e.g., “target-only” evaluation).
- **Failure mode**: sbatch script exits before launching the evaluation client.
- **Where observed**: Slurm job log (`slurm_script` / `slurm-%A.log`).

## Direct cause

- The sbatch script enables:
  - `set -u` (treat unset variables as an error)
- The evaluation client `srun` was emitted as:
  - `srun ... --nodelist ${nodes_array[0]} ...`
- `nodes_array` was only defined inside the deployment block (`if cfg.deployment.type != "none": ...`).
- Therefore, for `deployment.type=none`, `nodes_array` was undefined and `${nodes_array[0]}` crashed under nounset.

## Secondary risks (also addressed)

Even when deployment is enabled, `${nodes_array[0]}` can still fail if:

- `$SLURM_JOB_NODELIST` is unset/empty (non-standard environment) or only `$SLURM_NODELIST` is present.
- `scontrol` is unavailable on the node or not in `PATH`.
- `scontrol show hostnames ...` returns an empty list.

Any of these can result in an empty/unset array index under `set -u`.

### Approach

Introduce a **single, always-defined** “node pinning” variable for single-node sruns:

- `PRIMARY_NODE`

This is resolved at runtime in the sbatch script with safe fallbacks:

1. Prefer `SLURM_JOB_NODELIST`
2. Fallback to `SLURM_NODELIST`
3. Fallback to local `hostname`

### DIFF
--- a/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/container_metadata/registries.py
+++ b/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/container_metadata/registries.py
@@ -27,6 +27,7 @@ import requests
 
 from nemo_evaluator_launcher.common.logging_utils import logger
 
+
 def _get_docker_config_path() -> pathlib.Path:
     """Return the effective Docker `config.json` path.
 
@@ -45,6 +46,7 @@ def _first_env_set(*names: str) -> tuple[Optional[str], Optional[str]]:
         if v:
             return v, n
     return None, None
+
 
 # Docker Registry API v2 manifest Accept header.
 # IMPORTANT: include *manifest list* / *OCI index* types so multi-arch tags return
--- a/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
+++ b/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
@@ -622,6 +622,25 @@ def _create_slurm_sbatch_script(
     s += "set -x  # print commands and their arguments as they are executed\n"
     s += "\n"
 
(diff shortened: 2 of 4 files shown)


packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/container_metadata/registries.py
```diff
@@ -185,6 +204,18 @@ def _read_docker_credentials(registry_url: str) -> Optional[Tuple[str, str]]:
...
+            # does not invoke docker credential helpers.
+            if config.get("credsStore") or config.get("credHelpers"):
+                logger.debug(
+                    "Docker config uses credential helpers; inline auth missing",
+                    registry_url=registry_url,
+                    matched_key=matched_key or registry_url,
+                    creds_store=config.get("credsStore"),
+                    has_cred_helpers=bool(config.get("credHelpers")),
+                    config_path=str(docker_config_path),
+                )
```
AWarno, line 218: q: is it also part of this fix?

agronskiy: No, I sneaked this into from a simultaneous issue :D

PR #703, opened 2026-02-06
feat(env-vars): initial reimplementation of the env vars rfc

agronskiy:
# Unified env var handling with `.secrets.env` pipeline

This PR replaces the fragmented environment variable handling across executors with a unified, explicit prefix-based system. Previously, `execution.env_vars` values were written literally while `evaluation.env_vars` values were silently resolved via `os.getenv()` — leading to duplicate exports, secrets leaking into generated scripts, and cross-job env contamination on SLURM. Now, all env var values use explicit prefixes (`host:`, `lit:`, `runtime:`) and are routed through a shared `.secrets.env` pipeline: secrets are written to a separate file with disambiguated names (`VAR_<token>_<group>`), sourced at runtime, and re-exported to their original names right before each container command. Generated scripts (`run.sh`, `run.sub`) never contain plaintext secret values.

A new top-level `env_vars:` config section provides a single place to declare variables inherited by both deployment and evaluation, with section-level and task-level overrides. The CLI gains an `--env-file` flag (defaulting to `$PWD/.env`) that loads variables into the host environment before config resolution — making `$host:` references work seamlessly with dotenv workflows. Loaded keys are logged with redacted values, and shadowed keys produce a warning.

## Config example — before and after

```yaml
# BEFORE (no longer supported):
execution:
  env_vars:
    evaluation:
      HF_TOKEN: $HF_TOKEN        # ambiguous: literal? env ref?
    deployment:
      CACHE_DIR: /cache/path
evaluation:
  env_vars:
    HF_TOKEN: HF_TOKEN           # silently resolved via os.getenv()

# AFTER:
env_vars:
  HF_TOKEN: host:HF_TOKEN           # resolved from host env / .env file
  CACHE_DIR: lit:/cache/path         # literal, written as-is
  JOB_ID: runtime:SLURM_JOB_ID      # late-bound at execution time

deployment:
  env_vars:
    DEPLOY_ONLY: lit:some_value      # deployment-specific (not deprecated)

evaluation:
  tasks:
    - name: my_task
      env_vars:
        EXTRA: lit:task-specific     # task-level override
```

## Migration guide

1. **Automated migration**: Run `nel-migrate-config` (installed with the package) to add explicit prefixes and relocate `execution.env_vars` entries:
   ```bash
   # Preview changes (dry-run):
   nel-migrate-config your_config.yaml

   # Overwrite in place:
   nel-migrate-config your_config.yaml --write

   # Write to a new file:
   nel-migrate-config your_config.yaml -o migrated.yaml
   ```
   The script classifies bare values using heuristics: `UPPER_CASE` names become `host:`, paths/URLs/numbers become `lit:`, `$VAR` references become `host:VAR`, and `${oc.env:VAR}` Hydra resolvers become `host:VAR`. The old `execution.env_vars.deployment` / `execution.env_vars.evaluation` blocks are relocated to `deployment.env_vars` and top-level `env_vars` respectively. Non-migrable values like `${oc.decode:...}` are flagged in red for manual review.

2. **`execution.env_vars.evaluation` / `execution.env_vars.deployment`** are removed. Use top-level `env_vars:` for shared variables, `deployment.env_vars` for deployment-specific ones, and `evaluation.env_vars` / per-task `env_vars` for evaluation scoping.

3. **Optional: use `--env-file`** to load secrets from a `.env` file instead of exporting them in your shell. The launcher loads `$PWD/.env` by default if present, or pass `--env-file path/to/.env` explicitly.

### DIFF
--- a/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/configs/execution/slurm/default.yaml
+++ b/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/configs/execution/slurm/default.yaml
@@ -14,11 +14,11 @@
 # limitations under the License.
 #
 # Each slurm cluster has its own flavour, below we provide some defaults that might meet one's needs.
-type: slurm              # Executor is chosen based on this field
-hostname: ???            # SLURM headnode (login) hostname (required)
+type: slurm # Executor is chosen based on this field
+hostname: ??? # SLURM headnode (login) hostname (required)
 username: ${oc.env:USER} # Defaults to $USER env var
-account: ???             # SLURM account allocation (required)
-output_dir: ???          # Absolute path accessible on compute nodes (required)
+account: ??? # SLURM account allocation (required)
+output_dir: ??? # Absolute path accessible on compute nodes (required)
 partition: batch
 num_nodes: 1
 ntasks_per_node: 1
@@ -26,15 +26,12 @@ gres: gpu:8
 walltime: 01:00:00
 max_walltime: "120:00:00" # Maximum total runtime across all resumes. Set to null for unlimited.
 subproject: nemo-evaluator-launcher
-sbatch_comment: null  # Optional comment for SLURM job (translates to #SBATCH --comment='...')
+sbatch_comment: null # Optional comment for SLURM job (translates to #SBATCH --comment='...')
 
 # Deployment-specific SLURM configuration
 deployment:
-  n_tasks: 1             # Number of tasks for deployment srun (default: 1, for multi-instance set to num_nodes)
+  n_tasks: 1 # Number of tasks for deployment srun (default: 1, for multi-instance set to num_nodes)
 
-env_vars:
-  deployment: {}
-  evaluation: {}
 mounts:
   deployment: {}
   evaluation: {}


packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/env_vars.py
```diff
@@ -0,0 +1,487 @@
...
+    runtime_var_name: str
+
+
+EnvVarValue = EnvVarLiteral | EnvVarFromHost | EnvVarRuntime
+
+# Pattern matching a valid env var name (used for backward-compat heuristic)
+_ENV_VAR_NAME_RE = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*$")
+
+
+def parse_env_var_value(raw: str, default_type: str = "host") -> EnvVarValue:
```
marta-sd: suggestion: maybe we could allow for specifying `default_type=None` and then throwing an error in L110-L125 if it's neither host or lit? We could use it for cases where we explicitly want to force the new api

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/env_vars.py
```diff
@@ -0,0 +1,487 @@
...
+            "Move these variables to top-level env_vars or deployment.env_vars instead.",
+            category=DeprecationWarning,
+            stacklevel=2,
+        )
+        for target_name, raw_value in exec_deploy_vars.items():
+            parsed[target_name] = parse_env_var_value(
+                str(raw_value), default_type="lit"
+            )
+
+    # 3. Deprecated: cfg.deployment.env_vars — uses literal default
```
marta-sd: as discussed offline, we don't want to deprecate this path

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/run.template.sh
```diff
@@ -59,13 +59,22 @@ else
...
+        {% endif -%}
+
         echo "$(date -u +%Y-%m-%dT%H:%M:%SZ)" > "$logs_dir/stage.running"
         {% if task.deployment %}
         docker run --rm --shm-size=100g --gpus all {{ task.deployment.extra_docker_args }} \
         --name {{ task.deployment.container_name }} --entrypoint '' \
         -p {{ task.deployment.port }}:{{ task.deployment.port }} \
-        {% for env_var in task.deployment.env_vars -%}
-        -e {{ env_var }} \
+        {% for var_name in task.deployment.env_var_names -%}
```
marta-sd, line 76: I think we currently lack similar logic for slurm. If I read the code correctly, we don't pass env vars from top-level config, but only from `execution` and `deployment` ([here](https://github.com/NVIDIA-NeMo/Evaluator/blob/00a85b7554721d957cb330dd15086044a3d70ccf/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py#L1660))

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/run.template.sh
```diff
@@ -59,13 +59,22 @@ else
 
     # Docker run with eval factory command
     (
+        {% if task.secrets_env_content -%}
+        # Source secrets and re-export to original names (scoped to subshell)
+        source "$task_dir/.secrets.env"
+        {{ task.eval_reexport_cmd }}
```
marta-sd: I think we should do it directly before starting the eval container. Otherwise we override with `task.deployment_reexport_cmd` in the next line

agronskiy: Good catch, thanks, I'm refactoring this part now!

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/env_vars.py
agronskiy: Good point, changed

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/env_vars.py
agronskiy: This is a good catch, I changed and eyeballed it everywhere

packages/nemo-evaluator-launcher/scripts/migrate_config.py
```diff
@@ -0,0 +1,565 @@
...
+  - ${oc.env:VAR} Hydra resolvers → $host:VAR
+  - ${oc.decode:...} Hydra resolvers → listed as non-migrated (needs manual review)
+  - $VAR_NAME (dollar-prefixed) → $host:VAR_NAME
+  - Bare UPPER_CASE name matching [A-Z_][A-Z0-9_]* → $host:NAME
+  - Numbers (0, 1, etc.) → $lit:value
+  - Everything else (paths, URLs, mixed-case, spaces) → $lit:value
+
+Structural migration:
+  - execution.env_vars.deployment → deployment.env_vars
+  - execution.env_vars.evaluation → top-level env_vars
```
marta-sd: I think they should be migrated to `evaluation.env_vars` instead of top level. This way we'll be closer to the current behavior (ie we won't use those vars for deployment job)

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/run.template.sh
agronskiy: Yes this is a but MArta, changed now!

packages/nemo-evaluator-launcher/scripts/migrate_config.py
agronskiy: Replied in another thread, let's continue there to reduce redundancy

docs/libraries/nemo-evaluator-launcher/configuration/evaluation/index.md
```diff
@@ -87,21 +87,84 @@ evaluation:
...
+  HF_TOKEN: $host:HF_TOKEN
+  CACHE_DIR: $lit:/cache/huggingface
+
+# 2. Evaluation-level — applies to all evaluation tasks
+evaluation:
+  env_vars:
+    CUSTOM_VAR: $lit:some_value
+
+# 3. Task-level — applies to a single task only
 evaluation:
```
marta-sd: repeated key?
```suggestion
```

packages/nemo-evaluator-launcher/examples/slurm_vllm_advanced_hf_caching.yaml
```diff
@@ -57,26 +62,16 @@ execution:
...
-# the model will be downloaded from Hugging Face during deployment
-# and stored in the HF home directory for later use
 deployment:
   checkpoint_path: null
   hf_model_handle: meta-llama/Llama-3.1-8B-Instruct
   served_model_name: meta-llama/Llama-3.1-8B-Instruct
   tensor_parallel_size: 1
   data_parallel_size: 8
   extra_args: "--max-model-len 32768"
+  env_vars:
```
marta-sd: I think we can remove it (defined at top level)

packages/nemo-evaluator-launcher/examples/slurm_vllm_advanced_hf_caching.yaml
```diff
@@ -57,26 +62,16 @@ execution:
...
-    # if the checkpoint and datasets are already cached, you can set
-    # HF_DATASETS_OFFLINE=1 and TRANSFORMERS_OFFLINE=1 for offline mode
-    deployment:
-      HF_DATASETS_OFFLINE: 0
-      TRANSFORMERS_OFFLINE: 0
-    evaluation:
-      HF_DATASETS_OFFLINE: 0
-      TRANSFORMERS_OFFLINE: 0
-
-# set up deployment as usual
```
marta-sd, line 70: are these comments removed intentionally?

packages/nemo-evaluator-launcher/examples/slurm_vllm_advanced_hf_caching.yaml
```diff
@@ -57,26 +62,16 @@ execution:
       /path/to/hf_home: /root/.cache/huggingface
     mount_home: false   # don't mount home directory
 
-  env_vars:
-    # if the checkpoint and datasets are already cached, you can set
```
marta-sd, line 61: please move these comments to `env_vars` section

packages/nemo-evaluator-launcher/examples/slurm_vllm_multinode_ray_tp_pp.yaml
```diff
@@ -86,7 +86,7 @@ deployment:
   port: 8000
   extra_args: "--disable-custom-all-reduce --distributed-executor-backend ray --enforce-eager"
   env_vars:
-    VLLM_USE_RAY_COMPILED_DAG_CHANNEL_TYPE: 'shm'
+    VLLM_USE_RAY_COMPILED_DAG_CHANNEL_TYPE: $host:shm
```
marta-sd: I think this one should be `$lit`

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/cli/migrate_config.py
```diff
@@ -0,0 +1,564 @@
...
+import sys
+
+# Matches a valid env-var-shaped name: starts with letter/underscore, rest alphanumeric/underscore
+_ENV_VAR_NAME_RE = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*$")
+
+# Matches an UPPER_CASE env var name (the typical host-env pattern)
+_UPPER_CASE_RE = re.compile(r"^[A-Z_][A-Z0-9_]*$")
+
+# Prefixes that mean the value is already migrated
+_KNOWN_PREFIXES = ("$host:", "$lit:", "$runtime:")
```
marta-sd: I think these are defined somewhere as constants?

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/env_vars.py
```diff
@@ -0,0 +1,414 @@
...
+    redacted_lines: list[str] = []
+    for line in secrets_content.splitlines():
+        if line.startswith("export ") and "=" in line[len("export ") :]:
+            key, _, value = line[len("export ") :].partition("=")
+            # Strip surrounding quotes added for shell safety
+            unquoted = value.strip('"')
+            if key in literal_names:
+                redacted_lines.append(line)
+            elif len(unquoted) > 4:
+                redacted_lines.append(f'export {key}="***{unquoted[-4:]}"')
```
marta-sd, line 314: I think I've seen a function somewhere in this PR that implements this redacting, maybe we could use it here?

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/env_vars.py
```diff
@@ -0,0 +1,414 @@
...
+
+    # 4. API key — ensure the named env var is present in env_vars.
+    # If the user already declared it (e.g. NGC_API_TOKEN: $host:NGC_API_TOKEN),
+    # do nothing. Otherwise, add it as a host ref so it gets resolved from the host env.
+    if api_key_name and api_key_name not in raw_env_vars:
+        raw_env_vars[api_key_name] = f"$host:{api_key_name}"
+
+    # Check required env vars (excluding NEMO_EVALUATOR_DATASET_DIR)
+    all_var_names = set(raw_env_vars.keys())
+    for required_env_var in task_definition.get("required_env_vars", []):
```
marta-sd: `note`: I think it's never defined (I believe we lost this info when switching to metadata layer). Adding to my backlog to check

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
```diff
@@ -639,7 +660,7 @@ def _create_slurm_sbatch_script(
 
     # Resolve a primary node for single-node sruns (client/proxy/export).
     # This must be safe under `set -u` and work for deployment.type == "none".
-    # Prefer SLURM_JOB_NODELIST but fall back to SLURM_NODELIST; if neither exists,
+    # Prefer SLURM_JOBNODELIST but fall back to SLURM_NODELIST; if neither exists,
```
marta-sd: This comment doesn't match the code below. Where's the typo?

packages/nemo-evaluator-launcher/tests/unit_tests/test_slurm_executor.py
```diff
@@ -119,8 +119,13 @@ def mock_dependencies(self):
...
     ):
-        """Test new execution.env_vars.deployment configuration."""
-        # Add new env_vars structure
+        """Test deployment env vars via top-level env_vars."""
+        # Put env vars in top-level env_vars (new path)
+        base_config["env_vars"] = {
+            "DEPLOY_VAR1": "$lit:deploy_value1",
+            "DEPLOY_VAR2": "$lit:deploy_value2",
+        }
+        # Keep execution.env_vars.deployment for --container-env name pass-through
```
marta-sd: I think this is wrong - the test should pass without the need to specify `execution.env_vars.deployment`, top-level ones should be sufficient.

packages/nemo-evaluator-launcher/tests/unit_tests/test_slurm_executor.py
```diff
@@ -131,27 +136,37 @@ def test_new_execution_env_vars_deployment(
...
     ):
-        """Test new execution.env_vars.evaluation configuration."""
-        # Add new env_vars structure
+        """Test evaluation env vars via top-level env_vars."""
+        # Put env vars in top-level env_vars (new path)
+        base_config["env_vars"] = {
+            "EVAL_VAR1": "$lit:eval_value1",
+            "EVAL_VAR2": "$lit:eval_value2",
+        }
+        # Keep execution.env_vars.evaluation for --container-env name pass-through
```
marta-sd: same here - top-level env vars should suffice

packages/nemo-evaluator-launcher/examples/slurm_vllm_advanced_hf_caching.yaml
agronskiy: Some artifact of the configuration migration, I'm checking

packages/nemo-evaluator-launcher/examples/slurm_vllm_advanced_hf_caching.yaml
agronskiy: moved, thanks

packages/nemo-evaluator-launcher/examples/slurm_vllm_advanced_hf_caching.yaml
agronskiy: ✅

packages/nemo-evaluator-launcher/examples/slurm_vllm_multinode_ray_tp_pp.yaml
agronskiy: 🦅 👁️ ✅

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
agronskiy: `sed` artifact, spot on, reverted ✅

packages/nemo-evaluator-launcher/examples/local_vllm_logprobs.yaml
```diff
@@ -34,18 +33,15 @@ defaults:
 
 execution:
   output_dir: llama_local
-  env_vars:
-    deployment:
-      HF_TOKEN: ${oc.env:HF_TOKEN}  # needed to access meta-llama/Llama-3.1-8B gated model
-
-# specify deployment arguments
```
marta-sd, line 41: Accidentally removed comment

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/cli/migrate_config.py
```diff
@@ -0,0 +1,564 @@
...
+import sys
+
+# Matches a valid env-var-shaped name: starts with letter/underscore, rest alphanumeric/underscore
+_ENV_VAR_NAME_RE = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*$")
+
+# Matches an UPPER_CASE env var name (the typical host-env pattern)
+_UPPER_CASE_RE = re.compile(r"^[A-Z_][A-Z0-9_]*$")
+
+# Prefixes that mean the value is already migrated
+_KNOWN_PREFIXES = ("host:", "lit:", "runtime:")
```
marta-sd, line 50: maybe we could use constants like `EnvVarFromHost` etc instead of strings here and throughout the code?

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/env_vars.py
```diff
@@ -0,0 +1,414 @@
...
+
+    Raises:
+        ValueError: If the value has no recognized prefix or uses Hydra resolver syntax.
+    """
+    # Hard error for Hydra resolver syntax
+    if "${oc.env:" in raw or "${oc.decode:" in raw:
+        raise ValueError(
+            f"Hydra resolver syntax '{raw}' is not allowed in env var fields. "
+            "It resolves secrets into the config object, defeating secret isolation. "
+            "Use 'host:VAR_NAME' instead."
```
marta-sd: Let's use `EnvVarFromHost.PREFIX` to render the error message

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/env_vars.py
```diff
@@ -0,0 +1,414 @@
...
+    if raw.startswith(EnvVarLiteral.PREFIX):
+        return EnvVarLiteral(value=raw[len(EnvVarLiteral.PREFIX) :])
+    if raw.startswith(EnvVarFromHost.PREFIX):
+        return EnvVarFromHost(host_var_name=raw[len(EnvVarFromHost.PREFIX) :])
+    if raw.startswith(EnvVarRuntime.PREFIX):
+        return EnvVarRuntime(runtime_var_name=raw[len(EnvVarRuntime.PREFIX) :])
+
+    # No recognized prefix — build a helpful suggestion
+    if raw.startswith("$") and _ENV_VAR_NAME_RE.match(raw[1:]):
+        suggestion = f"host:{raw[1:]}"
```
marta-sd: same here - let's use constants instead of strings (`EnvVarFromHost.PREFIX` etc)

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/env_vars.py
marta-sd: Update: `def _redact_value` in `cli/run.py`

docs/libraries/nemo-evaluator-launcher/configuration/evaluation/index.md
agronskiy: ✅

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/env_vars.py
agronskiy: Yes, we lsot them concsiously, I kept those in the TaskIR to allow re-adding via FDF later. I'll clean up everything related to it in the codebase for now. The ticket is EVAL-691

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/env_vars.py
agronskiy: ✅  checked other places

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/env_vars.py
agronskiy: ✅

PR #710, opened 2026-02-09
feat: Add top-level task file uploads to MLflow and WandB exporters

piojanu:
- Implemented functionality to upload top-level task files (e.g., run.sub) in both MLflow and WandB exporters.
- Enhanced logging to include details of uploaded files for better traceability.


marta-sd (approved):

marta-sd:
I think we also need to modify mounts for slurm for auto-export to work ([here](https://github.com/NVIDIA-NeMo/Evaluator/blob/f028838e3f1e877922b013c5d54020521e1e9650/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py#L896))

piojanu:
This is probably not very secure. Our run.sub is full of keys. Close for now.

PR #721, opened 2026-02-11
fix: Improve skills install error message for missing GitHub tags

piojanu:
When running a pre-release or test-PyPI version of NEL, the auto-derived GitHub tag may not exist. Replace the raw 404 JSON with a helpful message guiding the user to use --ref.


gchlebus:
LGTM

PR #730, opened 2026-02-16
feat(exporter): stream logs via tar+ssh

piojanu:
## Summary

- **fix(exporter):** Replace `scp -r` with `tar+ssh` for log downloads — streams compressed logs directly, avoiding downloading unwanted subdirectories entirely
- **docs(skills/nel-assistant):** Add `--no-cache-dir` guidance for pip in `pre_cmd` to avoid cross-device link errors when pip cache and temp dirs are on different filesystems inside Docker containers


agronskiy:
`thinking loud`: there's a streaming API in executors (implemented only for local) -- I think we need the impl there at least

PR #731, opened 2026-02-16
feat: nel resume for local and slurm executors

anowaczynski-nvidia:


packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/executor.py
```diff
@@ -568,6 +569,126 @@ def get_status(id: str) -> List[ExecutionStatus]:
...
+            if exit_code is not None and exit_code != 0:
+                raise RuntimeError(
+                    f"Script run_all.sequential.sh failed immediately with exit code {exit_code}. "
+                    f"Check logs in {invocation_dir}/logs/"
+                )
+            return
+
+        # Parallel mode: resume each job individually
+        for job_id in jobs:
+            cls.resume_job(job_id)
```
marta-sd, line 631: I think we could just pass the JobData object (values from the dict) here directly. We've retrieved it from ExecDB already, no need to do it again inside `resume_job`

anowaczynski-nvidia: it's true, but the current `resume_job` signature is consistent with `get_status` and `kill_job` signatures, and multiple ExecDB retrievals are present in each, refactoring all of this together in separate PR makes more sense to me

marta-sd: sounds good

PR #734, opened 2026-02-17
fix: Update nemo-run script to use MBridge checkpoint

athitten:
1. Replaces nemo 2.0 ckpt with MBridge ckpt
2. Also adds polling mechanism to infer the success or failure of evaluation with Ray deployment. This is required since the previous approach of force killing Ray after evaluation was done, led to the nemo-run's status being displayed as FAILED even though evaluation was successful. The case where deployment failed wasnt handled before to kill the job, and the job continued running needing to be manually cancelled. The added changes in scripts/helpers.py with the method `_write_eval_marker()` explicitly writes success or failure markers and does a clean exit for nemo-run's Ray cluster in both sucess, failure cases.

### DIFF
--- a/scripts/evaluation_with_nemo_run.py
+++ b/scripts/evaluation_with_nemo_run.py
@@ -45,6 +45,7 @@ python \
   --pipeline_model_parallel_size {pipeline_model_parallel_size} \
   --expert_model_parallel_size {expert_model_parallel_size} \
   --max_batch_size {max_batch_size} \
+  --inference_max_seq_length {inference_max_seq_length} \
   {additional_args}
 """
 
@@ -62,6 +63,7 @@ python \
   --expert_model_parallel_size {expert_model_parallel_size} \
   --max_batch_size {max_batch_size} \
   --num_replicas {num_replicas} \
+  --inference_max_seq_length {inference_max_seq_length} \
   {additional_args} &
 # [snippet-deploy-end]
 DEPLOY_PID=$!
@@ -94,7 +96,7 @@ def get_parser():
     parser.add_argument(
         "--megatron_checkpoint",
         type=str,
-        default=None,
+        required=True,
         help="Megatron-Bridge checkpoint to be evaluated",
     )
     parser.add_argument(


scripts/evaluation_with_nemo_run.py
```diff
@@ -35,42 +35,67 @@
...
   --tensor_model_parallel_size {tensor_model_parallel_size} \
   --pipeline_model_parallel_size {pipeline_model_parallel_size} \
+  --expert_model_parallel_size {expert_model_parallel_size} \
   --max_batch_size {max_batch_size} \
   --num_replicas {num_replicas} \
-  {additional_args}
-"""
+  {additional_args} &
 # [snippet-deploy-end]
+DEPLOY_PID=$!
```
marta-sd, line 69: Is this ray-specific or we need the same logic for pytriton?

scripts/evaluation_with_nemo_run.py
```diff
@@ -35,42 +35,67 @@
...
+"""
 
 
 def get_parser():
     parser = argparse.ArgumentParser(description="NeMo2.0 Evaluation")
     parser.add_argument(
-        "--nemo_checkpoint",
+        "--megatron_checkpoint",
         type=str,
-        required=True,
```
marta-sd, line 72: Is this argument no longer required?

scripts/evaluation_with_nemo_run.py
```diff
@@ -287,13 +318,15 @@ def main():
...
-        "nemo_checkpoint": args.nemo_checkpoint,
+        "megatron_checkpoint": args.megatron_checkpoint,
         "server_port": args.server_port,
         "server_address": args.server_address,
         "max_input_len": args.max_input_len,
         "tensor_model_parallel_size": args.tensor_parallelism_size,
         "pipeline_model_parallel_size": args.pipeline_parallelism_size,
+        "expert_model_parallel_size": args.expert_model_parallel_size,
         "max_batch_size": args.batch_size,
+        "inference_max_seq_length": args.max_input_len,
```
marta-sd, line 331: We don't have it in commands templates, I think we need to add it there

scripts/evaluation_with_nemo_run.py
```diff
@@ -35,42 +35,67 @@
...
 # [snippet-deploy-end]
+DEPLOY_PID=$!
+
+# Poll for eval completion (success or failure) while deploy is running. This is required since deployment is still alive
+# even after evaluation is finished or evaluation fails.
+while kill -0 $DEPLOY_PID 2>/dev/null; do
+    if [ -f "${{LOG_DIR:-/tmp}}/EVAL_DONE" ]; then
+        kill $DEPLOY_PID 2>/dev/null || true
+        wait $DEPLOY_PID 2>/dev/null || true
+        if [ -f "${{LOG_DIR:-/tmp}}/EVAL_SUCCESS" ]; then
```
marta-sd, line 77: We need to document how this log dir needs to be configured for slurm (both jobs need access to same dir)

scripts/evaluation_with_nemo_run.py
athitten: Its optional and default is 4096 and for most benchmarks it works. So maybe we can just add a note in the docs to increase it if the benchmark has higher max_tokens like humeval ?

scripts/evaluation_with_nemo_run.py
athitten: Correct nemo 2.0 is deprecated so removing it.

scripts/evaluation_with_nemo_run.py
athitten: @marta-sd this doesn't need to be configured. Its automatically set [here](https://github.com/NVIDIA-NeMo/Run/blob/main/nemo_run/run/ray/templates/ray.sub.j2#L86-L87) in ray.sub.j2 that's used by nemo-run for Ray orchestration and logs_dir is computed in [slurm.py](https://github.com/NVIDIA-NeMo/Run/blob/main/nemo_run/run/ray/slurm.py#L155) based on what is set for  [remote_job_dir](https://github.com/NVIDIA-NeMo/Evaluator/blob/main/scripts/evaluation_with_nemo_run.py#L369) in our nemo-run script.

scripts/evaluation_with_nemo_run.py
athitten: This is needed just for Ray. Pytriton takes care of clean exiting when either of deployment or eval fails.

scripts/evaluation_with_nemo_run.py
marta-sd: but it's now `--megatron_checkpoint` flag that was marked as not required:
```
    parser.add_argument(
        "--megatron_checkpoint",
        type=str,
        default=None,
        help="Megatron-Bridge checkpoint to be evaluated",
    )
```
I think we need a checkpoint from the user

scripts/evaluation_with_nemo_run.py
marta-sd: I mean that in lines L35-L89 we have command templates and these templates don't include `{inference_max_seq_length}`. When we later do `TRITON_DEPLOY_SCRIPT.format(
            **commons_args, ...)` and same for ray, this parameter is not used. We need to update the templates to pass this value to the deployment command

scripts/evaluation_with_nemo_run.py
athitten: Oops missed that, let me fix. Good catch

scripts/evaluation_with_nemo_run.py
athitten: Thanks for the find, let me update it.

athitten: Fixed in a7b5d25e6a906bb09b2c2af4b646fdffd29db859

scripts/evaluation_with_nemo_run.py
athitten: Fixed in a7b5d25e6a906bb09b2c2af4b646fdffd29db859

PR #742, opened 2026-02-19
chore: Exporters refactor

marta-sd:


packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/exporters/wandb.py
```diff
@@ -296,205 +264,181 @@ def _log_artifacts(
...
-                    harnesses.append(harness)
+                for d in data:
+                    if d.task and d.task not in benchmarks:
+                        benchmarks.append(d.task)
+                    if d.harness and d.harness not in harnesses:
+                        harnesses.append(d.harness)
                 run.config.update(
                     {"benchmarks": benchmarks, "harnesses": harnesses},
                     allow_val_change=True,
                 )
```
pribalta, line 385: [Power Review] 🔴 **Bare exception swallowed silently in `_create_wandb_run`** · 97% confidence

The knowledge document explicitly prohibits `except Exception: pass` as it hides real failures. The multi_task config-update block catches all exceptions and discards them without logging or re-raising, making silent failures in W&B config propagation invisible.

💡 **Suggestion:** At minimum log the exception: `except Exception as e: logger.warning(f'Failed to update run config in W&B: {e}')`

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/exporters/utils.py
```diff
@@ -508,21 +367,172 @@ def scp_file(remote_path: str, local_path: Path) -> bool:
...
+
+    if not copy_artifacts and not copy_logs:
+        logger.warning("Both copy_artifacts and copy_logs are False, nothing to copy")
+        return
+
+    export_dir.mkdir(parents=True, exist_ok=True)
+
+    exported_files = []
+
+    def cp_file(original_path: Path | str, copy_path: Path | str) -> bool:
```
pribalta: [Power Review] 🔴 **subprocess used to invoke `cp` duplicates stdlib shutil.copy2** · 95% confidence

The architecture rule flags functions that duplicate existing stdlib utilities. Python's `shutil.copy2` exists precisely for file copying and is cross-platform. Using `subprocess.run(['cp', ...])` re-implements stdlib functionality, is Unix-only (will silently fail or error on Windows), and introduces a shell-subprocess boundary where none is needed.

💡 **Suggestion:** Replace the `cp_file` inner function with `shutil.copy2(src, dst)` and handle the return value via a try/except rather than a returncode check. The outer `shutil` import is already guarded elsewhere in the same file.

packages/nemo-evaluator-launcher/tests/unit_tests/test_cli_export_extended.py
```diff
@@ -602,9 +415,10 @@ def test_override_with_list_values(self, capsys):
 
         call_args = mock_export.call_args
         config = call_args[0][2]
+        print(config)
```
pribalta: [Power Review] ⚠️ **Debugging print() statement left in test code** · 95% confidence

A bare `print(config)` statement was added inside a test function (`test_override_with_list_values`). This matches the 'Debugging Artifacts Left in Code' anti-pattern — AI-generated code frequently leaves diagnostic print statements in production/test code. It produces noise in CI output and signals unfinished review.

💡 **Suggestion:** Remove `print(config)`. If the value needs to be inspected on failure, use `assert config == expected, f'Got: {config}'` or rely on pytest's built-in assertion introspection.

pribalta (commented):
## Power Review

> Reviewed 30 files, left 21 suggestions: 2 errors, 13 warnings, 6 suggestions.

### Dimensions Analyzed

| Dimension | 🔴 | ⚠️ | 💡 |
| anti_patterns | 0 | 4 | 1 |
| architecture | 1 | 3 | 1 |
| patterns | 1 | 1 | 1 |
| security-risks | 0 | 3 | 0 |
| testing-quality | 0 | 2 | 3 |
| **Total** | **2** | **13** | **6** |

---
> *A pull request for exporters refactor arrives,*
> *Our review engine carefully contrives.*
> *Still, only you can see the bigger view,*
> *So weigh these findings before you approve.*

---
*Automated review by [Power Review](https://gitlab-master.nvidia.com/dl/JoC/competitive_evaluation/power-review) — findings are advisory.*

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/exporters/utils.py
```diff
@@ -508,21 +367,172 @@ def scp_file(remote_path: str, local_path: Path) -> bool:
 
 
 # =============================================================================
-# PRIVATE HELPER FUNCTIONS
+# COPY ARTIFACTS UTILS
 # =============================================================================
+def copy_local_artifacts(
```
pribalta, line 372: [Power Review] ⚠️ **Missing return type annotation on public function `copy_local_artifacts`** · 92% confidence

The knowledge document flags missing type annotations on public interfaces as a warning. `copy_local_artifacts` is a newly-added public function with all parameter types annotated but no return type annotation. The function returns `List[str]` (or implicitly `None` on the early-return path).

💡 **Suggestion:** Add a return type: `def copy_local_artifacts(...) -> Optional[List[str]]:` (or `List[str]` if the implicit `None` return is intentional and should be documented).

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/cli/info.py
```diff
@@ -116,34 +153,25 @@ def execute(self) -> None:
...
-        args = sys.argv[1:]
-        copy_logs_flag = "--copy-logs" in args
-        copy_artifacts_flag = "--copy-artifacts" in args
-
-        if copy_logs_flag:
-            if self.copy_logs is None:
-                raise ValueError("--copy-logs requires a directory path")
-            if not self.copy_logs.strip():
-                raise ValueError("--copy-logs requires a directory path")
+        # TODO(martas): why do we need to check this? we should just use args values
```
pribalta: [Power Review] ⚠️ **TODO comment left in production code without a tracked issue** · 92% confidence

The knowledge document classifies unresolved `TODO`/`FIXME` comments as 'Debugging Artifacts Left in Code'. The comment `# TODO(martas): why do we need to check this? we should just use args values` reflects unresolved design uncertainty in production logic — it is not a tracked issue and could mislead future readers.

💡 **Suggestion:** Either resolve the uncertainty and remove the comment, or convert it to a tracked issue reference (e.g., `# TODO: see GH-1234`). This same concern applies to the cluster of TODO/FIXME comments added across `executors/slurm/executor.py` (lines 855, 867), `exporters/base.py` (lines 152, 270), `exporters/local.py` (lines 83, 147), and `exporters/wandb.py` (line 214).

packages/nemo-evaluator-launcher/tests/unit_tests/exporters/test_mlflow_exporter.py
```diff
@@ -26,363 +28,138 @@
 class TestMLflowExporter:
     def test_not_available(self):
```
pribalta: [Power Review] ⚠️ **test_not_available does not exercise the MLFLOW_AVAILABLE=False code path** · 92% confidence

The test patches MLFLOW_AVAILABLE=False and then calls exporter.export(['test123.0']), but the job 'test123.0' does not exist in the database and no mock_execdb fixture is used. The failure happens inside _get_jobs_data (DB lookup returns None → failed_jobs) before export_jobs is ever invoked. Because export_jobs is called with an empty data_for_export list, the 'if not self.is_available()' branch returns ([], [], []) harmlessly. The MLFLOW_AVAILABLE=False patch has zero effect on the test outcome — the assertions would pass even if that check were removed. This matches the 'Test Descriptions Don't Match Behavior' anti-pattern: the test claims to verify unavailability handling but actually verifies DB-lookup failure.

💡 **Suggestion:** Either call export_jobs directly with a DataForExport object (mirroring the pattern in test_wandb_exporter.py and test_gsheets_exporter.py), or add mock_execdb and write the job to the DB so that the call reaches export_jobs. Example: `data = DataForExport(...); successful, failed, skipped = exporter.export_jobs([data]); assert failed == [data.job_id]`

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/exporters/gsheets.py
```diff
@@ -28,382 +28,206 @@
...
-    def _get_or_update_headers(
-        self, worksheet, all_metrics: Dict[str, Dict[str, float]]
+    def _update_worksheet(
+        self, worksheet, data_for_export: List[DataForExport]
     ) -> List[str]:
         """Get existing headers or create/update them dynamically."""
 
         # Base columns
-        base_headers = [
+        # FIXME(martas): the logic for base_cols and for merging with existing results duplicates local.py
```
pribalta, line 157: [Power Review] ⚠️ **Self-acknowledged duplication of CSV/sheet merge logic with local.py** · 90% confidence

The architecture rule flags duplicated utility functions. The code itself documents the problem with a FIXME comment: '_update_worksheet' in gsheets.py and '_csv_upsert' in local.py share the same pattern — reading existing headers/rows, deduplicating by job ID, building a unified column list, and writing the result back. Geological layering: the new Google Sheets exporter was bolted on without integrating the shared merge logic.

💡 **Suggestion:** Extract the shared row-merge logic (base column list, deduplication by 'Job ID', column union from old + new rows, writing back) into a shared utility function in `exporters/utils.py` (e.g., `merge_result_rows`). Both `_update_worksheet` and `_csv_upsert` should call it, removing the acknowledged duplication.

pribalta, line 157: [Power Review] ⚠️ **Copy-paste duplication acknowledged by FIXME comment** · 88% confidence

The diff introduces a `FIXME` comment explicitly noting that `base_cols` handling and the existing-results-merging logic duplicates code in `local.py`. Per the 'Copy-Paste Code Blocks' anti-pattern, identical or near-identical blocks signal AI copy-paste and harm maintainability. The FIXME is also itself a debugging artifact left in production code.

💡 **Suggestion:** Extract the shared column-setup and row-merge logic into a utility function in `exporters/utils.py` (e.g., `build_column_headers` and `merge_existing_rows`) and call it from both `gsheets.py` and `local.py`. Remove the FIXME after refactoring.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/cli/info.py
```diff
@@ -27,11 +27,52 @@
...
 from nemo_evaluator_launcher.common.execdb import EXEC_DB_FILE, ExecutionDB, JobData
 from nemo_evaluator_launcher.common.logging_utils import logger
-from nemo_evaluator_launcher.exporters.local import LocalExporter
-from nemo_evaluator_launcher.exporters.utils import get_task_name
+from nemo_evaluator_launcher.exporters.utils import copy_artifacts as copy_artifacts_fn
 
-# Local exporter helper to copy logs and artifacts
-_EXPORT_HELPER = LocalExporter({})
+
+def get_job_paths(job_data: JobData) -> Dict[str, Any]:
```
pribalta, line 33: [Power Review] ⚠️ **get_job_paths re-implemented in CLI layer after removal from BaseExporter — geological layer** · 82% confidence

The architecture rule flags geological layering (new code bolted on without integrating). The diff removes `BaseExporter.get_job_paths`, which served as the single source of truth for executor-to-path mapping. Rather than relocating that function to a shared domain module (e.g., `exporters/utils.py`), it is re-created almost verbatim as a standalone function inside the CLI module `cli/info.py`. Path resolution is now split: the exporter layer uses a different approach (inspecting `output_dir`/`remote_rundir_path` in `copy_artifacts`), while the info CLI owns its own independent copy for display purposes.

💡 **Suggestion:** Move `get_job_paths` to `exporters/utils.py` so both the CLI info command and any future callers share a single implementation. The `cli/info.py` should import and call it rather than define its own copy.

packages/nemo-evaluator-launcher/tests/unit_tests/exporters/test_local_exporter.py
```diff
@@ -15,192 +15,381 @@
...
+
+    assert result.successful_jobs == [j1.job_id]
+    assert result.failed_jobs == []
+
+    # Check that output directory was created with artifacts
+    job_export_dir = output_dir / inv / j1.job_id
+    assert job_export_dir.exists()
+
+
+def test_export_with_format_csv(tmp_path: Path, mock_execdb, prepare_local_job):
```
pribalta, line 82: [Power Review] ⚠️ **test_export_with_format_csv verifies return value but not actual CSV output file** · 82% confidence

The purpose of LocalExporter with format='csv' is to write a CSV file to disk. The test asserts only that result.successful_jobs and result.failed_jobs have expected values, but never checks that processed_results.csv was created or contains any data. Contrast with test_export_with_format_json (line 26) which at least checks that the output directory was created. Omitting verification of the primary side-effect (the CSV file) is a Happy-Path-Only test gap: the test would pass even if the CSV-writing logic raised an exception that was silently swallowed before the return value was constructed.

💡 **Suggestion:** Add at minimum: `csv_path = output_dir / 'processed_results.csv'; assert csv_path.exists()`. A stronger assertion would read and parse the CSV to verify it contains the expected row(s) and headers.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/cli/export.py
```diff
@@ -96,172 +123,144 @@ def execute(self) -> None:
...
-                    if metadata.get("run_url"):
-                        print(f"    URL: {metadata['run_url']}")
-                    if metadata.get("summary_path"):
-                        print(f"    Summary: {metadata['summary_path']}")
-                    path_hint = job_result.get("dest") or metadata.get("output_dir")
-                    if self.dest == "local" and path_hint:
-                        print(f"    Path: {path_hint}")
-                else:
-                    print(f"  {job_id} failed: {job_result.get('message', '')}")
+    def _load_config_from_file(self, config_path: str) -> OmegaConf:
```
pribalta: [Power Review] ⚠️ **Complex config-loading and override-parsing domain logic embedded in CLI command class** · 80% confidence

The architecture rule states business logic must not leak into CLI commands — handlers should remain thin. `_load_config_from_file` performs YAML file I/O, nested key navigation, and destination-specific config extraction. `_apply_overrides` (line 217) implements a regex-based Hydra-style override parser with key-deletion semantics using OmegaConf internals. Both are substantial domain concerns that are not specific to the CLI surface.

💡 **Suggestion:** Extract `_load_config_from_file` and `_apply_overrides` into a dedicated `exporters/config.py` (or `common/export_config.py`) utility module. The `ExportCmd.execute` method should call these helpers without owning their implementation. This also makes the parsing logic independently testable.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/cli/export.py
```diff
@@ -96,172 +123,144 @@ def execute(self) -> None:
...
+
+        Args:
+            config_path: Path to the config file
+
+        Returns:
+            OmegaConf object containing the export configuration
+        """
+        import yaml
+
+        with open(config_path, "r") as f:
```
pribalta: [Power Review] ⚠️ **Missing input validation on user-provided file path at system boundary** · 80% confidence

The `_load_config_from_file` method accepts a raw string `config_path` from a CLI argument and immediately opens it (`open(config_path, 'r')`) without validating that it is a well-formed path, that it exists, or that it resolves within an expected directory. Per the 'Missing Input Validation at System Boundaries' anti-pattern, AI-generated code frequently skips this validation on CLI arguments and file paths.

💡 **Suggestion:** Add an existence check before opening: `p = Path(config_path); if not p.is_file(): raise FileNotFoundError(f'Config file not found: {config_path}')`. Optionally resolve the path and verify it does not escape an expected root directory.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
```diff
@@ -923,7 +908,7 @@ def _generate_auto_export_section(
     s += f"        cd {remote_task_subdir}/artifacts\n"
     for dest in destinations:
         s += f'        echo "Exporting to {dest}..."\n'
-        s += f'        nemo-evaluator-launcher export {job_id} --dest {dest} || echo "Export to {dest} failed"\n'
+        s += f'        nemo-evaluator-launcher export {job_id} --dest {dest} --config export_config.yml --job-dirs {cfg.execution.output_dir} || echo "Export to {dest} failed"\n'
```
pribalta, line 975: [Power Review] ⚠️ **Unquoted shell variables in generated Slurm script (CWE-78)** · 78% confidence

The values `{job_id}`, `{dest}`, and `{cfg.execution.output_dir}` are interpolated directly into a shell script string without shell quoting. If `cfg.execution.output_dir` contains spaces, semicolons, or other shell metacharacters (e.g., `/path with spaces/` or `/path; rm -rf /`), the resulting sbatch script would either fail or allow unintended command execution on the Slurm head node. The knowledge document flags AI-generated command injection (CWE-78) as an error-level issue when user-controlled data reaches shell interpreters without parameterization.

💡 **Suggestion:** Shell-quote all path and identifier values before interpolating into the script string. Use Python's `shlex.quote()` for paths: `shlex.quote(str(cfg.execution.output_dir))`. Similarly quote `job_id` and `dest`. Example: `s += f'        nemo-evaluator-launcher export {shlex.quote(job_id)} --dest {shlex.quote(dest)} --config export_config.yml --job-dirs {shlex.quote(str(cfg.execution.output_dir))} || echo "Export to {shlex.quote(dest)} failed"\n'`

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/exporters/utils.py
```diff
@@ -441,53 +300,53 @@ def scp_file(remote_path: str, local_path: Path) -> bool:
...
-                local_file.parent.mkdir(parents=True, exist_ok=True)
-                if scp_file(remote_file, local_file):
-                    exported_files.append(str(local_file))
+            # Bundle only required+optional artifacts; --ignore-failed-read
+            # skips missing optional files without failing
+            artifact_paths = " ".join(
+                f"artifacts/{a}" for a in get_relevant_artifacts()
+            )
+            remote_tar_cmd = (
+                f"cd {remote_path} && tar --ignore-failed-read -czf - {artifact_paths}"
```
pribalta, line 309: [Power Review] ⚠️ **Unquoted remote path in SSH shell command string (CWE-78)** · 76% confidence

The `remote_path` variable (originating from `job_data.data['remote_rundir_path']`) is interpolated unquoted into a shell command string (`remote_tar_cmd`) that is then executed on the remote host via SSH. SSH passes this string directly to the remote shell for evaluation. If `remote_path` contains shell metacharacters (semicolons, backticks, `$(...)`, or spaces), arbitrary commands could execute on the remote Slurm node under the operator's account. The same risk applies to the logs path at line 347. The knowledge document lists this pattern as CWE-78.

💡 **Suggestion:** Shell-quote `remote_path` before embedding it in the remote command string. Use `shlex.quote(remote_path)` when constructing `remote_tar_cmd`. Example: `remote_tar_cmd = f'cd {shlex.quote(remote_path)} && tar --ignore-failed-read -czf - {artifact_paths}'`. Apply the same fix to `remote_logs` at line 347.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/run.template.sh
```diff
@@ -134,38 +134,42 @@ else
...
+        echo "$(date -u +%Y-%m-%dT%H:%M:%SZ) Job {{ task.job_id }} completed successfully. Starting auto-export..."
 
+        cat > ${artifacts_dir}/export_config.yml << 'EOF'
+{{ auto_export_config_str | indent(2) }}
+EOF
         {% for dest in auto_export_destinations %}
-        echo "$(date -u +%Y-%m-%dT%H:%M:%SZ) Exporting job {{ task.job_id }} to {{ dest }}..." >> "$logs_dir/stdout.log"
-        nemo-evaluator-launcher export {{ task.job_id }} --dest {{ dest }} >> "$logs_dir/stdout.log" 2>&1
+        echo "$(date -u +%Y-%m-%dT%H:%M:%SZ) Exporting job {{ task.job_id }} to {{ dest }}..."
+        nemo-evaluator-launcher export {{ task.job_id }} --dest {{ dest }} --config ${artifacts_dir}/export_config.yml
```
pribalta, line 155: [Power Review] ⚠️ **Unquoted Jinja2 template variables in bash commands (CWE-78)** · 75% confidence

The Jinja2 variables `{{ task.job_id }}` and `{{ dest }}` are rendered directly into the bash script without shell quoting. Because the rendered script is later executed by bash, if either value contains whitespace or metacharacters (e.g., a job_id or destination name with a space), the command parsing would break or enable injection. The knowledge document identifies this pattern (data interpolated into shell commands without parameterization) as CWE-78.

💡 **Suggestion:** Wrap the Jinja2 variables in shell double-quotes or use a Jinja2 filter that shell-escapes values, e.g., `nemo-evaluator-launcher export "{{ task.job_id }}" --dest "{{ dest }}" --config "${artifacts_dir}/export_config.yml"`. Ensure a custom `shell_quote` filter or equivalent escaping is applied in the template engine.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/cli/export.py
```diff
@@ -96,172 +123,144 @@ def execute(self) -> None:
...
+    def _apply_overrides(self, config: OmegaConf) -> None:
+        import re
+
+        # matches ~export.dest.key and +export.dest.key and ++export.dest.key
+        hydra_pattern = re.compile(rf"^(?:~|\+\+|\+)?export\.{self.dest}\.(.+)$")
+        if not self.override:
+            return
+
+        to_del = []
+        fileterd_overrides = []
```
pribalta: [Power Review] 💡 **Misspelled variable name `fileterd_overrides` (inconsistent naming)** · 95% confidence

The knowledge document flags inconsistent naming as a suggestion. `fileterd_overrides` is a typo for `filtered_overrides`, which mismatches the naming intent and the codebase's `snake_case` convention. The same misspelling is reused at line 240, meaning any future reader must infer the intent.

💡 **Suggestion:** Rename `fileterd_overrides` → `filtered_overrides` at lines 226 and 240.

packages/nemo-evaluator-launcher/tests/unit_tests/exporters/test_local_exporter.py
```diff
@@ -15,192 +15,381 @@
...
+            data={},
+            config={"evaluation": {"tasks": [{"name": "simple_evals.humaneval"}]}},
+        )
+
+        # Create artifacts for both jobs
+        for jd in [j1, j2]:
+            _, job_dir = prepare_local_job(jd, with_required=True, with_optional=True)
+            artifacts_dir = job_dir / "artifacts"
+            # Add required files
+            (artifacts_dir / "run_config.yml").write_text(
```
pribalta: [Power Review] 💡 **Identical artifact file setup code duplicated across all TestLocalExporterManualScenarios methods** · 88% confidence

Every test in TestLocalExporterManualScenarios (test_successful_export_multiple_jobs, test_partial_failure_missing_artifacts, test_non_existing_job_id, test_remote_job_export_with_copy, test_json_format_updates_existing_file) copies the same two write_text calls verbatim for run_config.yml and results.yml. This is the Duplicated Test Setup anti-pattern: shared setup should be extracted into a fixture or helper to reduce maintenance burden and the risk of individual tests drifting out of sync.

💡 **Suggestion:** Extract the artifact-file creation into a local helper, e.g.:
```python
def _write_standard_artifacts(artifacts_dir: Path) -> None:
    (artifacts_dir / 'run_config.yml').write_text(
        'framework_name: test-harness\nconfig:\n  type: test_task\n'
        'target:\n  api_endpoint:\n    model_id: test-model\n'
    )
    (artifacts_dir / 'results.yml').write_text(
        'results:\n  tasks:\n    test_task:\n      metrics:\n'
        '        accuracy:\n          scores:\n            accuracy:\n              value: 0.9\nconfig: {}\n'
    )
```
Or add a pytest fixture to conftest.py.

packages/nemo-evaluator-launcher/tests/unit_tests/exporters/test_mlflow_exporter.py
```diff
@@ -26,363 +28,138 @@
...
-        monkeypatch.delenv("MLFLOW_TRACKING_URI", raising=False)
-        monkeypatch.setattr(
-            "nemo_evaluator_launcher.exporters.mlflow.MLFLOW_AVAILABLE",
-            True,
-            raising=True,
+        # Create job with artifacts
+        inv = "test001"
+        artifacts_dir = tmp_path / inv / f"{inv}.0" / "artifacts"
+        artifacts_dir.mkdir(parents=True)
+
```
pribalta: [Power Review] 💡 **Identical artifact directory setup duplicated across all MLflow exporter test methods** · 85% confidence

Five tests (test_export_job_ok, test_export_with_skip_existing, test_export_with_update_existing, test_log_config_params_flattens_config, test_log_config_params_with_max_depth) each copy the same three-step pattern: mkdir the artifacts directory, write run_config.yml, write results.yml with identical content. This is textbook Duplicated Test Setup. Extracting it into a fixture would shorten each test and ensure the artifact format is consistent.

💡 **Suggestion:** Add a pytest fixture (possibly extending the existing prepare_local_job fixture in conftest.py) that creates and returns a pre-populated artifacts directory. Each test can then request this fixture instead of repeating the setup inline.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/exporters/mlflow.py
```diff
@@ -28,626 +30,302 @@
...
-            tags.update(
-                {
-                    "invocation_id": job_data.invocation_id,
-                    "job_id": job_data.job_id,
-                    "task_name": benchmark,
-                    "benchmark": benchmark,
-                    "harness": harness,
-                    "executor": exec_type,
-                }
+        if not self.config.tracking_uri:
```
pribalta: [Power Review] 💡 **Overly defensive null-check inside internal path already guarded by Pydantic validator** · 75% confidence

The 'Overly Defensive Code in Internal Paths' anti-pattern flags redundant guards deep inside trusted internal code. `MLflowExporterConfig._validate_tracking_uri` raises a `ValueError` if `tracking_uri` is missing or empty, so `self.config.tracking_uri` is guaranteed to be non-empty when `export_jobs` is invoked. The second `if not self.config.tracking_uri` check on lines 130-136 is therefore unreachable dead code that adds noise.

💡 **Suggestion:** Remove the redundant guard block at lines 130-136. Trust the Pydantic validator contract that has already ensured `tracking_uri` is set.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/exporters/base.py
```diff
@@ -15,107 +15,347 @@
...
+        if config is None:
+            config = self.config_class()
+        elif isinstance(config, dict):
+            config = self.config_class.model_validate(config)
+        elif not isinstance(config, self.config_class):
+            raise TypeError(
+                f"Invalid config type: {type(config)}. Expected {self.config_class}"
+            )
+        self.config = config
+        self.job_dirs = self.config.job_dirs
```
pribalta: [Power Review] 💡 **Unnecessary instance variable aliases for Pydantic config fields** · 72% confidence

The architecture rule flags unnecessary intermediate variables. Five lines create instance attributes (`self.copy_logs`, `self.copy_artifacts`, etc.) that are simple aliases for identically-named fields on `self.config`. All downstream code in subclasses (e.g., mlflow.py, wandb.py) then accesses `self.config.only_required` and `self.only_required` interchangeably, creating two paths to the same value.

💡 **Suggestion:** Remove the alias assignments (lines 94–98) and update all subclass references to access `self.config.<field>` directly. This eliminates the indirection and makes the config object the single source of truth.

packages/nemo-evaluator-launcher/tests/unit_tests/exporters/test_gsheets_exporter.py
```diff
@@ -20,408 +20,269 @@
...
         )
 
-        artifacts_dir, tmpd = exporter._get_artifacts_locally(jd)
-        assert artifacts_dir is None and tmpd is None
-        assert str(temp_dir) in calls["rm"]
-
-    def test_get_artifacts_locally_missing_artifacts_dir(
-        self, monkeypatch, tmp_path: Path
+        # patch so that Path(service_account_file).exists() is True
+        monkeypatch.setattr(Path, "exists", lambda self: True)
```
pribalta, line 144: [Power Review] 💡 **Global Path.exists patch makes all file-existence checks return True within the test** · 72% confidence

Seven test methods apply `monkeypatch.setattr(Path, 'exists', lambda self: True)`, which overrides Path.exists for every call during the test, not just the target authentication file check. This means the tests cannot detect bugs where the code erroneously skips or fails a different file-existence guard (e.g., checking for the artifacts directory or the spreadsheet credentials). The pattern aligns with the Flaky Test Patterns concern around overly broad I/O mocking that passes locally but may obscure real failures.

💡 **Suggestion:** Narrow the patch to only the specific path that needs to exist, e.g., create the file on disk: `(tmp_path / 'creds.json').write_text('{}')` and pass that path as `service_account_file`. Alternatively, mock at the gspread level (only mock `gspread.service_account`) without touching Path.exists at all.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/cli/export.py
marta-sd: TODO

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/cli/export.py
marta-sd: TODO

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/cli/export.py
marta-sd: TODO

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/exporters/base.py
marta-sd: TODO

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/exporters/mlflow.py
marta-sd: TODO

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/exporters/utils.py
marta-sd: TODO

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/exporters/utils.py
marta-sd: TODO

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/exporters/wandb.py
marta-sd: TODO

packages/nemo-evaluator-launcher/tests/unit_tests/exporters/test_local_exporter.py
marta-sd: TODO

packages/nemo-evaluator-launcher/tests/unit_tests/exporters/test_local_exporter.py
marta-sd: TODO

packages/nemo-evaluator-launcher/tests/unit_tests/test_cli_export_extended.py
marta-sd: TODO

packages/nemo-evaluator-launcher/tests/unit_tests/exporters/test_mlflow_exporter.py
marta-sd: TODO

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/package_info.py
```diff
@@ -15,8 +15,8 @@
 
 # Below is the _next_ version that will be published, not the currently published one.
 MAJOR = 0
-MINOR = 1
-PATCH = 94
+MINOR = 2
```
agronskiy, line 18: `blocking`: should the same go into nemo-evaluator's package_info

marta-sd: I need to merge #792 before bumping nemo-evaluator so I planned to add the change there

PR #757, opened 2026-02-23
feat: Declarative multi-node Ray config and native vLLM data-parallel support

AdamRajfer:
## Summary

- **Declarative `vllm_ray.yaml`**: Move the Ray cluster bootstrap script from a Jinja template (`ray_setup.sh.j2`) + programmatic injection into a standalone Hydra config (`deployment/vllm_ray.yaml`). The full bash script lives in the `command` field using OmegaConf-safe syntax (backticks for command substitution, `let` for arithmetic).
- **`base_command` in `vllm.yaml`**: Extract the `vllm serve ...` invocation into a reusable `base_command` key so `vllm_ray.yaml` can reference it via `${deployment.base_command}`.
- **Native vLLM data-parallel example**: Add `slurm_vllm_multinode_dp.yaml` showing multi-node deployment using vLLM's built-in data parallelism (no Ray required).
- **Remove Jinja template**: Delete `ray_setup.sh.j2` and the programmatic Ray injection block in `executor.py`. Multi-line commands are now handled generically via base64 encoding.
- **Updated docs**: Rewrite the Multi-Node Deployment section in `slurm.md` to cover both non-Ray (data parallelism) and Ray (tensor/pipeline parallelism) approaches.

### OmegaConf compatibility
OmegaConf rejects `$(...)` and `$((...))` in YAML strings. The Ray script uses:
- Backticks `` `cmd` `` instead of `$(cmd)` for command substitution
- `let "x = a / b"` instead of `$((a / b))` for arithmetic
- `|| true` on `let` statements that can evaluate to 0 (to avoid `set -e` failures)

## Test plan

- [x] All 121 unit tests pass
- [x] Dry-run: single-node, multi-node single-instance, multi-instance, multi-node multi-instance
- [x] Cluster run: multi-node multi-instance Ray (TP=8, PP=2, 2 instances x 2 nodes)
- [x] Cluster run: multi-node native data-parallel (no Ray)

### DIFF
--- a/packages/nemo-evaluator-launcher/examples/slurm_vllm_multinode_multiinstance_ray_tp_pp.yaml
+++ b/packages/nemo-evaluator-launcher/examples/slurm_vllm_multinode_multiinstance_ray_tp_pp.yaml
@@ -94,7 +94,7 @@ deployment:
   data_parallel_size: 1
   gpu_memory_utilization: 0.90
   port: 8000
-  extra_args: "--disable-custom-all-reduce"
+  extra_args: "--disable-custom-all-reduce --enforce-eager"
 
 evaluation:
   nemo_evaluator_config:
--- a/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/configs/deployment/vllm_ray.yaml
+++ b/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/configs/deployment/vllm_ray.yaml
@@ -18,75 +18,54 @@
 # Use this config when num_nodes / num_instances > 1.  It inherits every field
 # from vllm.yaml and overrides `command` with a built-in Ray cluster setup
 # script that:
-#   - Starts a Ray head on rank 0, waits for all workers to join, then
-#     launches vLLM with --distributed-executor-backend (default: ray).
-#   - Starts a Ray worker on every other rank (blocks forever so only the
-#     head runs the deployment command).
+#   - Starts a Ray head on PROC_ID 0, polls until all NODES_PER_INSTANCE workers
+#     have joined, then launches vLLM with --distributed-executor-backend (default: ray).
+#   - Starts a Ray worker on every other PROC_ID (blocks until Ray head exits).
 #
-# Supports both single-instance (NUM_INSTANCES == 1) and multi-instance
-# (NUM_INSTANCES > 1, instance-local ranks computed automatically).
+# Environment variables the executor must inject:
+#   PROC_ID              — rank of this process within the instance (0 = head node)
+#   MASTER_IP            — IP of the head node for this instance
+#   NODES_PER_INSTANCE   — number of nodes per instance (used to poll until all workers join)
+#
+# Optional env_vars set by this config (override in your yaml if needed):
+#   VLLM_ENGINE_READY_TIMEOUT_S          — startup timeout (default: 3600s)
+#   VLLM_USE_RAY_COMPILED_DAG_CHANNEL_TYPE — channel type for Ray DAG (default: shm)
 #
 # OmegaConf escaping note:
 #   $VAR           → passed through as literal (safe for bash)
 #   ${key}         → resolved as Hydra interpolation
 #   `cmd`          → use backticks for bash command substitution (OmegaConf rejects $())
-#   let "x = a/b"  → use let for arithmetic (OmegaConf rejects $(()))
 defaults:
   - vllm
   - _self_
 
 distributed_executor_backend: ray
-ray_compiled_dag_channel_type: shm  # "auto", "shm", or "nccl" (default in vLLM: auto)
+
+env_vars:
+  VLLM_ENGINE_READY_TIMEOUT_S: 'lit:3600'  # allow enough time for large models to load across nodes
+  VLLM_USE_RAY_COMPILED_DAG_CHANNEL_TYPE: 'lit:shm'  # "auto", "shm", or "nccl" (default in vLLM: auto)
 
 command: |
   #!/bin/bash
   set -euo pipefail
-
-  export VLLM_USE_RAY_COMPILED_DAG_CHANNEL_TYPE=${deployment.ray_compiled_dag_channel_type}
 
   RAY_PORT=6379
   NODE_PORT=8266
   OBJ_PORT=8267
   RAY_FIXED_PORTS="--node-manager-port=$NODE_PORT --object-manager-port=$OBJ_PORT --metrics-export-port=8269 --dashboard-agent-grpc-port=8270 --dashboard-agent-listen-port=8271 --runtime-env-agent-port=8272"
 
-  NUM_NODES=${execution.num_nodes}
-  NUM_INSTANCES=${execution.num_instances}
-  let "NODES_PER_INSTANCE = NUM_NODES / NUM_INSTANCES"
-
-  # Determine mode: multi-instance (NUM_INSTANCES > 1) or single-instance
-  if [ "$NUM_INSTANCES" -gt 1 ]; then
-      # Multi-instance mode: compute per-instance rank from PROC_ID
-      let "INSTANCE_ID = PROC_ID / NODES_PER_INSTANCE" || true
-      let "INSTANCE_RANK = PROC_ID % NODES_PER_INSTANCE" || true
-      export INSTANCE_ID INSTANCE_RANK
-      let "HEAD_FIELD = INSTANCE_ID * NODES_PER_INSTANCE + 1"
-      export INSTANCE_MASTER_IP=`echo "$ALL_NODE_IPS" | cut -d',' -f"$HEAD_FIELD"`
-      export MASTER_IP="$INSTANCE_MASTER_IP"
-      RANK="$INSTANCE_RANK"
-      EXPECTED_NODES="$NODES_PER_INSTANCE"
-      echo "=== Multi-Instance Ray Setup ==="
-      echo "INSTANCE_ID=$INSTANCE_ID, INSTANCE_RANK=$INSTANCE_RANK, MASTER_IP=$MASTER_IP"
-      echo "NODES_PER_INSTANCE=$NODES_PER_INSTANCE, NUM_INSTANCES=$NUM_INSTANCES"
-  else
-      # Single-instance mode
-      RANK="$PROC_ID"
-      EXPECTED_NODES="$NUM_TASKS"
-      echo "=== Single-Instance Ray Setup ==="
-      echo "PROC_ID=$PROC_ID, MASTER_IP=$MASTER_IP"
-      echo "NUM_TASKS=$NUM_TASKS"
-  fi
+  echo "=== Ray Setup: PROC_ID=$PROC_ID, MASTER_IP=$MASTER_IP ==="
   echo "Hostname: `hostname`"
 
-  if [ "$RANK" -eq 0 ]; then
+  if [ "$PROC_ID" -eq 0 ]; then
       # HEAD node
       export VLLM_HOST_IP=$MASTER_IP
       echo "Starting Ray HEAD on `hostname` with VLLM_HOST_IP=$VLLM_HOST_IP"
       ray start --head --port=$RAY_PORT $RAY_FIXED_PORTS
 
       export RAY_ADDRESS=$MASTER_IP:$RAY_PORT
-      echo "RAY_ADDRESS=$RAY_ADDRESS — waiting for $EXPECTED_NODES node(s) to join..."
-
-      until [ "`ray status 2>/dev/null | grep -c 'node_'`" -ge "$EXPECTED_NODES" ]; do
+      echo "Waiting for $NODES_PER_INSTANCE node(s) to join Ray cluster..."
+      until [ "`ray status 2>/dev/null | grep -c 'node_'`" -ge "$NODES_PER_INSTANCE" ]; do
           sleep 10
       done
 
--- a/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/configs/execution/slurm/default.yaml
+++ b/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/configs/execution/slurm/default.yaml
@@ -31,7 +31,7 @@ sbatch_comment: null # Optional comment for SLURM job (translates to #SBATCH --c
 
 # Deployment-specific SLURM configuration
 deployment:
-  n_tasks: ${execution.num_nodes}  # Number of tasks for deployment srun (default: num_nodes)
+  n_tasks: ${execution.num_nodes}  # total deployment srun tasks across all instances; executor divides by num_instances per srun. Override if your deployment requires more tasks per node (e.g. ntasks_per_node > 1).
 
 mounts:
   deployment: {}
--- a/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
+++ b/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
@@ -866,7 +866,7 @@ def _create_slurm_sbatch_script(
 
     # terminate the server after all evaluation clients finish
     if cfg.deployment.type != "none":
-        s += "kill $SERVER_PID  # terminate the server to finish gracefully\n"
+        s += 'for _pid in "${SERVER_PIDS[@]}"; do kill "$_pid" 2>/dev/null || true; done  # terminate servers\n'
         if cfg.execution.num_instances > 1:
             s += "kill $PROXY_PID  # terminate proxy to finish gracefully\n"
         s += "\n"
@@ -1677,9 +1677,13 @@ def _generate_deployment_srun_command(
     deployment_mounts_list,
     remote_task_subdir,
     deployment_env_var_names: list[str] | None = None,
-    instance_id: int = 0,
 ):
-    """Generate the deployment srun command with proper node/ntask configuration.
+    """Generate per-instance deployment srun commands.
+
+    Loops over num_instances and launches a dedicated srun for each instance on
+    its own node subset.  Multi-instance partitioning lives here; the deployment
+    command receives PROC_ID (rank within the instance) and MASTER_IP (head of
+    the instance) and only needs to handle single-instance setup.
 
     Returns:
         tuple: (script_string, is_potentially_unsafe, debug_comment)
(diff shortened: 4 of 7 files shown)


docs/deployment/launcher-orchestrated/slurm.md
```diff
@@ -138,6 +138,105 @@ evaluation:
...
+  nodes_per_instance: 2      # Each instance spans 2 nodes
+  checkpoint_path: null
+  hf_model_handle: deepseek-ai/DeepSeek-R1
+  served_model_name: deepseek-ai/DeepSeek-R1
+  tensor_parallel_size: 8
+  pipeline_parallel_size: 2
+  data_parallel_size: 1
+  port: 8000
+  extra_args: "--disable-custom-all-reduce --distributed-executor-backend ray --enforce-eager"
+  pre_cmd: |
```
agronskiy: Nice! 

`suggestion`: Can we make it less verbose and more automatic regarding the script injected pre_cmd to avoid copy pasting? That's tad too much of verbosity even for our already verbose config IMO. Scripts like slurm_ray.sh from Nemo-run can be inspiration. 

`suggestion`: more logging output to be added into that script to show the parameters and configs of the launched scripts.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
```diff
@@ -518,6 +518,37 @@ def kill_job(job_id: str) -> None:
...
+    if nodes_per_instance > 1:
+        num_nodes = cfg.execution.num_nodes
+        if num_nodes % nodes_per_instance != 0:
+            raise ValueError(
+                f"num_nodes ({num_nodes}) must be divisible by "
+                f"nodes_per_instance ({nodes_per_instance})"
+            )
+        if not cfg.deployment.get("multiple_instances", False):
+            raise ValueError(
+                "nodes_per_instance > 1 requires deployment.multiple_instances=True"
```
prokotg: why?

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
```diff
@@ -518,6 +518,37 @@ def kill_job(job_id: str) -> None:
...
+            )
+        n_tasks = cfg.execution.get("deployment", {}).get("n_tasks", 1)
+        if n_tasks != num_nodes:
+            raise ValueError(
+                f"nodes_per_instance > 1 requires execution.deployment.n_tasks "
```
prokotg: Must user set it (it should be automatic)?

packages/nemo-evaluator-launcher/examples/slurm_vllm_multinode_multiinstance_ray_tp_pp.yaml
```diff
@@ -0,0 +1,137 @@
...
+  multiple_instances: true   # Enable HAProxy load balancing across instances
+  nodes_per_instance: 2      # Each instance spans 2 nodes
+  checkpoint_path: null
+  hf_model_handle: deepseek-ai/DeepSeek-R1
+  served_model_name: deepseek-ai/DeepSeek-R1
+  tensor_parallel_size: 8
+  pipeline_parallel_size: 2
+  data_parallel_size: 1
+  port: 8000
+  extra_args: "--disable-custom-all-reduce --distributed-executor-backend ray --enforce-eager"
```
marta-sd: Users were complaining in the past about `--enforce-eager` flag as it hurts the performance. Is it required? If so, please add a comment indicating it's needed (ideally with short explanation why). If no -- let's remove it

packages/nemo-evaluator-launcher/examples/slurm_vllm_multinode_multiinstance_ray_tp_pp.yaml
```diff
@@ -0,0 +1,137 @@
...
+  nemo_evaluator_config:
+    config:
+      params:
+        parallelism: 128
+        request_timeout: 3600
+        temperature: 0.6
+        top_p: 0.95
+        max_new_tokens: 32768
+  tasks:
+    - name: gsm8k_cot_instruct
```
marta-sd: Don't we need to set up adapter server with reasoning interceptor?

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
```diff
@@ -1658,13 +1733,18 @@ def _generate_deployment_srun_command(
...
+            f"bash -c '{instance_prefix}{create_pre_script_cmd.cmd} && "
             f"source deployment_pre_cmd.sh && "
             f"{escaped_deployment_cmd}'"
         )
         s += "{} &\n\n".format(wrapped_command)
     else:
-        s += "{} &\n\n".format(cfg.deployment.command)  # run asynchronously
+        if instance_prefix:
+            # Wrap in bash -c to inject instance vars
+            escaped_deployment_cmd = cfg.deployment.command.replace("'", "'\"'\"'")
```
marta-sd: I don't understand this replacement, are you sure it's correct?

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
AWarno: different multinode deployments requite different value for it, e.g. it is different for the trt-llm and vllm, sometimes it is number of gpus

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
AdamRajfer: Yes, this is correct — it's a standard bash idiom for escaping single quotes inside a single-quoted string. Since the whole command is wrapped in `bash -c '...'`, any single quotes inside the deployment command would break the quoting. `.replace("'", "'\"'\"'")` works by:

1. End the current single-quoted string: `'`
2. Append a double-quoted literal single quote: `"'"`
3. Resume the single-quoted string: `'`

So for example, if the deployment command contained `it's`, it becomes `it'"'"'s`, which bash concatenates back into `it's`. This is equivalent to the commonly seen `'\''` pattern but uses double-quoting for the escaped quote instead.

docs/deployment/launcher-orchestrated/slurm.md
AdamRajfer: Done!

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
AdamRajfer: Fixed!

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
AdamRajfer: Thanks for the comment. Please check if it looks good now. I am validating this field only for vLLM now.

packages/nemo-evaluator-launcher/examples/slurm_vllm_multinode_multiinstance_ray_tp_pp.yaml
AdamRajfer: Removed!

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
AdamRajfer: Anyways, please ignore this change as I introduced a broader redesign and it doesn't apply anymore.

packages/nemo-evaluator-launcher/examples/slurm_vllm_multinode_dp.yaml
```diff
@@ -68,27 +68,21 @@ execution:
   partition: batch
   walltime: 01:00:00
 
-  num_nodes: 2 # Number of SLURM nodes for multi-node deployment
-  deployment:
-    n_tasks: ${execution.num_nodes}  # For multi-node vLLM deployment, must match num_nodes
+  num_nodes_per_instance: 2  # Single instance spanning 2 nodes → Ray auto-injected
```
AWarno: I wonder whether we need to make this a breaking change for num_nodes, or if we could simply avoid introducing both num_nodes + num_nodes_per_instance. It may seem like a stylistic issue, but users are likely already familiar with num_nodes, since that is the same term used in Slurm.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/configs/execution/slurm/default.yaml
```diff
@@ -30,7 +32,7 @@ sbatch_comment: null  # Optional comment for SLURM job (translates to #SBATCH --
 
 # Deployment-specific SLURM configuration
 deployment:
-  n_tasks: 1             # Number of tasks for deployment srun (default: 1, for multi-instance set to num_nodes)
```
AWarno: The rule for n_tasks is that it should be ${execution.num_nodes} in most cases. For some deployments, such as TRT-LLM multinode setups, it is related to the number of GPUs (for example: https://github.com/NVIDIA-NeMo/Skills/blob/8a32b13312480cabf8298877d275badd955d669c/nemo_skills/pipeline/utils/server.py#L208
).

If we want a more convenient setup for users we can  provide sensible default in the default Slurm executor default config (${execution.num_nodes} ).

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
```diff
@@ -1588,6 +1727,37 @@ def _generate_deployment_srun_command(
 
     s += "# deployment server\n"
 
+    # Wrap deployment command with Ray setup for multi-node
+    if cfg.execution.num_nodes_per_instance > 1:
+        if cfg.deployment.get("type") != "vllm":
+            raise ValueError(
+                f"Multi-node (num_nodes_per_instance > 1) is only supported for "
+                f"vLLM deployments, got deployment.type={cfg.deployment.get('type')}"
+            )
```
AWarno: Why?

packages/nemo-evaluator-launcher/examples/slurm_vllm_multinode_dp.yaml
prokotg: @AWarno I agree but I think it depends on how we want to handle cumulative number of nodes. I see two options (oth have pros and cons):

1. Treat num_nodes as essentially num_nodes_per_instance. Then multiply num_nodes * number of instances to get cummulative num_nodes. However, there will be no guard if user mistakenly inputs large number of instances (say you want to have 4, but you twitched an you entered 44) 
2. have all: num_nodes (global),  num_nodes_per_instance and instances. This way we can verify if what the user entered make sense. But we do introduce overhead to config creation. 

Also, I wanted to touch base on partial node deployment. Consider the following example:

- node with 8 GPUs
- 2 nodes available
- Deployment with 4 TP 

We should have 4 instances. In vllm that's possible to do because there will be one vllm instance per node with `--tensor-parallelism 4 --data-parallelism 2` but what about deployments that do not have DP mechanism? Would that work?

If we do not plan to add this right now, can we please add it to known issues?

AWarno: I don't think we should deprecated num_nodes having in my mind that we have jobs with deployment none for slurm - and the name is confusing for the user

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/configs/execution/slurm/default.yaml
AdamRajfer: Done!

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
AdamRajfer: Revoked!

docs/deployment/launcher-orchestrated/slurm.md
```diff
@@ -96,6 +97,66 @@ execution:
...
+execution:
+  num_nodes: 2              # Single instance spanning 2 nodes
+
+deployment:
+  tensor_parallel_size: 8
+  pipeline_parallel_size: 2
+```
+
+The `vllm_ray` config automatically handles head/worker node coordination — no `pre_cmd` is needed.
+
```
AWarno: do we need this comment about pre_cmd?

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/configs/deployment/vllm_ray.yaml
```diff
@@ -0,0 +1,110 @@
...
+      export MASTER_IP="$INSTANCE_MASTER_IP"
+      RANK="$INSTANCE_RANK"
+      EXPECTED_NODES="$NODES_PER_INSTANCE"
+      echo "=== Multi-Instance Ray Setup ==="
+      echo "INSTANCE_ID=$INSTANCE_ID, INSTANCE_RANK=$INSTANCE_RANK, MASTER_IP=$MASTER_IP"
+      echo "NODES_PER_INSTANCE=$NODES_PER_INSTANCE, NUM_INSTANCES=$NUM_INSTANCES"
+  else
+      # Single-instance mode: use SLURM_PROCID
+      RANK="$SLURM_PROCID"
+      EXPECTED_NODES="$SLURM_NTASKS"
```
AWarno: cane we make these envs names slurm agnostic?, q: why do we need all of these variables? I think we need just MASTER_IP and $SLURM_PROCID (which can be name to PROCID or sth)

docs/deployment/launcher-orchestrated/slurm.md
AdamRajfer: Removed, good catch.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/configs/deployment/vllm_ray.yaml
AdamRajfer: Done\! The script now maps SLURM vars to generic names at the top (`PROC_ID=$SLURM_PROCID`, `NUM_TASKS=$SLURM_NTASKS`) and uses the generic names throughout. Re `ALL_NODE_IPS` — we still need it in multi-instance mode to compute per-instance head IPs, so it cannot be dropped entirely. Note: cannot use `${VAR:-default}` bash syntax due to OmegaConf restrictions, so the aliases are simple assignments (SLURM always provides these vars in srun context).

packages/nemo-evaluator-launcher/examples/slurm_vllm_multinode_multiinstance_ray_tp_pp.yaml
AdamRajfer: Yes, it is already configured — see `evaluation.nemo_evaluator_config.target.api_endpoint.adapter_config.process_reasoning_traces: true` (lines 107-113).

packages/nemo-evaluator-launcher/examples/slurm_vllm_multinode_dp.yaml
AdamRajfer: Agreed, `num_nodes` is not deprecated. The current design uses `num_nodes` (total SLURM nodes, same semantics as `--nodes`) + `num_instances` (number of independent deployment instances). `nodes_per_instance` is computed internally as `num_nodes / num_instances` — no extra field for the user to set.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/configs/deployment/vllm_ray.yaml
```diff
@@ -0,0 +1,113 @@
...
+command: |
+  #!/bin/bash
+  set -euo pipefail
+
+  export VLLM_USE_RAY_COMPILED_DAG_CHANNEL_TYPE=${deployment.ray_compiled_dag_channel_type}
+
+  # Map SLURM variables to scheduler-agnostic names
+  PROC_ID=$SLURM_PROCID
+  NUM_TASKS=$SLURM_NTASKS
+
```
AWarno: What I mean is that we should export these values to agnostic variables on the Slurm side before reaching this script, since they do not change much. This script should not be tightly coupled to Slurm.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/configs/deployment/vllm_ray.yaml
```diff
@@ -0,0 +1,113 @@
...
+  RAY_PORT=6379
+  NODE_PORT=8266
+  OBJ_PORT=8267
+  RAY_FIXED_PORTS="--node-manager-port=$NODE_PORT --object-manager-port=$OBJ_PORT --metrics-export-port=8269 --dashboard-agent-grpc-port=8270 --dashboard-agent-listen-port=8271 --runtime-env-agent-port=8272"
+
+  NUM_NODES=${execution.num_nodes}
+  NUM_INSTANCES=${execution.num_instances}
+  let "NODES_PER_INSTANCE = NUM_NODES / NUM_INSTANCES"
+
```
AWarno: How is multi-node, multi instance deployment handled without Ray? Right now, this script is tied to Slurm, but it doesn’t have to be. This Ray setup mechanism could be made common for any multi-node environment.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/configs/deployment/vllm_ray.yaml
AWarno: this is the original script: https://github.com/vllm-project/vllm/blob/main/examples/online_serving/multi-node-serving.sh

AdamRajfer: Good point — moved the mapping to the SLURM executor (`executor.py`). It now exports `PROC_ID` and `NUM_TASKS` inside the `bash -c` container wrapper before running the deployment script. The vllm_ray script no longer references any SLURM variables.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/configs/deployment/vllm_ray.yaml
AdamRajfer: The vllm_ray script is now fully scheduler-agnostic — it only uses `PROC_ID`, `NUM_TASKS`, `MASTER_IP`, and `ALL_NODE_IPS`, all exported by the executor. For multi-node multi-instance without Ray, users override `deployment.command` with custom logic (see `slurm_vllm_multinode_dp.yaml`). If we add non-SLURM executors with multi-node support in the future, they would just need to export the same generic variables and the vllm_ray script would work as-is.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/configs/deployment/vllm_ray.yaml
AdamRajfer: Thanks for the reference. Done — the mapping now lives in `executor.py` and the vllm_ray script is SLURM-free.

PR #758, opened 2026-02-23
ci: Enable generated changelogs for packages

chtruong814:
ci: Enable generated changelogs for packages

The changelog will include PRs based on whether the PR updated code in the evaluator or evaluator-launcher packages.  It filters out the release PRs that include the title "beep boop" in them.  

Here are example dry-runs. These examples were done by forcing to create the changelog using a much older release. I'm unsure what this looks like if there are no changes.  We may need to see how that looks first
https://github.com/NVIDIA-NeMo/Evaluator/actions/runs/22288288286/job/64470880862#step:5:138
https://github.com/NVIDIA-NeMo/Evaluator/actions/runs/22288283088/job/64470871401#step:5:95

If no PRs for those packages, just says "No changes".  Unsure what we want to say in that case.


chtruong814:
Will revert this. Accidentally merged.

PR #759, opened 2026-02-24
feat: syntactic sugar overrides for tasks

anowaczynski-nvidia:


packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/override_sugar.py
```diff
@@ -0,0 +1,301 @@
...
+    # Must start with "evaluation.tasks."
+    if not key_value.startswith(_TASKS_PREFIX):
+        return False
+
+    rest = key_value[len(_TASKS_PREFIX) :]
+
+    # The very next dot-segment must be non-integer to qualify as sugar.
+    # For a bare "evaluation.tasks.mmlu" (no more dots, no =), we still consider it
+    # sugar (it would be an error later, but it's not a valid Hydra index override either).
+    first_segment = rest.split(".", 1)[0].split("=", 1)[0]
```
marta-sd, line 83: I don't understand what we do here. For what inputs the if below will be `False`?

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/override_sugar.py
```diff
@@ -0,0 +1,301 @@
...
+
+import re
+from collections import defaultdict
+from typing import TYPE_CHECKING
+
+import yaml
+from omegaconf import OmegaConf
+
+if TYPE_CHECKING:
+    from omegaconf import DictConfig
```
marta-sd: Why do we need it? Ca't we just import `DictConfig`?

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/override_sugar.py
```diff
@@ -0,0 +1,301 @@
...
+# Internal helpers
+# ---------------------------------------------------------------------------
+
+
+def _build_task_name_index(cfg: "DictConfig") -> dict[str, list[int]]:
+    """Return ``{task_name: [idx, …]}`` from ``cfg.evaluation.tasks``."""
+    mapping: dict[str, list[int]] = defaultdict(list)
+    tasks = cfg.evaluation.tasks
+    for idx, task in enumerate(tasks):
+        name = task.name if hasattr(task, "name") else str(task)
```
marta-sd: Why we need this if? Is this for supporting `-o evaluation.tasks=["mmlu"]` syntax? If so, I think we need this in more places

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/override_sugar.py
```diff
@@ -0,0 +1,301 @@
...
+        mapping[name].append(idx)
+    return dict(mapping)
+
+
+def _strip_hydra_prefix(override: str) -> tuple[str, str]:
+    """Return ``(prefix, key_value)`` where *prefix* is ``''``, ``'++'``, ``'+'``, or ``'~'``."""
+    m = _HYDRA_PREFIX_RE.match(override)
+    assert m is not None
```
marta-sd: Why we assert here but not in other L73? Also, please raise a descriptive error here instead of using `assert`

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/override_sugar.py
```diff
@@ -0,0 +1,301 @@
...
+        value_str = rest[eq_idx + 1 :]
+        rest = rest[:eq_idx]
+
+    # rest is now "<task_name_segments>.<remaining_path>" (all dots).
+    # Try all splits from longest candidate name to shortest.
+    segments = rest.split(".")
+    best_name: str | None = None
+    best_remaining: str | None = None
+
+    for i in range(len(segments), 0, -1):
```
marta-sd: I'm confused. Why we do look for longest candidate? If we have `evaluations.tasks.mmlu.a.b.c` we should use `mmlu` as task name, `a.b.c` are params -> the shortest candidate is the most likely match

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/override_sugar.py
```diff
@@ -0,0 +1,301 @@
...
+    raise ValueError(
+        f"Task '{task_name}' appears at indices {indices} — cannot resolve unambiguously.\n"
+        f"Use index-based override instead:\n" + "\n".join(hint_lines)
+    )
+
+
+def _delete_key(cfg: "DictConfig", dotted_path: str) -> None:
+    """Delete a key from *cfg* at *dotted_path*, mirroring Hydra's ``~`` semantics."""
+    parts = dotted_path.rsplit(".", 1)
+    if len(parts) == 2:
```
marta-sd, line 260: I don't get how this function works. Why we have this special case for two parts? In the else below we expect len == 1, >2, or both?

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/cli/run.py
```diff
@@ -135,12 +135,23 @@ def execute(self) -> None:
             # Create RunConfig from the loaded data
             config = OmegaConf.create(config_dict)
         else:
+            from nemo_evaluator_launcher.common.override_sugar import (
```
pribalta, line 204: [Power Review] ⚠️ **Inline import inside execute() is a Geological Layers smell** · 85% confidence

The knowledge document flags 'Geological Layers' — new code bolted on top of old without proper integration. Placing the import of `partition_overrides` and `apply_task_name_overrides` inside the `else` block of `execute()` is exactly this pattern: it avoids touching the module's import block and instead inserts a local patch mid-method. Standard Python module organisation requires top-level imports unless there is an explicit circular-import or lazy-load reason (which is absent here and unexplained by any comment).

💡 **Suggestion:** Move the two imports to the top of `run.py` alongside the other module-level imports. If a circular-import issue exists, document it with an inline comment.

packages/nemo-evaluator-launcher/tests/unit_tests/test_override_sugar.py
```diff
@@ -0,0 +1,348 @@
...
+        assert hydra == overrides
+        assert sugar == []
+
+
+# ---------------------------------------------------------------------------
+# apply_task_name_overrides — basic
+# ---------------------------------------------------------------------------
+
+
+class TestApplyTaskNameOverrides:
```
pribalta, line 228: [Power Review] ⚠️ **Happy-path-only: no test for tasks missing a 'name' attribute** · 80% confidence

The knowledge document warns against happy-path-only test suites: 'test invalid input, empty collections, null/None values'. In `override_sugar.py` line 155, `_build_task_name_index` has a defensive fallback: `name = task.name if hasattr(task, 'name') else str(task)`. This branch — a task dict/object with no `name` key — is never exercised by any test. If the fallback produces unexpected str representations, `apply_task_name_overrides` will silently mismatch task names without any test catching it.

💡 **Suggestion:** Add a fixture and test where at least one task in `evaluation.tasks` lacks a `name` key (e.g., `{"type": "custom"}`). Assert that `apply_task_name_overrides` either resolves the task via `str(task)` or raises a clear error, validating the actual fallback behaviour.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/override_sugar.py
```diff
@@ -0,0 +1,301 @@
...
+
+
+def _parse_value(raw: str) -> object:
+    """Parse a Hydra-style override value string into a Python object.
+
+    Uses YAML parsing to handle strings, numbers, booleans, lists, and dicts.
+    """
+    if not raw:
+        return ""
+    try:
```
pribalta, line 280: [Power Review] ⚠️ **Swallowed exception without logging in `_parse_value`** · 78% confidence

The knowledge document's 'Bare Except / Swallowed Exceptions' pattern states: 'Never swallow exceptions without logging or re-raising.' Here `yaml.YAMLError` is caught and silently discarded — the function returns `raw` as a fallback with no log message, making malformed value strings invisible to operators. While the exception type is specific (not bare `except`), the complete absence of logging means parse failures are undetectable at runtime.

💡 **Suggestion:** Add a debug/warning log before falling back, e.g.:
```python
import logging
_log = logging.getLogger(__name__)
...
except yaml.YAMLError as e:
    _log.debug("Failed to YAML-parse override value %r, treating as raw string: %s", raw, e)
    return raw
```

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/override_sugar.py
```diff
@@ -0,0 +1,301 @@
...
+    for i in range(len(segments), 0, -1):
+        candidate = ".".join(segments[:i])
+        remaining = ".".join(segments[i:])
+        # Check exact match
+        if candidate in name_to_indices:
+            best_name = candidate
+            best_remaining = remaining
+            break
+        # Check suffix match
+        suffix_matches = [n for n in all_names if n.endswith(f".{candidate}") or n == candidate]
```
pribalta: [Power Review] ⚠️ **Dead code: `or n == candidate` in suffix match is unreachable** · 78% confidence

`all_names` is derived from `list(name_to_indices.keys())`. The suffix-match block on line 203 is only reached when the exact-match check on line 198 (`if candidate in name_to_indices`) failed and broke out. Since `all_names` contains exactly the keys of `name_to_indices`, any `n` in `all_names` that equals `candidate` would already have satisfied the exact-match and triggered the `break` on line 200. Therefore `n == candidate` in the suffix predicate can never be True, making it dead code that misleads readers about the function's logic.

💡 **Suggestion:** Remove `or n == candidate` from the suffix match comprehension. The line should read: `suffix_matches = [n for n in all_names if n.endswith(f".{candidate}")]`

packages/nemo-evaluator-launcher/tests/unit_tests/test_override_sugar.py
```diff
@@ -0,0 +1,348 @@
...
+        cfg = apply_task_name_overrides(dotted_name_cfg, overrides)
+        assert cfg.evaluation.tasks[2].nemo_evaluator_config.config.params.parallelism == 16
+
+
+# ---------------------------------------------------------------------------
+# apply_task_name_overrides — error cases
+# ---------------------------------------------------------------------------
+
+
+class TestErrors:
```
pribalta, line 420: [Power Review] ⚠️ **TestErrors class contains success-path tests — test descriptions don't match behavior** · 75% confidence

The knowledge document flags 'Test Descriptions Don't Match Behavior' as a WARNING: test names/groupings describe one thing but test another. The class `TestErrors` (line 302) contains `test_full_name_disambiguation` (line 326) and `test_tilde_deletes_key` (line 336), both of which verify correct successful operation — not error cases. This misleads anyone reading the test suite into thinking these are negative tests when they exercise the happy path.

💡 **Suggestion:** Move `test_full_name_disambiguation` and `test_tilde_deletes_key` into `TestApplyTaskNameOverrides` or a dedicated `TestDeleteOverride` / `TestDisambiguation` class. Keep `TestErrors` strictly for tests that assert `ValueError` (or similar) is raised.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/override_sugar.py
pribalta: [Power Review] 💡 **Overly defensive assert on a regex that always matches** · 85% confidence

The regex `^(\+\+|\+|~)?(.*)$` uses `?` on the prefix group and `.*` for the rest, making it match any input string — `re.match()` will never return None for this pattern. The `assert m is not None` line is therefore an unreachable defensive check inside an internal helper, fitting the 'Overly Defensive Code in Internal Paths' anti-pattern. Additionally, `assert` statements are silently disabled when Python runs with `-O`, making them unreliable as runtime guards even when they could be false.

💡 **Suggestion:** Remove the assert entirely. If a guard is genuinely needed (e.g. for future regex changes), replace it with an explicit `if m is None: raise RuntimeError(...)`, but given the regex, the guard is unnecessary.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/override_sugar.py
```diff
@@ -0,0 +1,301 @@
...
+
+
+def _strip_hydra_prefix(override: str) -> tuple[str, str]:
+    """Return ``(prefix, key_value)`` where *prefix* is ``''``, ``'++'``, ``'+'``, or ``'~'``."""
+    m = _HYDRA_PREFIX_RE.match(override)
+    assert m is not None
+    return (m.group(1) or ""), m.group(2)
+
+
+def _resolve_task_segment(
```
pribalta, line 166: [Power Review] 💡 **Suffix matching logic duplicated between _resolve_task_segment and _lookup_task** · 72% confidence

The knowledge document flags duplicated utility logic as an ERROR-level concern when the same operation is re-implemented instead of being reused. Both `_resolve_task_segment` (line 203: `n.endswith(f'.{candidate}') or n == candidate`) and `_lookup_task` (line 237: `name.endswith(f'.{task_name}')`) independently implement suffix matching with slightly different forms. The distinction between the two functions is subtle (longest-prefix split vs. index resolution), but the core matching predicate is the same logical operation expressed twice.

💡 **Suggestion:** Extract a `_suffix_matches(candidate: str, all_names: list[str]) -> list[str]` helper and call it from both functions. This makes the matching rule a single source of truth and removes the behavioural divergence (e.g., `_resolve_task_segment` includes `or n == candidate` which `_lookup_task` does not).

pribalta (commented):
## Power Review

> Reviewed 3 files, left 7 suggestions: 5 warnings, 2 suggestions.

### Dimensions Analyzed

| Dimension | 🔴 | ⚠️ | 💡 |
| anti_patterns | 0 | 1 | 1 |
| architecture | 0 | 1 | 1 |
| patterns | 0 | 1 | 0 |
| testing-quality | 0 | 2 | 0 |
| **Total** | **0** | **5** | **2** |

---
> *The diff brings change to syntactic sugar overrides for ground,*
> *With 7 things our review has found.*
> *Machines can scan but miss the subtle thread,*
> *A human touch ensures the code is read.*

⏱️ Reviewed in **1.3 min** (est. manual review: ~32 min, **25x faster**)

---

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/override_sugar.py
anowaczynski-nvidia: Claude Code explanation:

> When is `first_segment` empty?  
>
> It's a guard against malformed edge cases where there's nothing meaningful after evaluation.tasks.:
> - evaluation.tasks. (trailing dot, empty rest) → ""
> - evaluation.tasks..foo (double dot, rest starts with .) → ""
> - evaluation.tasks.=value (no name, just =) → ""
>
> All degenerate inputs that aren't valid as either sugar or Hydra overrides, so returning False lets them fall through to Hydra (which will report its own error).

Should we raise an error ourselves here?

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/override_sugar.py
anowaczynski-nvidia: task names can contain dots, for example `lm-evaluation-harness.mmlu` and `nemo_skills.ns_hle_aa`

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/override_sugar.py
anowaczynski-nvidia: `rsplit(".", 1)` splits on the last dot, giving at most 2 parts
  - len == 2: path has at least one dot, e.g. "evaluation.tasks.0.xxx" → parent = "evaluation.tasks.0", key = "xxx", 
  - len == 1: no dots at all, e.g. "evaluation_tasks_0" (very unlikely this function will be called with `dotted_path` without dots)
we can just:
```python
parent_path, key = parts
parent = OmegaConf.select(cfg, parent_path)
```
and never worry about parts having len == 1

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/override_sugar.py
marta-sd: yes, we don't want to pass anything malformed through

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/override_sugar.py
marta-sd: sure, they can, but why we go from the least likely candidate to the most likely one? I think we should invert the for loop

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/override_sugar.py
marta-sd: let's keep it as it is and add a comment the the explanation you just gave me ^

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/override_sugar.py
anowaczynski-nvidia: rewritten without for loop for clarity

PR #762, opened 2026-02-24
feat: Multi-turn design

wprazuch:


packages/nemo-evaluator/examples/byob/mt_bench/benchmark.py
```diff
@@ -0,0 +1,145 @@
...
+    prompt = render_judge_prompt(
+        MT_BENCH_JUDGE_TEMPLATE,
+        conversation=conv_text,
+        criteria=f"Category: {category}. Pay special attention to the follow-up response quality.",
+    )
+
+    try:
+        session = sample.config.get("_judge_session") or _get_judge_session(config)
+        judge_response = judge_call(config, prompt, session=session)
+    except Exception:
```
pribalta, line 117: [Power Review] 🔴 **Bare except swallows exception without logging** · 97% confidence

The knowledge document classifies bare `except Exception: ...` that silently swallows exceptions as an ERROR-level pattern violation. Here the exception from `judge_call` is caught but neither logged nor re-raised, making failures invisible to operators.

💡 **Suggestion:** Log the exception before returning the error sentinel:
```python
except Exception as e:
    logger.warning("Judge call failed in mt_bench scorer", error=str(e))
    return {"quality": 0.0, "judge_grade": "CALL_ERROR"}
```

packages/nemo-evaluator/examples/byob/self_correction/benchmark.py
```diff
@@ -0,0 +1,89 @@
...
+    conversation = [
+        {"role": "user", "content": sample.metadata.get("question", "")},
+        {"role": "assistant", "content": turn1_response},
+        {"role": "user", "content": "Are you sure? Think again and answer with just yes or no."},
+    ]
+
+    # Make turn-2 call using the message array (tests PromptType widening)
+    try:
+        turn2_response = sample.model_call_fn(conversation, "chat")
+    except Exception:
```
pribalta, line 63: [Power Review] 🔴 **Bare except swallows exception without logging** · 97% confidence

The knowledge document classifies `except Exception: pass` (or equivalent silent returns) as an ERROR-level violation. The model call failure is silently swallowed — callers and operators have no visibility into what went wrong.

💡 **Suggestion:** Log the exception before returning:
```python
except Exception as e:
    logger.warning("Turn-2 model call failed", error=str(e))
    return {"correct_t1": False, "consistent": False, "correct_t2": False}
```

packages/nemo-evaluator/src/nemo_evaluator/byob/runner.py
```diff
@@ -437,12 +450,20 @@ def model_call_fn(prompt: str, endpoint_type: str) -> str:
                 session=session,
             )
 
+    # Select evaluation strategy
```
pribalta, line 453: [Power Review] 🔴 **Strategy selection logic duplicated across runner.py and native_runner.py** · 95% confidence

The knowledge document flags duplicated utility logic as an ERROR-level violation: 'AI often regenerates logic that already exists in the codebase.' The same 5-line strategy resolution block (read `strategy` key from `extra_config`, construct `MultiTurnStrategy`) appears identically in both `runner.py` (lines 453–458) and `native_runner.py` (lines 117–123). This is a textbook case of AI-generated duplication that should be centralised.

💡 **Suggestion:** Extract a factory function `build_strategy(extra_config: dict) -> Optional[EvalStrategy]` into `eval_logic.py` (alongside `MultiTurnStrategy`) and call it from both runners. This gives a single place to add future strategies (e.g., `tree_of_thought`, `rag`) without forking the resolution logic again.

packages/nemo-evaluator/src/nemo_evaluator/byob/runner.py
```diff
@@ -103,9 +104,14 @@ def call_model_chat(
     if api_key:
         headers["Authorization"] = f"Bearer {api_key}"
 
+    if isinstance(prompt, list):
```
pribalta, line 107: [Power Review] ⚠️ **Prompt-to-messages coercion logic duplicated in runner.py and native_harness.py** · 92% confidence

The knowledge document warns against duplicated utility functions. The identical four-line block that coerces `prompt` (string or list) into a `messages` array appears in `runner.py:call_model_chat` (lines 107–110) and in `native_harness.py:make_model_call_fn_via_server` (lines 181–184). Duplicating this conversion means any future format change (e.g., adding a `system` role prefix) must be applied in two places.

💡 **Suggestion:** Extract a shared helper `_to_messages(prompt: PromptType) -> List[Dict[str, str]]` into a utility module (e.g., `byob/utils.py` or directly in `decorators.py` near `PromptType`). Import and call it from both `call_model_chat` and `make_model_call_fn_via_server`.

packages/nemo-evaluator/src/nemo_evaluator/byob/native_runner.py
```diff
@@ -114,12 +114,21 @@ def execute(
                 field_mapping=bench.field_mapping,
             )
 
+            # Select evaluation strategy
+            strategy = None
```
pribalta, line 118: [Power Review] ⚠️ **Copy-paste strategy selection logic duplicated across two runners** · 92% confidence

The strategy selection block (lines 118–123 in native_runner.py and lines 454–458 in runner.py) is near-identical. Per the Anti-Patterns knowledge document, copy-paste code blocks within the same PR harm maintainability and signal AI-generated duplication.

💡 **Suggestion:** Extract the strategy selection into a shared helper, e.g., `build_strategy(extra_config: dict) -> Optional[EvalStrategy]` in `eval_logic.py`, and call it from both runners.

packages/nemo-evaluator/src/nemo_evaluator/byob/runner.py
```diff
@@ -143,9 +150,15 @@ def call_model_completions(
         Generated response text.
 
     Raises:
+        ValueError: If prompt is a message array (use chat endpoint instead).
         requests.HTTPError: On non-2xx response.
         requests.Timeout: On timeout.
     """
+    if isinstance(prompt, list):
```
pribalta, line 157: [Power Review] ⚠️ **Completions-endpoint ValueError message duplicated across runner.py and native_harness.py** · 88% confidence

The knowledge document flags duplicated utility code. The exact same `ValueError` message string appears in `runner.py:call_model_completions` (lines 157–161) and `native_harness.py:make_model_call_fn_via_server` (lines 193–197). If the guidance text needs to change it must be updated in both files.

💡 **Suggestion:** Define a module-level constant `_MULTI_TURN_COMPLETIONS_ERROR = "Completions endpoint does not support ..."` (or use the extracted `_to_messages` helper that raises on list input), and reference it from both call sites. Alternatively, consolidating around the `_to_messages` helper from the finding above naturally removes this duplication too.

packages/nemo-evaluator/examples/byob/mt_bench/benchmark.py
```diff
@@ -0,0 +1,145 @@
...
+  nemo-evaluator run_eval --eval_type byob_mt_bench.mt-bench ...
+"""
+
+from nemo_evaluator.byob import benchmark, scorer, ScorerInput
+from nemo_evaluator.byob.judge import (
+    JudgeConfig,
+    judge_call,
+    parse_grade,
+    render_judge_prompt,
+    _get_judge_session,
```
pribalta, line 37: [Power Review] ⚠️ **Example benchmark imports a private (`_`-prefixed) internal function from judge.py** · 87% confidence

The knowledge document flags broken layer boundaries: 'Business logic must not leak into HTTP handlers, CLI commands, or UI components.' The inverse also applies — user-facing example code should not reach into private internals of library modules. `_get_judge_session` is a `_`-prefixed function, signalling it is an implementation detail of `judge.py`. Exposing it to example code couples the example to the internal session management strategy and breaks the public/private contract.

💡 **Suggestion:** Either promote `_get_judge_session` to a public function (rename, document, test it) if it is genuinely needed by external callers, or expose a higher-level public API (e.g., an optional `session` parameter on `judge_score`) so example benchmarks never need to touch session internals.

packages/nemo-evaluator/src/nemo_evaluator/byob/runner.py
```diff
@@ -412,7 +425,7 @@ def main():
     # Create model_call_fn that wraps the HTTP calls
     timeout = args.timeout_per_sample
 
-    def model_call_fn(prompt: str, endpoint_type: str) -> str:
+    def model_call_fn(prompt, endpoint_type: str) -> str:
```
pribalta, line 428: [Power Review] ⚠️ **Type annotation removed from `prompt` parameter of `model_call_fn`** · 85% confidence

The knowledge document flags missing type annotations on public interfaces as WARNING-level. The original code had `prompt: str`; this PR intentionally widened the type but dropped the annotation entirely instead of using the newly introduced `PromptType` alias. The function is passed as a typed `Callable[[PromptType, str], str]` throughout the codebase, so its signature should reflect that.

💡 **Suggestion:** Annotate using the canonical type alias that was introduced in this same PR:
```python
def model_call_fn(prompt: PromptType, endpoint_type: str) -> str:
```
Import `PromptType` from `nemo_evaluator.byob.eval_logic` or `nemo_evaluator.byob.decorators` at the top of `main()`'s scope.

pribalta (commented):
## Power Review

> Reviewed 13 files, left 16 suggestions: 3 errors, 12 warnings, 1 suggestion.

### Dimensions Analyzed

| Dimension | 🔴 | ⚠️ | 💡 |
| anti_patterns | 0 | 2 | 0 |
| architecture | 1 | 4 | 0 |
| hallucination-detection | 0 | 2 | 0 |
| patterns | 2 | 2 | 0 |
| testing-quality | 0 | 2 | 1 |
| **Total** | **3** | **12** | **1** |

---
> *Your PR reshapes the multi-turn design flow,*
> *A warning here — weak spots shouldn’t grow.*
> *For AI writes fast but sometimes writes it wrong,*
> *So human eyes must check where things belong.*

⏱️ Reviewed in **1.5 min** (est. manual review: ~44 min, **30x faster**)

---

packages/nemo-evaluator/src/nemo_evaluator/byob/eval_logic.py
```diff
@@ -193,10 +196,86 @@ def evaluate_sample(
         return sample_scores, prediction
 
 
+class MultiTurnStrategy:
```
pribalta, line 199: [Power Review] ⚠️ **MultiTurnStrategy class introduced with no unit test coverage** · 85% confidence

The `MultiTurnStrategy` class (lines 199–272) is non-trivial new logic: it iterates over turns, builds conversation history, handles model errors mid-turn, handles scorer errors, and emits `trajectory` data. The diff adds no unit tests for any of these paths. Per the knowledge document, a test suite without negative tests (error paths, edge cases) for new functionality is incomplete. The missing turns field path (`skipped_missing_field`), mid-turn model failure (`skipped_model_error`), and scorer failure (`skipped_scorer_error`) are all untested.

💡 **Suggestion:** Add unit tests in `test_byob_runner.py` or a new `test_eval_logic.py` covering at minimum: (1) normal multi-turn flow with 2 turns producing correct `trajectory`, (2) row missing the `turns_field` returns `skipped_missing_field`, (3) `model_call_fn` raising mid-turn returns `skipped_model_error` with partial trajectory, (4) scorer raising returns `skipped_scorer_error` with the last response preserved.

packages/nemo-evaluator/tests/unit_tests/byob/test_byob_runner.py
```diff
@@ -349,8 +349,8 @@ def test_save_predictions_content_structure(self, tmp_path, mock_benchmark, samp
                 # Check types
                 assert isinstance(pred_dict["sample_id"], int), \
                     f"Line {line_num}: sample_id should be int"
-                assert isinstance(pred_dict["prompt"], str), \
-                    f"Line {line_num}: prompt should be str"
+                assert isinstance(pred_dict["prompt"], (str, list)), \
```
pribalta, line 352: [Power Review] ⚠️ **Widened type assertion does not exercise the new list (multi-turn) prompt path** · 78% confidence

The assertion was updated from `isinstance(pred_dict['prompt'], str)` to `isinstance(pred_dict['prompt'], (str, list))` to accommodate the new PromptType. However, if the test fixtures only produce single-turn benchmarks (str prompts), the `list` branch of the assertion is never actually evaluated. This creates a false sense of coverage: the assertion passes regardless of whether a multi-turn prediction is ever written to the predictions file. Per the knowledge document, happy-path-only tests that skip new code paths are a warning-level issue.

💡 **Suggestion:** Add a parallel test case (or parameterize the existing one) with a `MultiTurnStrategy`-backed mock benchmark that produces `list` prompts. Verify that saved predictions contain `prompt` values that are lists of dicts and that all trajectory/conversation fields are correctly serialized.

packages/nemo-evaluator/examples/byob/mt_bench/benchmark.py
```diff
@@ -0,0 +1,145 @@
...
+    name="mt-bench",
+    dataset="hf://HuggingFaceH4/mt_bench_prompts?split=train",
+    # prompt is ignored by MultiTurnStrategy (it reads row["prompt"] directly)
+    prompt="{prompt}",
+    target_field="reference",
+    endpoint_type="chat",
+    requirements=["datasets"],
+    extra={
+        "judge": {
+            "url": "https://inference-api.nvidia.com/v1",
```
pribalta, line 82: [Power Review] ⚠️ **Fabricated NVIDIA inference API base URL** · 75% confidence

The URL `https://inference-api.nvidia.com/v1` does not match NVIDIA's documented cloud inference endpoint. NVIDIA's production NIM API base URL is `https://integrate.api.nvidia.com/v1`. The subdomain `inference-api.nvidia.com` does not correspond to any known NVIDIA service and is likely a hallucinated endpoint that will produce connection errors at runtime.

💡 **Suggestion:** Replace `https://inference-api.nvidia.com/v1` with `https://integrate.api.nvidia.com/v1` (NVIDIA's documented NIM cloud API endpoint), or verify the correct URL against NVIDIA's current API documentation.

packages/nemo-evaluator/examples/byob/mt_bench/benchmark.py
```diff
@@ -0,0 +1,145 @@
...
+Rate the overall quality of the assistant's responses on a scale of 1 to 10, \
+where 1 is extremely poor and 10 is excellent.
+
+Think step by step, then output your rating on a new line:
+GRADE: [[score]]
+where score is an integer from 1 to 10.\
+"""
+
+
+def _format_conversation(conversation):
```
pribalta, line 62: [Power Review] ⚠️ **Public helper function `_format_conversation` missing type annotations** · 72% confidence

The knowledge document flags missing type annotations on public interfaces as WARNING-level. While prefixed with `_`, this function is module-level and its parameter type (`List[Dict[str, str]]`) and return type (`str`) are unambiguous and should be annotated.

💡 **Suggestion:** ```python
from typing import Dict, List

def _format_conversation(conversation: List[Dict[str, str]]) -> str:
```

File: packages/nemo-evaluator/examples/byob/mt_bench/benchmark.py [hunk 13]
```diff
@@ -0,0 +1,145 @@
...
+    dataset="hf://HuggingFaceH4/mt_bench_prompts?split=train",
+    # prompt is ignored by MultiTurnStrategy (it reads row["prompt"] directly)
+    prompt="{prompt}",
+    target_field="reference",
+    endpoint_type="chat",
+    requirements=["datasets"],
+    extra={
+        "judge": {
+            "url": "https://inference-api.nvidia.com/v1",
+            "model_id": "gcp/google/gemini-2.5-flash",
```
pribalta, line 83: [Power Review] ⚠️ **Potentially fabricated model ID namespace prefix `gcp/`** · 72% confidence

The model ID `gcp/google/gemini-2.5-flash` uses a `gcp/` prefix that does not appear in NVIDIA's NIM model catalog naming conventions. Google models on NVIDIA's inference platform are referenced as `google/<model-name>` (e.g., `google/gemini-2.5-flash`), not `gcp/google/<model-name>`. The `gcp/` namespace prefix is likely an AI-invented routing prefix that will cause a 404 or model-not-found error at runtime.

💡 **Suggestion:** Change `"model_id": "gcp/google/gemini-2.5-flash"` to `"model_id": "google/gemini-2.5-flash"` and verify the model identifier against NVIDIA's current NIM model catalog.

packages/nemo-evaluator/src/nemo_evaluator/byob/eval_logic.py
```diff
@@ -193,10 +196,86 @@ def evaluate_sample(
...
+    Expects the dataset row to have a ``turns`` field (list of prompt strings)
+    or a ``prompt`` field that is a list. Each turn is sent as a message array
+    with full conversation history. The scorer receives the final response and
+    the full trajectory.
+    """
+
+    def __init__(self, turns_field: str = "prompt"):
+        self.turns_field = turns_field
+
+    def evaluate_sample(
```
pribalta, line 211: [Power Review] ⚠️ **MultiTurnStrategy.evaluate_sample exceeds 50-line guideline (~62 lines)** · 72% confidence

The knowledge document warns: 'AI generates long, flat functions instead of composing smaller units. Functions exceeding ~50 lines are harder to test, review, and maintain.' The `evaluate_sample` method (lines 211–272) runs to approximately 62 lines, combining turn-loop execution, error-path construction, scorer invocation, and result assembly in a single flat body.

💡 **Suggestion:** Extract the inner turn loop into `_run_turns(row, turns_field, model_call_fn, endpoint_type) -> Tuple[List[Dict], List[str]]` and the scorer invocation+result construction into `_score_conversation(...)`. This makes each path independently testable and keeps `evaluate_sample` as a thin coordinator.

packages/nemo-evaluator/src/nemo_evaluator/core/native_harness.py
```diff
@@ -16,23 +16,27 @@
 """Native harness interface and registry for in-process evaluation execution."""
 
 from contextlib import contextmanager
-from typing import Callable, Dict, Optional, Protocol, TypeAlias, runtime_checkable
+from typing import Callable, Dict, List, Optional, Protocol, TypeAlias, Union, runtime_checkable
```
pribalta, line 19: [Power Review] ⚠️ **Unused imports `List` and `Union` added to native_harness.py** · 70% confidence

The diff adds `List` and `Union` to the `typing` imports in this file, but all changed code paths in native_harness.py use the `PromptType` alias imported from `byob.decorators` — not `List` or `Union` directly. Per the Anti-Patterns knowledge document, unused imports constitute dead code that confuses readers and should be removed.

💡 **Suggestion:** Remove `List` and `Union` from the `typing` import in native_harness.py since `PromptType` (imported from `nemo_evaluator.byob.decorators`) already encapsulates those types.

packages/nemo-evaluator/src/nemo_evaluator/byob/runner.py
pribalta, line 157: [Power Review] 💡 **New ValueError path in call_model_completions is not covered by tests** · 75% confidence

`call_model_completions` now raises `ValueError` when passed a message array. This is a newly introduced error path that belongs to changed code in this diff. Per the knowledge document, tests should cover error paths and expected exceptions—not just the happy path. No test in the diff exercises this guard.

💡 **Suggestion:** Add a unit test (e.g., in `test_byob_runner.py`) that calls `call_model_completions` with a list prompt and asserts `ValueError` is raised with a message mentioning the chat endpoint.

PR #680, opened 2026-02-01
feat: Add telemetry support to nemo-evaluator and nemo-evaluator-launcher

gchlebus:
## Summary

Add anonymous usage telemetry to understand adoption patterns and guide roadmap decisions.

**nemo-evaluator:**
- Add `TelemetryHandler` with async batching, retry logic, and graceful degradation
- Add `EvaluationTaskEvent` with task, harness, model, duration, and status fields
- Emit STARTED event at evaluation begin, SUCCESS/FAILURE at end
- Support opt-out via `NEMO_TELEMETRY_ENABLED=false` env var

**nemo-evaluator-launcher:**
- Re-export shared telemetry infrastructure from nemo-evaluator (DRY)
- Add `LauncherJobEvent` with executor, deployment, model, tasks, exporters, status
- Add `--no-telemetry` / `-T` CLI flag
- Propagate telemetry env vars (`NEMO_TELEMETRY_SESSION_ID`, `NEMO_TELEMETRY_ENABLED`, `NEMO_TELEMETRY_ENDPOINT`) to child processes in local, slurm, and lepton executors

**Both packages:**
- Add `httpx` dependency for async HTTP requests
- Add comprehensive unit tests (38 total)
- Session ID propagation enables correlating launcher and evaluator events

## Schema alignment

Event names updated to match the registered telemetry schema:
- `EvaluationTaskEvent._event_name`: `"evaluation_task"` → `"EvaluationTaskEvent"`
- `LauncherJobEvent._event_name`: `"launcher_job"` → `"LauncherJobEvent"`

The telemetry endpoint is configurable via `NEMO_TELEMETRY_ENDPOINT` (defaults to production). To test against staging:
```bash
export NEMO_TELEMETRY_ENDPOINT=https://events.telemetry.data-uat.nvidia.com/v1.1/events/json
```

## Key design decisions

- **Opt-out by default**: Telemetry enabled unless `NEMO_TELEMETRY_ENABLED=false`
- **User notification**: `Telemetry enabled. Set NEMO_TELEMETRY_ENABLED=false to disable.`
- **Graceful degradation**: Telemetry failures never disrupt main application
- **Session correlation**: Launcher generates session ID, propagates to evaluator via env var
- **Code reuse**: Launcher imports shared infrastructure from nemo-evaluator

## Test plan

- [x] Unit tests for telemetry module (16 tests in nemo-evaluator)
- [x] Unit tests for telemetry module (22 tests in nemo-evaluator-launcher)
[...]

PR #767, opened 2026-02-24
feat: add anonymous telemetry to nemo-evaluator

gchlebus:
## Summary
- Add `TelemetryHandler` with async batching, retry logic, and graceful degradation
- Add `EvaluationTaskEvent` with task, framework, model, duration, and status fields
- Support three telemetry levels (`0`=off, `1`=minimal, `2`=full) via `NEMO_EVALUATOR_TELEMETRY_LEVEL` env var or persistent config file (`~/.config/nemo-evaluator/config.yaml`)
- Redact `model_id` at level 1 (MINIMAL)
- Add telemetry and configuration documentation

This is part 1 of splitting the telemetry PR (#680). After this PR gets merged, the PR #680 (launcher changes) will follow. This PR intentionally keeps all docs from the original PR to reduce the work required for the PR split.

## Test plan
- [x] Unit tests pass (`uv run --group test pytest tests/unit_tests/ -v` — 439 passed)
- [x] Pre-commit clean
- [ ] Integration tests (require GPU environment)

🤖 Generated with [Claude Code](https://claude.com/claude-code). Corrected by Human™️

### DIFF
--- a/packages/nemo-evaluator/src/nemo_evaluator/telemetry.py
+++ b/packages/nemo-evaluator/src/nemo_evaluator/telemetry.py
@@ -401,12 +401,6 @@ class TelemetryHandler:
             source_client_version=self._source_client_version,
             session_id=self._session_id,
         )
-        import json as _json
-
-        logger.warning(
-            "Telemetry payload (debug)",
-            payload=_json.dumps(payload, indent=2, default=str),
-        )
         try:
             response = await client.post(NEMO_TELEMETRY_ENDPOINT, json=payload)
             # 2xx, 400, 422 are all considered complete (no retry)


pribalta (commented):
## Power Review

> Reviewed 19 files, left 17 suggestions: 4 errors, 8 warnings, 5 suggestions.

### Dimensions Analyzed

| Dimension | 🔴 | ⚠️ | 💡 |
| anti_patterns | 0 | 1 | 1 |
| architecture | 0 | 3 | 1 |
| patterns | 2 | 0 | 2 |
| testing-quality | 2 | 4 | 1 |
| **Total** | **4** | **8** | **5** |

---
> *New logic lands in add anonymous telemetry to space,*
> *Our agents checked each file and trace.*
> *Yet automated eyes can overlook a flaw,*
> *So review these notes — that’s the golden law.*

⏱️ Reviewed in **2.9 min** (est. manual review: ~60 min, **21x faster**)

---

packages/nemo-evaluator/tests/integration_tests/test_telemetry_e2e.py
```diff
@@ -0,0 +1,89 @@
...
+            ),
+        ),
+        patch(
+            "nemo_evaluator.core.evaluate.monitor_memory_usage",
+            return_value=(MagicMock(model_dump=lambda **k: {"test": "result"}), {}),
+        ),
+    )
+
+
+def test_evaluate_sends_telemetry_at_default_level(tmp_path):
```
pribalta, line 67: [Power Review] 🔴 **Test with no assertions: test_evaluate_sends_telemetry_at_default_level** · 95% confidence

Per the Testing Quality knowledge document (ERROR: Tests With No Assertions): 'Every test function must contain at least one explicit assertion.' This test calls `evaluate()` but never asserts anything about the result, the telemetry events sent, or their content. It passes solely by not crashing.

💡 **Suggestion:** Add assertions that verify telemetry events were actually sent with expected content. For example, capture enqueued events and assert on their count, event names, model field value, and status transitions (STARTED → SUCCESS).

packages/nemo-evaluator/tests/integration_tests/test_telemetry_e2e.py
```diff
@@ -0,0 +1,89 @@
...
+            eval_cfg=EvaluationConfig(
+                output_dir=str(tmp_path),
+                params=ConfigParams(),
+                type="integration-test-task",
+            ),
+            target_cfg=EvaluationTarget(),
+        )
+
+
+def test_evaluate_sends_telemetry_at_minimal_level(tmp_path, monkeypatch):
```
pribalta, line 98: [Power Review] 🔴 **Test with no assertions: test_evaluate_sends_telemetry_at_minimal_level** · 95% confidence

Per the Testing Quality knowledge document (ERROR: Tests With No Assertions): 'Every test function must contain at least one explicit assertion.' This test calls `evaluate()` at telemetry level 1 but never asserts anything — not that events were sent, not that the model was redacted, not the event structure.

💡 **Suggestion:** Add assertions verifying the telemetry events have `model='redacted'` at level 1. Capture the events (e.g., via a spy on `enqueue`) and assert the model field is 'redacted' and the correct number of events were emitted.

packages/nemo-evaluator/tests/integration_tests/test_telemetry_e2e.py
pribalta, line 67: [Power Review] 🔴 **Test with zero assertions: test_evaluate_sends_telemetry_at_default_level** · 95% confidence

Per the knowledge document: 'AI produces tests that run code but verify nothing — they "pass" by not crashing. Every test must assert an expected outcome.' This test calls evaluate() but has no assert statement. It only verifies the code doesn't crash, not that telemetry events were actually sent, had the correct model name, or were directed to the right endpoint.

💡 **Suggestion:** Add assertions that verify the expected telemetry behavior — e.g., capture enqueued events and assert on their fields (task, model, status), or assert that the staging endpoint received a POST with the expected payload structure.

packages/nemo-evaluator/tests/integration_tests/test_telemetry_e2e.py
pribalta, line 98: [Power Review] 🔴 **Test with zero assertions: test_evaluate_sends_telemetry_at_minimal_level** · 95% confidence

Per the knowledge document: 'Every test must assert an expected outcome.' This test sets telemetry level to 1 (MINIMAL) and calls evaluate(), but never asserts that the model field was redacted. The docstring says 'model should be redacted' but the test doesn't verify this claim.

💡 **Suggestion:** Add assertions verifying that the telemetry events enqueued at level 1 have model='redacted'. Consider using an approach similar to the CapturingHandler pattern in test_evaluate.py to capture and inspect the events.

packages/nemo-evaluator/tests/conftest.py
```diff
@@ -0,0 +1,49 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from unittest.mock import AsyncMock, patch
+
+import pytest
+
+NEMO_TELEMETRY_STAGING_ENDPOINT = (
```
pribalta, line 20: [Power Review] ⚠️ **Duplicated staging endpoint constant across conftest files** · 90% confidence

The constant NEMO_TELEMETRY_STAGING_ENDPOINT is defined identically in both tests/conftest.py (line 20) and tests/integration_tests/conftest.py (line 27). The knowledge document flags 'functions that duplicate existing helpers' as an error-level concern. While these are constants rather than functions, the same principle applies — duplicated literals are a maintenance risk that can drift silently.

💡 **Suggestion:** Define NEMO_TELEMETRY_STAGING_ENDPOINT once in tests/conftest.py and import it in tests/integration_tests/conftest.py: `from tests.conftest import NEMO_TELEMETRY_STAGING_ENDPOINT`.

packages/nemo-evaluator/tests/unit_tests/test_telemetry.py
```diff
@@ -0,0 +1,324 @@
...
+        )
+        with patch.dict(os.environ, {"NEMO_EVALUATOR_TELEMETRY_LEVEL": "2"}):
+            with patch("nemo_evaluator.config.CONFIG_FILE", config_file):
+                assert get_telemetry_level() == TelemetryLevel.DEFAULT
+
+
+class TestSessionId:
+    """Tests for session ID generation and retrieval."""
+
+    def test_session_id_generation(self):
```
pribalta, line 118: [Power Review] ⚠️ **Test with no explicit assertion: test_session_id_generation** · 85% confidence

Per the Testing Quality knowledge document (ERROR: Tests With No Assertions): 'Every test function must contain at least one explicit assertion.' The test calls `uuid.UUID(session_id)` which would raise on invalid input, but there is no explicit `assert` statement. While the implicit exception-based check has value, it violates the documented rule requiring explicit assertions.

💡 **Suggestion:** Add an explicit assertion, e.g.: `assert uuid.UUID(session_id)` or `assert isinstance(uuid.UUID(session_id), uuid.UUID)`.

packages/nemo-evaluator/tests/unit_tests/test_telemetry.py
```diff
@@ -0,0 +1,324 @@
...
+        from nemo_evaluator.telemetry import (
+            TELEMETRY_SESSION_ID_ENV_VAR,
+            get_session_id,
+        )
+
+        test_id = "test-session-123"
+        with patch.dict(os.environ, {TELEMETRY_SESSION_ID_ENV_VAR: test_id}):
+            assert get_session_id() == test_id
+
+    def test_get_session_id_generates_new(self):
```
pribalta, line 143: [Power Review] ⚠️ **Test with no explicit assertion: test_get_session_id_generates_new** · 85% confidence

Per the Testing Quality knowledge document (ERROR: Tests With No Assertions): 'Every test function must contain at least one explicit assertion.' The test calls `uuid.UUID(session_id)` as implicit validation but has no explicit `assert` statement.

💡 **Suggestion:** Add an explicit assertion, e.g.: `assert uuid.UUID(session_id)` or `session = uuid.UUID(session_id); assert session.version == 4`.

packages/nemo-evaluator/tests/integration_tests/conftest.py
```diff
@@ -0,0 +1,75 @@
...
+Send failures cause test failures by default. Set TELEMETRY_ALLOW_SEND_FAILURES=1
+to suppress (e.g. when running without network access).
+"""
+
+import os
+from unittest.mock import patch
+
+import pytest
+
+NEMO_TELEMETRY_STAGING_ENDPOINT = (
```
pribalta: [Power Review] ⚠️ **Duplicated staging endpoint constant across conftest files** · 85% confidence

The constant `NEMO_TELEMETRY_STAGING_ENDPOINT` with value `https://events.telemetry.data-uat.nvidia.com/v1.1/events/json` is identically defined in both `tests/conftest.py` (line 20) and `tests/integration_tests/conftest.py` (line 27). Both conftest files also share a similar patching pattern for `nemo_evaluator.telemetry.NEMO_TELEMETRY_ENDPOINT`. Per the knowledge document: 'Identical or near-identical code blocks within the same PR signal AI copy-paste. Duplication harms maintainability.'

💡 **Suggestion:** Extract `NEMO_TELEMETRY_STAGING_ENDPOINT` into a shared test utilities module (e.g., `tests/telemetry_test_helpers.py`) and import it from both conftest files. This ensures the URL is defined once and updated in one place.

packages/nemo-evaluator/src/nemo_evaluator/core/evaluate.py
```diff
@@ -316,6 +309,89 @@ def run_evaluation_core():
     return evaluation_result
 
 
+def evaluate(
```
pribalta, line 312: [Power Review] ⚠️ **Monolithic evaluate() function (~65 lines of logic)** · 85% confidence

The refactored evaluate() function spans lines 312–393 (~65 lines of code excluding the docstring), exceeding the ~50-line guideline. It mixes three distinct concerns: telemetry lifecycle management (handler creation, start, stop), model name extraction and redaction logic, and the actual evaluation delegation. The knowledge document warns that 'AI generates long, flat functions instead of composing smaller units' and recommends extracting logical blocks into well-named helpers.

💡 **Suggestion:** Extract the telemetry orchestration into a dedicated helper or context manager (e.g., `_telemetry_context(eval_cfg, target_cfg)` returning a context manager that yields the handler and handles start/stop/events). This would keep evaluate() as a thin coordinator, consistent with the existing `_run_evaluation()` extraction pattern.

packages/nemo-evaluator/tests/unit_tests/test_telemetry.py
```diff
@@ -0,0 +1,324 @@
...
+            model="llama-3.1-8b",
+            status=StatusEnum.STARTED,
+        )
+
+        serialized = event.model_dump(by_alias=True)
+        assert serialized["status"] == "started"
+        assert serialized["executionDurationSeconds"] == 0.0
+
+
+class TestTelemetryHandler:
```
pribalta, line 194: [Power Review] ⚠️ **Happy-path-only tests for TelemetryHandler** · 75% confidence

Per the Testing Quality knowledge document (WARNING: Happy-Path-Only Tests): 'A test suite without negative tests is incomplete.' The `TestTelemetryHandler` class tests initialization, enqueue-when-disabled, enqueue-when-enabled, and graceful degradation for invalid events. However, it has no tests for error/edge paths: retry logic (DLQ behavior), max_queue_size triggering a flush, HTTP error handling (413 split, 5xx retry, 408 timeout), or the start/stop lifecycle.

💡 **Suggestion:** Add tests for: (1) events exceeding `max_queue_size` trigger a flush, (2) DLQ retry logic when `_send_events` fails, (3) events dropped after `max_retries`, (4) HTTP 413 payload splitting behavior, (5) HTTP 5xx/408 moving events to DLQ.

packages/nemo-evaluator/src/nemo_evaluator/telemetry.py
```diff
@@ -0,0 +1,455 @@
...
+                "ts": _get_iso_timestamp(queued.timestamp),
+                "parameters": queued.event.model_dump(by_alias=True),
+                "name": queued.event._event_name,
+            }
+            for queued in events
+        ],
+    }
+
+
+class TelemetryHandler:
```
pribalta, line 232: [Power Review] ⚠️ **Mixed sync/async paradigms within TelemetryHandler** · 72% confidence

TelemetryHandler exposes a synchronous public API (start, stop, enqueue, flush) but uses async internals (_astart, _astop, _flush_events, _send_events) bridged by _run_sync with two distinct execution paths (asyncio.run vs ThreadPoolExecutor). The knowledge document warns about 'callbacks mixed with async/await' as a mixed-paradigm signal. The _run_sync bridging method (lines 330–344) creates a new ThreadPoolExecutor on each invocation when a running loop exists, adding complexity and potential resource concerns.

💡 **Suggestion:** Consider making TelemetryHandler consistently synchronous by using httpx.Client (sync) instead of httpx.AsyncClient for HTTP calls. The flush interval timer could use threading.Timer or a background thread instead of asyncio. This eliminates the sync/async bridging entirely and simplifies the handler, since all callers (evaluate()) are synchronous.

packages/nemo-evaluator/tests/integration_tests/test_telemetry_e2e.py
```diff
@@ -0,0 +1,89 @@
...
+from nemo_evaluator.api.api_dataclasses import (
+    ConfigParams,
+    Evaluation,
+    EvaluationConfig,
+    EvaluationTarget,
+)
+from nemo_evaluator.core.evaluate import evaluate
+
+
+def _mock_eval(tmp_path):
```
pribalta, line 44: [Power Review] ⚠️ **Flaky test pattern: network dependency in integration tests** · 70% confidence

Per the Testing Quality knowledge document (WARNING: Flaky Test Patterns): 'tests that depend on ... network availability ... pass locally but fail in CI.' The integration conftest enables real HTTP sends to a staging endpoint. While `TELEMETRY_ALLOW_SEND_FAILURES` provides an escape hatch, by default these tests will fail without network access to the staging endpoint, making them flaky in restricted CI environments.

💡 **Suggestion:** Consider documenting the network requirement more prominently (e.g., a pytest marker like `@pytest.mark.network`) so CI can selectively skip these tests, or default `TELEMETRY_ALLOW_SEND_FAILURES=1` and only enforce send success in a dedicated telemetry CI job.

packages/nemo-evaluator/tests/unit_tests/test_telemetry.py
```diff
@@ -0,0 +1,324 @@
...
+        """Test that handler silently fails on invalid events."""
+        from nemo_evaluator.config import TelemetryLevel
+        from nemo_evaluator.telemetry import TelemetryHandler
+
+        handler = TelemetryHandler(telemetry_level=TelemetryLevel.DEFAULT)
+        handler.enqueue("not an event")  # type: ignore
+        assert len(handler._events) == 0
+
+
+class TestBuildPayload:
```
pribalta, line 260: [Power Review] 💡 **Missing boundary test: build_payload with empty events list** · 75% confidence

Per the Testing Quality knowledge document (SUGGESTION: Missing Boundary Tests): 'add tests at boundary values for each input parameter.' `build_payload()` accesses `events[0].event._schema_version` which would raise `IndexError` on an empty list. There is no test for this boundary condition.

💡 **Suggestion:** Add a test that calls `build_payload([], ...)` and verifies the expected behavior (either an `IndexError` is raised or the function handles it gracefully).

packages/nemo-evaluator/src/nemo_evaluator/telemetry.py
```diff
@@ -0,0 +1,455 @@
...
+        "deviceGdprFuncOptIn": "None",  # do not change
+        "deviceGdprTechOptIn": "None",  # do not change
+        "deviceId": "undefined",  # do not change
+        "deviceMake": "undefined",  # do not change
+        "deviceModel": "undefined",  # do not change
+        "deviceOS": "undefined",  # do not change
+        "deviceOSVersion": "undefined",  # do not change
+        "deviceType": "undefined",  # do not change
+        "eventProtocol": "1.6",  # do not change
+        "eventSchemaVer": events[0].event._schema_version,
```
pribalta, line 208: [Power Review] 💡 **Missing empty-list guard in public build_payload function** · 70% confidence

The public function `build_payload` accesses `events[0].event._schema_version` on line 207 without checking that `events` is non-empty. While current internal callers guard against empty lists (`_flush_events` checks `if events_to_send:`, and `_send_events_with_client` checks `if not events: return`), the function itself is public (no underscore prefix) and will raise an `IndexError` if called with an empty list. Per the knowledge document: 'AI skips validation on user input, API payloads... Validate at every trust boundary.'

💡 **Suggestion:** Add an early guard at the top of `build_payload`: `if not events: raise ValueError('events must not be empty')` — or return an empty/default payload. This makes the function safe to call independently and documents the precondition explicitly.

packages/nemo-evaluator/tests/unit_tests/test_telemetry.py
pribalta, line 118: [Power Review] 💡 **Implicit assertion via exception instead of explicit assert** · 70% confidence

Per the knowledge document: 'Every test must assert an expected outcome.' This test calls uuid.UUID(session_id) which raises on invalid input, but uses no explicit assert statement. While it does verify behavior implicitly, an explicit assertion makes the test's intent clearer and produces better failure messages.

💡 **Suggestion:** Add an explicit assertion, e.g.: `assert uuid.UUID(session_id) is not None` or `assert isinstance(uuid.UUID(session_id), uuid.UUID)`.

packages/nemo-evaluator/tests/unit_tests/test_telemetry.py
pribalta, line 143: [Power Review] 💡 **Implicit assertion via exception instead of explicit assert** · 70% confidence

Per the knowledge document: 'Every test must assert an expected outcome.' Same pattern as test_session_id_generation — uuid.UUID() call verifies implicitly but lacks an explicit assert statement.

💡 **Suggestion:** Add an explicit assertion, e.g.: `assert uuid.UUID(session_id) is not None`.

packages/nemo-evaluator/src/nemo_evaluator/telemetry.py
```diff
@@ -0,0 +1,455 @@
...
+
+def get_session_id() -> str:
+    """Get session ID from environment or generate a new one."""
+    return os.getenv(TELEMETRY_SESSION_ID_ENV_VAR) or _generate_session_id()
+
+
+_TELEMETRY_TAG = "\033[1;33m[TELEMETRY]\033[0m"
+
+
+def show_telemetry_notification() -> None:
```
pribalta, line 102: [Power Review] 💡 **Core package references launcher CLI command in log message** · 70% confidence

The show_telemetry_notification() function in the core package (nemo-evaluator) hardcodes the launcher CLI command string 'nemo-evaluator-launcher config set telemetry.level'. Per the project architecture, the one-way dependency rule states that core must not depend on launcher. While this is a string reference rather than an import, it creates conceptual coupling — if the launcher CLI changes its command syntax, the core log message becomes stale.

💡 **Suggestion:** Consider making the notification message generic (e.g., 'See documentation for configuring telemetry levels') or have the launcher inject its own notification at the orchestration layer where it already depends on core.

packages/nemo-evaluator/src/nemo_evaluator/telemetry.py
```diff
@@ -0,0 +1,456 @@
...
+    """Generate a new UUID session ID."""
+    return str(uuid.uuid4())
+
+
+def get_session_id() -> str:
+    """Get session ID from environment or generate a new one."""
+    return os.getenv(TELEMETRY_SESSION_ID_ENV_VAR) or _generate_session_id()
+
+
+_TELEMETRY_TAG = "\033[1;33m[TELEMETRY]\033[0m"
```
marta-sd, line 99: nit: we should use our coloring function instead

marta-sd: Edit: we can't because they are defined in launcher, not core

PR #780, opened 2026-02-27
feat: enable progress tracking in status CLI (EVAL-377)

gchlebus:
## Summary

Enable progress tracking (as # of processed requests) in the `nel status` CLI command (EVAL-377). The infrastructure already existed but was disabled as WIP. This PR unhides it, fixes accuracy issues, and switches from percentages to raw request counts.

Screenshot 2026-03-03 at 22 15 49

### Problem

1. **Progress was hidden** — the `status` command stripped the `progress` field from output (WIP guard).
2. **Percentage-based progress was broken for meta-tasks** — MMLU has 57 subtasks, so `limit_samples=20` produced 1140 requests against a denominator of 20, showing 5700%.
3. **Retry and cache inflation** — failed requests (non-200) and cached replays in auto-chained Slurm jobs inflated the count.

### Solution

**Display raw request count instead of percentages.** This works universally across single-turn, multi-turn, meta-tasks (MMLU), and n-repeat evaluations without needing to compute dataset sizes.

#### CLI (`status.py`)
- Add **"Requests Processed"** column to the status table (between Status and executor info)
- Include progress in JSON output (removed `pop("progress")` WIP guard)
- Format: raw integer or `"-"` for unknown/missing
- Rename `_format_progress` → `_format_requests_processed`

#### Progress Tracking Interceptor (`progress_tracking_interceptor.py`)
- **Skip `cache_hit` responses** — prevents double-counting in auto-chained Slurm jobs where cached responses are replayed
- **Skip non-200 responses** — prevents retry inflation (failed attempts return non-200)
- Rename log fields from `samples_processed` → `requests_processed` for clarity

#### Executors (local + slurm)
- Simplify `_get_progress()` to return `Optional[int]` (raw request count) instead of `Optional[float]` (ratio)
- `_get_progress` no longer calls `_get_dataset_size` / parses `run_config.yml` (functions kept for future reuse)

### Breaking Changes

- **`status --json` output**: `progress` field is now an `int` (request count) instead of `float` (0.0–1.0 ratio). Confirm no downstream consumers depend on the old format.
- **Log field rename**: `samples_processed` → `requests_processed` in structured logs from the progress tracking interceptor.
- **Progress tracking HTTP updates (POSTs to `progress_tracking_url`)**: The values reported are now lower/more accurate. Previously the counter incremented on every response including cache hits and failed retries; now it only counts successful (HTTP 200), non-cached responses. The POST mechanism and payload format are unchanged — only the reported numbers differ.

### DIFF
--- a/packages/nemo-evaluator/src/nemo_evaluator/adapters/interceptors/progress_tracking_interceptor.py
+++ b/packages/nemo-evaluator/src/nemo_evaluator/adapters/interceptors/progress_tracking_interceptor.py
@@ -212,6 +212,11 @@ class ProgressTrackingInterceptor(ResponseInterceptor, PostEvalHook):
         # NOTE: log field renamed from samples_processed → requests_processed
         # to reflect that the counter tracks API requests, not evaluation samples.
         if ar.rctx.cache_hit or ar.r.status_code != 200:
+            self.logger.debug(
+                "Skipping progress count",
+                cache_hit=ar.rctx.cache_hit,
+                status_code=ar.r.status_code,
+            )
             return ar
 
         curr_samples = 0


packages/nemo-evaluator/src/nemo_evaluator/adapters/interceptors/progress_tracking_interceptor.py
```diff
@@ -205,14 +205,23 @@ def _update_on_timer(self, interval_seconds: float):
...
         self, ar: AdapterResponse, context: AdapterGlobalContext
     ) -> AdapterResponse:
+        # Only count successful, non-cached responses as genuine progress.
+        # This avoids inflating the count from retries (failed attempts return
+        # non-200) and from cache replays during auto-chained Slurm jobs
+        # (cached responses carry cache_hit=True).
+        # NOTE: log field renamed from samples_processed → requests_processed
+        # to reflect that the counter tracks API requests, not evaluation samples.
+        if ar.rctx.cache_hit or ar.r.status_code != 200:
+            return ar
```
wprazuch, line 220: I would add `self.logger.debug` to note it is being skipped

gchlebus: Added in 1e9d6e48 — logs `cache_hit` and `status_code` at debug level when skipping the progress count.

PR #786, opened 2026-03-01
feat: add local no-docker mode for nemo-evaluator-launcher

mishana:
## Summary
- add `--no-docker` flag to `nemo-evaluator-launcher run`
- add `execution.use_docker` (default: `true`) for local executor config
- support host-process execution path (no `docker run`) in local executor scripts
- keep Docker as default behavior for backward compatibility
- add tests and docs for no-docker mode

## Rationale
This change enables running `nemo-evaluator-launcher` in environments where users do not have root access and cannot run Docker at all. A concrete example is workloads running on GKE Autopilot pods, where Docker-in-containerd patterns are prohibited.

This is especially useful for R&D workflows where teams want to run evaluations inside managed, restricted compute environments while still using the launcher orchestration and task config experience.

## Notes
- no-docker mode is restricted to `execution.type=local` with `deployment.type=none`
- existing Docker-based local flows remain unchanged by default

## Validation
- `uv run pytest tests/unit_tests/test_get_eval_factory_command.py tests/unit_tests/test_local_executor.py tests/unit_tests/test_cli_integration.py -q`
- `uv run ruff check ...` on modified files

### DIFF
--- a/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/helpers.py
+++ b/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/helpers.py
@@ -16,6 +16,7 @@
 import base64
 import datetime
 import os
+import shlex
 from dataclasses import dataclass
 from typing import Optional
 
@@ -61,8 +62,9 @@ def _str_to_echo_command(str_to_save: str, filename: str) -> CmdAndReadableComme
     debug_str = "\n".join(
         [f"# Contents of {filename}"] + ["# " + s for s in str_to_save.splitlines()]
     )
+    quoted_filename = shlex.quote(filename)
     return CmdAndReadableComment(
-        cmd=f'echo "{str_to_save_b64}" | base64 -d > {filename}', debug=debug_str
+        cmd=f'echo "{str_to_save_b64}" | base64 -d > {quoted_filename}', debug=debug_str
     )
 
 
@@ -167,6 +169,7 @@ def get_eval_factory_command(
     cfg: DictConfig,
     user_task_config: DictConfig,
     task_definition: dict,
+    output_dir: str = CONTAINER_RESULTS_DIR,
 ) -> CmdAndReadableComment:
     # This gets the eval_factory_config merged from both top-level and task-level.
     merged_nemo_evaluator_config = get_eval_factory_config(
@@ -214,7 +217,7 @@ def get_eval_factory_command(
     _set_nested_optionally_overriding(
         merged_nemo_evaluator_config,
         ["config", "output_dir"],
-        CONTAINER_RESULTS_DIR,
+        output_dir,
     )
     api_key_name = get_api_key_name(cfg)
     if api_key_name:
@@ -275,7 +278,7 @@ def get_eval_factory_command(
     if config_path:
         create_unresolved_config_cmd = _str_to_echo_command(
             open(config_path, "r").read(),
-            filename=f"{CONTAINER_RESULTS_DIR}/launcher_unresolved_config.yaml",
+            filename=f"{output_dir}/launcher_unresolved_config.yaml",
         )
         commands.append(create_unresolved_config_cmd.cmd)
         debug.append(create_unresolved_config_cmd.debug)
--- a/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/executor.py
+++ b/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/executor.py
@@ -23,6 +23,7 @@ import pathlib
 import platform
 import shlex
 import shutil
+import signal
 import subprocess
 import time
 from typing import Iterator, List, Optional, Tuple, Union
@@ -83,11 +84,17 @@ class LocalExecutor(BaseExecutor):
         Raises:
             RuntimeError: If the run script fails.
         """
+        use_docker = bool(cfg.execution.get("use_docker", True))
+
         # Check if docker is available (skip in dry_run mode)
-        if not dry_run and shutil.which("docker") is None:
+        if use_docker and not dry_run and shutil.which("docker") is None:
             raise RuntimeError(
                 "Docker is not installed or not in PATH. "
                 "Please install Docker to run local evaluations."
+            )
+        if not use_docker and cfg.deployment.type != "none":
+            raise ValueError(
+                "execution.use_docker=false is only supported with deployment.type=none."
             )
 
         # Generate invocation ID for this evaluation run
@@ -194,11 +201,15 @@ class LocalExecutor(BaseExecutor):
             dataset_mount_container = None
             dataset_env_var_value = None
             if "dataset_dir" in task:
-                dataset_mount_host = task["dataset_dir"]
-                # Get container mount path (default to /datasets if not specified)
-                dataset_mount_container = task.get("dataset_mount_path", "/datasets")
-                # Set NEMO_EVALUATOR_DATASET_DIR to the container mount path
-                dataset_env_var_value = dataset_mount_container
+                if use_docker:
+                    dataset_mount_host = task["dataset_dir"]
+                    # Get container mount path (default to /datasets if not specified)
+                    dataset_mount_container = task.get("dataset_mount_path", "/datasets")
+                    # Set NEMO_EVALUATOR_DATASET_DIR to the container mount path
+                    dataset_env_var_value = dataset_mount_container
+                else:
+                    # In no-docker mode, pass dataset_dir directly to local process.
+                    dataset_env_var_value = task["dataset_dir"]
 
             # Build env_groups for secrets file generation
             env_groups = {}
@@ -225,7 +236,12 @@ class LocalExecutor(BaseExecutor):
             task_output_dir = output_dir / task.name
             task_output_dir.mkdir(parents=True, exist_ok=True)
             eval_factory_command_struct = get_eval_factory_command(
-                cfg, task, task_definition
+                cfg,
+                task,
+                task_definition,
+                output_dir=(
+                    "/results" if use_docker else str(task_output_dir / "artifacts")
+                ),
             )
             eval_factory_command = eval_factory_command_struct.cmd
             # The debug comment for placing into the script and easy debug. Reason
@@ -257,6 +273,7 @@ class LocalExecutor(BaseExecutor):
                 "dataset_mount_host": dataset_mount_host,
                 "dataset_mount_container": dataset_mount_container,
                 "dataset_env_var_value": dataset_env_var_value,
+                "run_with_docker": use_docker,
             }
             evaluation_tasks.append(evaluation_task)
 
@@ -271,6 +288,7 @@ class LocalExecutor(BaseExecutor):
                     evaluation_tasks=[evaluation_task],
                     auto_export_destinations=auto_export_destinations,
                     extra_docker_args=extra_docker_args,
+                    has_docker_tasks=use_docker,
                 ).rstrip("\n")
                 + "\n"
             )
@@ -288,6 +306,7 @@ class LocalExecutor(BaseExecutor):
                     evaluation_tasks=evaluation_tasks,
                     auto_export_destinations=auto_export_destinations,
                     extra_docker_args=extra_docker_args,
+                    has_docker_tasks=use_docker,
                 ).rstrip("\n")
                 + "\n"
             )
@@ -386,8 +405,13 @@ class LocalExecutor(BaseExecutor):
                     executor="local",
                     data={
                         "output_dir": str(evaluation_task["output_dir"]),
-                        "container": evaluation_task["client_container_name"],
+                        "container": (
+                            evaluation_task["client_container_name"]
+                            if use_docker
+                            else ""
+                        ),
                         "eval_image": evaluation_task["eval_image"],
+                        "use_docker": use_docker,
                     },
                     config=OmegaConf.to_object(cfg),
                 )
@@ -711,33 +735,46 @@ class LocalExecutor(BaseExecutor):
                 f"Job {job_id} is not a local job (executor: {job_data.executor})"
             )
 
-        # Get container name from database
-        container_name = job_data.data.get("container")
-        if not container_name:
-            raise ValueError(f"No container name found for job {job_id}")
+        use_docker = bool(job_data.data.get("use_docker", True))
+        output_dir = pathlib.Path(job_data.data.get("output_dir", ""))
+        container_name = job_data.data.get("container") or ""
 
         killed_something = False
 
-        # First, try to stop the Docker container if it's running
-        result = subprocess.run(
-            shlex.split(f"docker stop {container_name}"),
-            capture_output=True,
-            text=True,
-            timeout=30,
-        )
-        if result.returncode == 0:
-            killed_something = True
-        # Don't raise error if container doesn't exist (might be still pulling)
+        # Try to stop script process group if a pid file is present.
+        pid_file = output_dir / "logs" / "stage.pid"
+        if pid_file.exists():
+            try:
+                pid = int(pid_file.read_text().strip())
+                if hasattr(os, "killpg"):
+                    os.killpg(pid, signal.SIGTERM)
+                else:
+                    os.kill(pid, signal.SIGTERM)
+                killed_something = True
+            except (OSError, ValueError):
+                pass
 
-        # Find and kill Docker processes for this container
-        result = subprocess.run(
-            shlex.split(f"pkill -f 'docker run.*{container_name}'"),
-            capture_output=True,
-            text=True,
-            timeout=10,
-        )
-        if result.returncode == 0:
-            killed_something = True
+        if use_docker and container_name:
+            # First, try to stop the Docker container if it's running
+            result = subprocess.run(
+                shlex.split(f"docker stop {container_name}"),
+                capture_output=True,
+                text=True,
+                timeout=30,
+            )
+            if result.returncode == 0:
+                killed_something = True
+            # Don't raise error if container doesn't exist (might be still pulling)
+
+            # Find and kill Docker processes for this container
+            result = subprocess.run(
+                shlex.split(f"pkill -f 'docker run.*{container_name}'"),
+                capture_output=True,
+                text=True,
+                timeout=10,
+            )
+            if result.returncode == 0:
+                killed_something = True
 
         # If we successfully killed something, mark as killed
         if killed_something:
@@ -758,7 +795,13 @@ class LocalExecutor(BaseExecutor):
         # Use common helper to get informative error message based on job status
         current_status = status_list[0].state if status_list else None
         error_msg = LocalExecutor.get_kill_failure_message(
-            job_id, f"container: {container_name}", current_status
+            job_id,
+            (
+                f"container: {container_name}"
+                if container_name
+                else f"pid_file: {pid_file}"
+            ),
+            current_status,
         )
         raise RuntimeError(error_msg)
 
--- a/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/executor.py
+++ b/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/executor.py
@@ -68,6 +68,61 @@ from nemo_evaluator_launcher.executors.base import (
 from nemo_evaluator_launcher.executors.registry import register_executor
 
 
+def _get_local_available_tasks() -> dict[str, set[str]]:
+    """Return locally installed NeMo Evaluator tasks grouped by harness."""
+    try:
+        from nemo_evaluator.api import get_available_evaluations
+    except ImportError as e:
+        raise RuntimeError(
+            "execution.use_docker=false requires `nemo-evaluator` to be installed locally. "
+            "Install nemo-evaluator (with the harness/task wheels you need), or enable Docker execution."
+        ) from e
+
+    framework_task_mapping, _, _ = get_available_evaluations()
+    return {
+        framework: set(tasks.keys())
+        for framework, tasks in framework_task_mapping.items()
+    }
+
+
+def _validate_task_available_locally(
+    *,
+    task_query: str,
+    task_definition: dict,
+    available_tasks_by_harness: dict[str, set[str]],
+) -> None:
+    """Validate that a task exists in locally installed NeMo Evaluator packages."""
+    harness_name = str(task_definition.get("harness") or "")
+    task_name = str(task_definition.get("task") or "")
+
+    if harness_name:
+        harness_tasks = available_tasks_by_harness.get(harness_name)
+        if harness_tasks is None:
+            available_harnesses = sorted(available_tasks_by_harness.keys())
+            raise ValueError(
+                f"Task '{task_query}' requires harness '{harness_name}', but this harness is not installed locally. "
+                f"Installed harnesses: {available_harnesses or ['<none>']}. "
+                "Install the corresponding NeMo Evaluator wheel, or run with Docker."
+            )
+        if task_name not in harness_tasks:
+            available_tasks = sorted(harness_tasks)
+            raise ValueError(
+                f"Task '{task_query}' is not available in installed harness '{harness_name}'. "
+                f"Available tasks in this harness: {available_tasks or ['<none>']}. "
+                "Install a wheel that contains this task, or run with Docker."
+            )
+        return
+
+    matching_harnesses = [
+        harness for harness, tasks in available_tasks_by_harness.items() if task_name in tasks
+    ]
+    if not matching_harnesses:
+        raise ValueError(
+            f"Task '{task_query}' is not available in locally installed NeMo Evaluator packages. "
+            "Install a wheel that contains this task, or run with Docker."
+        )
+
+
 @register_executor("local")
 class LocalExecutor(BaseExecutor):
     @classmethod
@@ -96,6 +151,9 @@ class LocalExecutor(BaseExecutor):
             raise ValueError(
                 "execution.use_docker=false is only supported with deployment.type=none."
             )
+        local_available_tasks: dict[str, set[str]] | None = None
+        if not use_docker:
+            local_available_tasks = _get_local_available_tasks()
 
         # Generate invocation ID for this evaluation run
         invocation_id = generate_invocation_id()
@@ -143,6 +201,12 @@ class LocalExecutor(BaseExecutor):
                 container=task.get("container"),
                 endpoint_type=task.get("endpoint_type"),
             )
+            if not use_docker:
+                _validate_task_available_locally(
+                    task_query=task.name,
+                    task_definition=task_definition,
+                    available_tasks_by_harness=local_available_tasks or {},
+                )
 
             # Track unlisted tasks for safeguard check
             if task_definition.get("is_unlisted", False):
--- a/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/run.template.sh
+++ b/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/run.template.sh
@@ -14,8 +14,10 @@
 # limitations under the License.
 
 
-# check if docker exists
+# check if docker exists when any task uses docker
+{% if has_docker_tasks %}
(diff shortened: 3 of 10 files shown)


prokotg:
hi @mishana, thank you for your contribution! Really appreciated!

This looks overall very wel. One important missing piece is that when someone uses `nel` with `--no-docker`, it is now assumed that the appropriate package is installed (the task from the config is available in the python environment). In fact, right now we're re-using container process in which we check if the task is available in any container and then pull that particular container (except we do not pull). Here, we should utilise Nemotron Evaluator Core API to check if the user has a wheel that contains appropriate task. Please let us know if that's something you'd like to contribute as well (I can point to relevant resources)

mishana:
Addressed. I added a no-docker preflight validation that uses NeMo Evaluator Core API (`nemo_evaluator.api.get_available_evaluations`) to verify each requested harness/task is actually installed locally before execution.

Details:
- `execution.use_docker=false` now checks local harness/task availability early and fails with actionable install guidance when missing.
- Added unit coverage for success/failure paths in no-docker mode.
- Updated local executor docs to explicitly call out the local harness-wheel requirement.

This is now in commit `ff8c248` on this PR branch.

prokotg:
/ok to test [665f689](https://github.com/NVIDIA-NeMo/Evaluator/pull/786/commits/665f689d72587fa04a5e7c5676576683e1b8469e)

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/helpers.py
prokotg, line 202:
```suggestion
    if task_definition.get("is_unlisted") or (cfg.execution.type == "local" and cfg.execution.use_docker):
```

Normally, we send only `task` invocation because that's unique to a container. Now, in a mixed environment, that's no longer the case.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/executor.py
```diff
@@ -711,33 +799,46 @@ def kill_job(job_id: str) -> None:
...
-        result = subprocess.run(
-            shlex.split(f"pkill -f 'docker run.*{container_name}'"),
-            capture_output=True,
-            text=True,
-            timeout=10,
-        )
-        if result.returncode == 0:
-            killed_something = True
+        # Try to stop script process group if a pid file is present.
+        pid_file = output_dir / "logs" / "stage.pid"
```
prokotg, line 809: This unfortunately introduces TOCTOU vulnerability. If anything happens to `stage.pid` file (overwrite, symlink race) we risk killing an alien process or other malicious actions. Since we operate on third party process, we cannot use typical `flock` approach. If we were to stay with PID file, then checking start time / process info before killing could be beneficial. For the scope of this MR, it's okay to skip this

prokotg (changes requested):
For now, TOCTOU is not obligatory. There is one suggestion which is important. Please fix linting, we'll run tests and hopefully have it merged. Thank you!

pribalta (commented):
## Power Review

> Reviewed 10 files, left 18 suggestions: 1 error, 12 warnings, 5 suggestions.

### Dimensions Analyzed

| Dimension | 🔴 | ⚠️ | 💡 |
| ai-code-hygiene/architecture | 0 | 2 | 0 |
| ai-code-hygiene/patterns | 0 | 1 | 0 |
| code-correctness/api-contracts | 0 | 1 | 0 |
| code-correctness/boundary-conditions | 1 | 1 | 2 |
| code-correctness/concurrency | 0 | 1 | 0 |
| code-correctness/data-integrity | 0 | 3 | 0 |
| code-correctness/error-handling | 0 | 1 | 2 |
| code-correctness/logic-correctness | 0 | 1 | 1 |
| code-correctness/resource-lifecycle | 0 | 1 | 0 |
| **Total** | **1** | **12** | **5** |

---
> *Across the add local no-docker mode files we swept,*
> *And 18 suggestions we have kept.*
> *AI may code fast and read at scale,*
> *But human judgment tips the final scale.*

⏱️ Reviewed in **6.9 min** (est. manual review: ~38 min, **6x faster**)

---

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/executor.py
```diff
@@ -711,33 +799,46 @@ def kill_job(job_id: str) -> None:
...
+        if pid_file.exists():
+            try:
+                pid = int(pid_file.read_text().strip())
```
pribalta, line 812: [Power Review] 🔴 **Missing validation that PID is positive before os.killpg — PID 0 kills caller's process group** · 95% confidence

Per the knowledge document (Missing Input Validation, CWE-20): data arriving from outside the trust boundary (file read) is used without checking range. The PID is read from a file (`stage.pid`) and passed directly to `os.killpg(pid, signal.SIGTERM)`. If the file contains `0`, `os.killpg(0, SIGTERM)` sends SIGTERM to every process in the *calling* process group — potentially killing the launcher itself and sibling processes. A negative value would also be invalid. The `int()` conversion alone does not guard against these dangerous values.

💡 **Suggestion:** After parsing the PID, validate it is strictly positive before use: `if pid <= 0: raise ValueError(f'Invalid PID: {pid}')`. This should be placed before the `os.killpg()` call and within the existing try/except block so that invalid values are caught gracefully.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/executor.py
```diff
@@ -711,33 +799,46 @@ def kill_job(job_id: str) -> None:
...
+                if hasattr(os, "killpg"):
+                    os.killpg(pid, signal.SIGTERM)
+                else:
+                    os.kill(pid, signal.SIGTERM)
+                killed_something = True
+            except (OSError, ValueError):
```
pribalta, line 818: [Power Review] ⚠️ **Swallowed exception in PID-based kill path** · 90% confidence

Per the knowledge document (ERROR: Swallowed Exception With No Action): 'An exception is caught and silently ignored. The caller has no way to know the operation failed.' Here, `except (OSError, ValueError): pass` discards all diagnostic information when the PID file read or process kill fails. A `ValueError` from `int()` (corrupt PID file) or an `OSError` other than 'no such process' (e.g., EPERM — permission denied) would be silently lost, making debugging kill failures very difficult. Downgraded to warning because the `killed_something` flag does communicate functional outcome and the code falls through to Docker-based killing as a secondary strategy.

💡 **Suggestion:** At minimum, log the exception at debug/warning level so operators can diagnose kill failures. Consider also narrowing the OSError handling — e.g., only silencing `ProcessLookupError` (errno ESRCH) when the process is already dead, while letting permission errors propagate.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/run.template.sh
```diff
@@ -57,14 +59,17 @@ else
...
-    # Docker run with eval factory command
+    # Execute evaluation task
     (
         {% if task.secrets_env_content -%}
-        # Source secrets (scoped to subshell); re-exports happen before each docker run
+        # Source secrets (scoped to subshell)
         source "$task_dir/.secrets.env"
         {% endif -%}
 
+        echo "$$" > "$logs_dir/stage.pid"
```
pribalta, line 69: [Power Review] ⚠️ **$$` in subshell stores parent shell PID, not subshell PID** · 88% confidence

Per knowledge doc 'Wrong Variable or Field Referenced': `$$` inside a bash subshell `(...)` always expands to the PID of the top-level script process, not the subshell's PID. Since this `echo` is inside a `(...)` subshell block (starting at line 63), every task's `stage.pid` file will contain the same PID — the parent script's PID. The `kill_job` function in `executor.py` then calls `os.killpg(pid, signal.SIGTERM)` on this PID, which sends SIGTERM to the entire process group. In parallel execution mode, killing any single task would terminate ALL running tasks instead of just the targeted one. This is a stale/wrong value bug — the stored PID does not identify the process the code intends to kill.

💡 **Suggestion:** Use `$BASHPID` (available since Bash 4.0, 2009) instead of `$$` to capture the actual subshell PID. Additionally, consider whether `os.killpg` in `kill_job` is the correct signal mechanism — if the subshell shares its process group with the parent, you may need `os.kill(pid, signal.SIGTERM)` or have the subshell create its own process group (e.g., via `set -m` or wrapping with `setsid`).

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/executor.py
pribalta, line 818: [Power Review] ⚠️ **Swallowed exceptions in kill_job PID cleanup** · 85% confidence

The knowledge document states: 'Never swallow exceptions without logging or re-raising.' While the exceptions caught here are specific (OSError, ValueError) rather than bare, they are silently discarded with `pass`. An OSError could indicate a permission-denied condition (not just process-already-gone), and a ValueError could indicate a corrupt PID file — both worth logging for debuggability of kill failures.

💡 **Suggestion:** Add a debug/warning log for the caught exceptions. For example: `except (OSError, ValueError) as exc: logger.debug("Could not signal process from %s: %s", pid_file, exc)`. This preserves the non-fatal behavior while leaving a diagnostic trail.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/executor.py
pribalta, line 809: [Power Review] ⚠️ **Filesystem TOCTOU on pid_file.exists() then pid_file.read_text()** · 85% confidence

The knowledge document explicitly flags `os.path.exists(path)` then file operation as a Non-Atomic Check-Then-Act (TOCTOU) anti-pattern (CWE-367): 'filesystem TOCTOU — os.path.exists(path) then open(path) ... another process can replace the file between check and use. Instead: open first and handle the error.' Here, `pid_file.exists()` is checked on line 810, then `pid_file.read_text()` is called on line 812. The shell template's `trap 'rm -f "$logs_dir/stage.pid"' EXIT` (run.template.sh line 70) actively removes this file on process exit, creating a real window where the file disappears between check and read. While the `except (OSError, ValueError): pass` block on line 818 would catch the resulting FileNotFoundError, the `exists()` guard is redundant and constitutes a textbook TOCTOU pattern.

💡 **Suggestion:** Remove the `pid_file.exists()` check and rely solely on exception handling. This is both more correct and more concise:
```python
try:
    pid = int(pid_file.read_text().strip())
    if hasattr(os, 'killpg'):
        os.killpg(pid, signal.SIGTERM)
    else:
        os.kill(pid, signal.SIGTERM)
    killed_something = True
except (OSError, ValueError):
    pass
```

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/executor.py
```diff
@@ -83,12 +139,21 @@ def execute_eval(cls, cfg: DictConfig, dry_run: bool = False) -> str:
         Raises:
             RuntimeError: If the run script fails.
         """
+        use_docker = bool(cfg.execution.get("use_docker", True))
```
pribalta, line 142: [Power Review] ⚠️ **Geological layers: use_docker conditionals scattered throughout execute_eval** · 85% confidence

Per the knowledge document (Geological Layers): 'New code bolted on top of old code without integrating — duplicate data paths, parallel config mechanisms, wrapper-on-wrapper stacking.' The diff adds ~11 `use_docker` conditional branches scattered across `execute_eval` (lines 142, 145, 150, 155, 204, 268, 306, 340, 355, 373, 472). Each conditional duplicates a small decision about Docker vs host execution rather than abstracting the difference behind a clean boundary. This makes the method harder to reason about and will compound if more execution modes are added.

💡 **Suggestion:** Extract Docker-specific and host-specific task-building logic into separate private methods (e.g., `_build_docker_task_config()` and `_build_host_task_config()`) or introduce a lightweight strategy object. The main `execute_eval` method should orchestrate at a high level, delegating mode-specific details to these helpers. This keeps the conditional surface to one or two dispatch points instead of eleven.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/executor.py
```diff
@@ -711,33 +799,46 @@ def kill_job(job_id: str) -> None:
...
-        )
-        if result.returncode == 0:
-            killed_something = True
+        # Try to stop script process group if a pid file is present.
+        pid_file = output_dir / "logs" / "stage.pid"
+        if pid_file.exists():
+            try:
+                pid = int(pid_file.read_text().strip())
+                if hasattr(os, "killpg"):
+                    os.killpg(pid, signal.SIGTERM)
```
pribalta, line 814: [Power Review] ⚠️ **`os.killpg` assumes PID is a process group leader, but subshell inherits parent PGID** · 82% confidence

Per knowledge doc 'Wrong Variable or Field Referenced' and 'Stale or Cached Value Used After Update': `os.killpg(pid, signal.SIGTERM)` sends a signal to the process GROUP identified by `pid`. However, the subshell spawned by `(...)` inherits the parent script's process group — its PID is NOT a process group ID. If `$$` is fixed to `$BASHPID`, then `os.killpg(subshell_pid, ...)` would fail with ESRCH because the subshell PID is not a PGID. If `$$` remains as-is (parent PID), `os.killpg` kills the entire script's process group (all tasks). Either way, the kill mechanism does not correctly target a single task.

💡 **Suggestion:** Either: (1) use `os.kill(pid, signal.SIGTERM)` to signal just the subshell process (and rely on SIGTERM propagation or explicit child cleanup), or (2) have the subshell create its own process group in the template (e.g., prepend the subshell body with a `set -m` or use `setsid`) so that `os.killpg` is valid.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/executor.py
```diff
@@ -711,33 +799,46 @@ def kill_job(job_id: str) -> None:
                 f"Job {job_id} is not a local job (executor: {job_data.executor})"
             )
 
-        # Get container name from database
-        container_name = job_data.data.get("container")
-        if not container_name:
-            raise ValueError(f"No container name found for job {job_id}")
+        use_docker = bool(job_data.data.get("use_docker", True))
+        output_dir = pathlib.Path(job_data.data.get("output_dir", ""))
```
pribalta, line 803: [Power Review] ⚠️ **Empty output_dir from job database creates PID file path relative to cwd** · 80% confidence

Per the knowledge document (Unhandled Empty or Null Input): a function receives an empty string and proceeds without a guard. `job_data.data.get("output_dir", "")` returns an empty string if the key is missing (e.g., for job records created before this PR, or corrupted data). `pathlib.Path("")` resolves to the current working directory, so `pid_file` becomes `./logs/stage.pid`. If such a file coincidentally exists in the cwd, `kill_job` could read an unrelated PID and send SIGTERM to the wrong process group — a potentially destructive action.

💡 **Suggestion:** Validate that `output_dir` is non-empty before constructing the PID file path. For example: `if not output_dir.parts: raise ValueError(f"No output_dir found for job {job_id}")` or skip the PID-file kill path entirely when output_dir is empty.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/executor.py
```diff
@@ -225,7 +300,12 @@ def execute_eval(cls, cfg: DictConfig, dry_run: bool = False) -> str:
             task_output_dir = output_dir / task.name
             task_output_dir.mkdir(parents=True, exist_ok=True)
             eval_factory_command_struct = get_eval_factory_command(
-                cfg, task, task_definition
+                cfg,
+                task,
+                task_definition,
+                output_dir=(
+                    "/results" if use_docker else str(task_output_dir / "artifacts")
```
pribalta, line 307: [Power Review] ⚠️ **Hardcoded "/results" duplicates CONTAINER_RESULTS_DIR constant** · 80% confidence

The `get_eval_factory_command` function defines its default parameter as `output_dir: str = CONTAINER_RESULTS_DIR` (helpers.py line 172). However, the caller in executor.py passes the magic string `"/results"` for the Docker path instead of referencing the same constant. This is a Tight Coupling to Implementation Detail issue — if `CONTAINER_RESULTS_DIR` is ever updated, this hardcoded value will silently diverge, leading to path mismatches between the executor and the eval factory command configuration.

💡 **Suggestion:** Import and use the `CONTAINER_RESULTS_DIR` constant: `output_dir=(CONTAINER_RESULTS_DIR if use_docker else str(task_output_dir / "artifacts"))`. This ensures the Docker path always matches the function's documented default.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/executor.py
```diff
@@ -83,12 +139,21 @@ def execute_eval(cls, cfg: DictConfig, dry_run: bool = False) -> str:
         Raises:
```
pribalta, line 139: [Power Review] ⚠️ **Monolithic function: execute_eval spans 340+ lines** · 75% confidence

Per the knowledge document (Monolithic Functions >50 lines): 'AI generates long, flat functions instead of composing smaller units. Functions exceeding ~50 lines are harder to test, review, and maintain.' Based on the diff line numbers, `execute_eval` spans from approximately line 139 to at least line 480, making it well over 300 lines. This PR adds further branches and logic, exacerbating the problem.

💡 **Suggestion:** Extract cohesive blocks into well-named helpers: validation/preconditions (`_validate_execution_config()`), task iteration and definition resolution, script generation, and job execution. Each extracted method should be independently testable. The no-docker additions present a natural refactoring opportunity.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/executor.py
pribalta, line 809: [Power Review] ⚠️ **PID file not cleaned up after failed or successful kill in kill_job** · 75% confidence

Per the knowledge document's 'Temporary File or Directory Not Cleaned Up' rule: a temporary file is created but never deleted, especially on error paths. The kill_job method reads `stage.pid` and attempts os.killpg/os.kill. If the process is already dead (stale PID), the OSError is caught and silently ignored (line 818), but the PID file itself is never removed. Additionally, even on a successful SIGTERM send, the method proceeds immediately without waiting for the process's EXIT trap to fire and clean up the file. This leaves stale PID files that could cause subsequent kill_job calls to target a reused PID — an analog to the 'Use After Release' pattern in PID namespace.

💡 **Suggestion:** After the try/except block (around line 819), add cleanup of the PID file regardless of whether the kill succeeded or failed. For example: `try: pid_file.unlink(missing_ok=True) except OSError: pass`. This makes cleanup idempotent and prevents stale PID files from accumulating.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/executor.py
```diff
@@ -711,33 +799,46 @@ def kill_job(job_id: str) -> None:
...
+                else:
+                    os.kill(pid, signal.SIGTERM)
+                killed_something = True
+            except (OSError, ValueError):
+                pass
+
+        if use_docker and container_name:
+            # First, try to stop the Docker container if it's running
+            result = subprocess.run(
+                shlex.split(f"docker stop {container_name}"),
```
pribalta, line 824: [Power Review] ⚠️ **`shlex.split` on f-string with `container_name` may split malformed names** · 72% confidence

Per knowledge doc 'Injection: Untrusted Data in Executable Context' (CWE-78): `container_name` originates from stored job data (`job_data.data.get('container')`). While `subprocess.run` is called with a list (not `shell=True`), `shlex.split(f"docker stop {container_name}")` will split on whitespace and interpret shell quoting in `container_name`. A container name containing spaces or shell metacharacters would produce unexpected argument splitting. The same pattern applies to line 835 with `pkill -f`. Although this pattern existed pre-diff, the new code at line 804 (`container_name = job_data.data.get("container") or ""`) now allows empty string through, and the guard at line 821 prevents execution with empty names, but a name with embedded spaces/quotes would still be mis-split.

💡 **Suggestion:** Use an explicit argument list instead of `shlex.split` on an f-string: `subprocess.run(["docker", "stop", container_name], ...)`. This avoids the splitting ambiguity entirely.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/executor.py
pribalta, line 814: [Power Review] ⚠️ **os.killpg uses PID as PGID — only works if script is a process group leader** · 72% confidence

os.killpg(pgid, sig) sends a signal to an entire process group identified by its process group ID (PGID). The code reads a PID from a file (written by `echo "$$"` in the shell template) and passes it directly to os.killpg as if it were a PGID. This only works if the script process happens to be its own process group leader (PID == PGID), which is not guaranteed — especially when the script is launched via subprocess.run() without start_new_session=True, as the child inherits the parent's process group. For non-docker jobs, the PID-file mechanism is the *only* kill path (docker stop is skipped), so if killpg fails with ProcessLookupError (caught silently by except OSError), kill_job will always report failure for no-docker jobs. This matches the 'Incorrect Operator or Operand' error pattern — PID and PGID are related but distinct values.

💡 **Suggestion:** Either (a) use os.getpgid(pid) to resolve the actual PGID before calling os.killpg, or (b) use os.kill(pid, signal.SIGTERM) to signal the script process directly (though children may survive), or (c) ensure the script is started in its own process group (e.g., subprocess.Popen with start_new_session=True) so that PID == PGID is guaranteed.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/executor.py
```diff
@@ -136,6 +201,12 @@ def execute_eval(cls, cfg: DictConfig, dry_run: bool = False) -> str:
                 container=task.get("container"),
                 endpoint_type=task.get("endpoint_type"),
             )
+            if not use_docker:
+                _validate_task_available_locally(
+                    task_query=task.name,
+                    task_definition=task_definition,
+                    available_tasks_by_harness=local_available_tasks or {},
```
pribalta, line 208: [Power Review] 💡 **Defensive `or {}` is unreachable — consider assertion instead** · 85% confidence

At line 208, `available_tasks_by_harness=local_available_tasks or {}` uses `or {}` as a fallback. However, `local_available_tasks` is set to a dict at line 156 (inside `if not use_docker`), and `_validate_task_available_locally` is only called at line 204 (inside `if not use_docker`). So `local_available_tasks` is guaranteed to be a non-None dict at this point. The `or {}` can never activate, making it a tautological guard. This falls under the 'Redundant or Tautological Condition' suggestion pattern.

💡 **Suggestion:** Replace `local_available_tasks or {}` with just `local_available_tasks`, optionally preceded by `assert local_available_tasks is not None` to document the invariant.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/executor.py
pribalta, line 142: [Power Review] 💡 **Missing test for boundary between use_docker=true and use_docker=false in kill_job** · 80% confidence

Per the knowledge document (Missing Test for 'Just Inside / Just Outside' Pair): the `kill_job` method now branches on `use_docker` (line 802/821) but the test files in this diff do not include any tests for the `kill_job` path in no-docker mode — specifically testing: (1) kill with valid PID file, (2) kill with missing PID file, (3) kill with empty/invalid PID file content, (4) kill with use_docker=False and no container name. These boundary transitions should be exercised.

💡 **Suggestion:** Add test cases in `test_local_executor.py` covering `kill_job` for no-docker jobs: valid PID file, missing PID file, malformed PID file content (empty, non-numeric, zero, negative), and the transition where `use_docker=False` with no container name.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/executor.py
```diff
@@ -67,6 +68,61 @@
...
+
+
+def _validate_task_available_locally(
+    *,
+    task_query: str,
+    task_definition: dict,
+    available_tasks_by_harness: dict[str, set[str]],
+) -> None:
+    """Validate that a task exists in locally installed NeMo Evaluator packages."""
+    harness_name = str(task_definition.get("harness") or "")
```
pribalta, line 95: [Power Review] 💡 **Missing defensive handling when both harness and task name are empty in _validate_task_available_locally** · 70% confidence

Per the knowledge document (Lack of Defensive default/else Branch): the function handles two cases — `harness_name` is truthy (lines 98-114) and `harness_name` is falsy (lines 116-123). In the falsy branch, if `task_name` is also empty (both `harness` and `task` keys are missing or empty in `task_definition`), the code searches for `""` across all harnesses and produces a generic error message. A more explicit guard at the top of the function would provide a clearer, earlier error message for this malformed-input boundary.

💡 **Suggestion:** Add an early guard: `if not harness_name and not task_name: raise ValueError(f"Task '{task_query}' has no harness or task name defined in its task definition.")` before the main branching logic.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/run.template.sh
```diff
@@ -126,9 +131,19 @@ else
...
-    exit_code=$?
+        {% else %}
+        {% if task.eval_reexport_cmd -%}
+        # Re-export eval env vars to original names
+        {{ task.eval_reexport_cmd }}
+        {% endif -%}
+        {% if task.dataset_env_var_value -%}
+        export NEMO_EVALUATOR_DATASET_DIR="{{ task.dataset_env_var_value }}"
+        {% endif -%}
+        {{ task.eval_factory_command }} > "$logs_dir/client_stdout.log" 2>&1
```
pribalta, line 142: [Power Review] 💡 **Host eval command exit code not explicitly checked** · 70% confidence

Per the knowledge document (SUGGESTION: Shell Command Without Error Check): 'A shell script runs commands without checking the exit code.' In the no-docker branch, `{{ task.eval_factory_command }} > "$logs_dir/client_stdout.log" 2>&1` runs the evaluation command, and `exit_code=$?` captures its status. However, unlike the Docker path which has explicit success/failure echo statements and a structured exit, the no-docker path has no diagnostic logging on failure. If the eval command fails, the only record is in `client_stdout.log` — there's no stderr message indicating the command failed, its exit code, or guidance for the operator.

💡 **Suggestion:** Add a failure diagnostic similar to the Docker path, e.g.:
```bash
{{ task.eval_factory_command }} > "$logs_dir/client_stdout.log" 2>&1
exit_code=$?
if [ $exit_code -ne 0 ]; then
    echo "Evaluation command failed with exit code $exit_code" >&2
fi
```

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/local/executor.py
pribalta, line 803: [Power Review] 💡 **Empty-string fallback for output_dir could create relative PID file path** · 70% confidence

Per the knowledge document (WARNING: Error Message Without Context / ERROR: Null/None Dereference on Fallible Return): `job_data.data.get("output_dir", "")` defaults to an empty string. `pathlib.Path("") / "logs" / "stage.pid"` produces the relative path `logs/stage.pid`, which could inadvertently match a file in the process's current working directory. While `pid_file.exists()` guards against reading a nonexistent file, if such a relative path happens to exist, `kill_job` could send SIGTERM to an unrelated process.

💡 **Suggestion:** Validate that `output_dir` is non-empty and absolute before using it for PID file lookups. If missing, skip the PID-based kill or raise an error early:
```python
output_dir_str = job_data.data.get("output_dir", "")
if not output_dir_str:
    raise ValueError(f"No output directory found for job {job_id}")
output_dir = pathlib.Path(output_dir_str)
```

PR #792, opened 2026-03-02
feat: add benchmark-required capabilities

marta-sd:


packages/nemo-evaluator/tests/unit_tests/core/test_capabilities.py
```diff
@@ -0,0 +1,63 @@
...
+
+from nemo_evaluator.api.api_dataclasses import Evaluation
+from nemo_evaluator.api.capabilities import BENCHMARK_CAPABILITIES
+from nemo_evaluator.core.input import verify_capabilities
+from nemo_evaluator.core.utils import MisconfigurationError
+
+
+@pytest.mark.parametrize("capability", BENCHMARK_CAPABILITIES.keys())
+def test_capability_is_valid(capability):
+    "model_id" not in BENCHMARK_CAPABILITIES[capability].payload
```
pribalta: [Power Review] 🔴 **Boolean expression result discarded — test asserts nothing** · 100% confidence

Per the knowledge document rule 'Return Value of Pure Method Discarded': a pure expression is evaluated on a line by itself and its result is thrown away. The expression `"model_id" not in BENCHMARK_CAPABILITIES[capability].payload` produces a boolean but it is neither assigned nor asserted. The test will unconditionally pass regardless of whether `model_id` is present in the payload, defeating its purpose entirely.

💡 **Suggestion:** Add `assert` before the expression: `assert "model_id" not in BENCHMARK_CAPABILITIES[capability].payload`

pribalta: [Power Review] 🔴 **Expression statement instead of assertion — test never validates anything** · 98% confidence

The line `"model_id" not in BENCHMARK_CAPABILITIES[capability].payload` is a bare expression that evaluates to a boolean but is silently discarded. The test always passes regardless of the payload contents. This is a 'Use of Uninitialized Variable' analog — the test result is never captured or asserted, making the validation dead code. The intended invariant (that `model_id` is not pre-set in payloads) is never actually checked.

💡 **Suggestion:** Add `assert` at the beginning: `assert "model_id" not in BENCHMARK_CAPABILITIES[capability].payload`

packages/nemo-evaluator/src/nemo_evaluator/api/capabilities.py
```diff
@@ -0,0 +1,128 @@
+from typing import Any, Dict
+
+from pydantic import BaseModel, Field
+
+
+class Capability(BaseModel):
+    name: str = Field(description="Name of the benchmark capability, e.g. 'logprobs'")
+    description: str = Field(
```
pribalta, line 24: [Power Review] 🔴 **Type annotation says `str` but default is `None` — type mismatch** · 95% confidence

The `description` field is annotated as `str` but has `default=None`. This means the field can hold `None` at runtime, but the type annotation claims it is always a `str`. Downstream code (or type checkers) relying on the annotation will not guard against `None`, risking `AttributeError` or `TypeError`. This is a CWE-704 (Incorrect Type Conversion or Cast) violation per the knowledge document's 'Type Mismatch or Unsafe Cast' rule.

💡 **Suggestion:** Change the annotation to `Optional[str]` (i.e., `description: Optional[str] = Field(...)`) to accurately reflect that `None` is a valid value.

packages/nemo-evaluator/src/nemo_evaluator/api/capabilities.py
```diff
@@ -0,0 +1,128 @@
...
+        description="Description of the benchmark capability, e.g. 'Uses log probabilities to compute the score'",
+        default=None,
+    )
+    payload: Dict[str, Any] = Field(
```
pribalta, line 27: [Power Review] 🔴 **Type annotation says `Dict[str, Any]` but default is `None` — type mismatch** · 95% confidence

The `payload` field is annotated as `Dict[str, Any]` but has `default=None`. This is the same CWE-704 type mismatch as the `description` field. Critically, `verify_capabilities` in `input.py` calls `.payload.copy()` on this field without a `None` guard — if a `Capability` is ever created without an explicit payload, this will raise `AttributeError: 'NoneType' object has no attribute 'copy'`.

💡 **Suggestion:** Change the annotation to `Optional[Dict[str, Any]]` and add a `None` check before calling `.copy()` in `verify_capabilities`.

pribalta (commented):
## Power Review

> Reviewed 7 files, left 24 suggestions: 8 errors, 14 warnings, 2 suggestions.

### Dimensions Analyzed

| Dimension | 🔴 | ⚠️ | 💡 |
| api-contracts | 1 | 3 | 1 |
| boundary-conditions | 1 | 3 | 0 |
| data-integrity | 3 | 3 | 1 |
| error-handling | 2 | 2 | 0 |
| logic-correctness | 1 | 1 | 0 |
| resource-lifecycle | 0 | 2 | 0 |
| **Total** | **8** | **14** | **2** |

---
> *Across the add benchmark-required capabilities files we swept,*
> *And 24 suggestions we have kept.*
> *AI may code fast and read at scale,*
> *But human judgment tips the final scale.*

⏱️ Reviewed in **2.3 min** (est. manual review: ~25 min, **11x faster**)

---

packages/nemo-evaluator/src/nemo_evaluator/core/input.py
```diff
@@ -592,3 +593,42 @@ def validate_configuration(run_config: dict) -> Evaluation:
...
+    }
+    if evaluation.target.api_endpoint.api_key_name is not None:
+        headers["Authorization"] = (
+            f"Bearer {os.getenv(evaluation.target.api_endpoint.api_key_name)}"
+        )
+    for capability in evaluation.config.required_capabilities:
+        payload = BENCHMARK_CAPABILITIES[capability].payload.copy()
+        payload["model_id"] = model_id
+
+        response = requests.post(url, json=payload, headers=headers)
```
pribalta: [Power Review] 🔴 **HTTP request without timeout can hang indefinitely** · 95% confidence

Per the knowledge document's 'Missing Cleanup on Error Path' rule: resources acquired (network connections) are not bounded. `requests.post()` without a `timeout` parameter will block indefinitely if the endpoint is unresponsive, causing the entire evaluation pipeline to hang with no recovery path. This is also related to CWE-755 (Improper Handling of Exceptional Conditions).

💡 **Suggestion:** Add an explicit `timeout` parameter, e.g. `requests.post(url, json=payload, headers=headers, timeout=30)`.

packages/nemo-evaluator/src/nemo_evaluator/core/input.py
```diff
@@ -592,3 +593,42 @@ def validate_configuration(run_config: dict) -> Evaluation:
...
+            f"target.api_endpoint.model_id is not provided but the following capabilities are required: {', '.join(evaluation.config.required_capabilities)}"
+        )
+    url = evaluation.target.api_endpoint.url
+    model_id = evaluation.target.api_endpoint.model_id
+    headers = {
+        "Content-Type": "application/json",
+        "Accept": "application/json",
+    }
+    if evaluation.target.api_endpoint.api_key_name is not None:
+        headers["Authorization"] = (
```
pribalta: [Power Review] 🔴 **os.getenv returning None silently produces 'Bearer None' auth header** · 95% confidence

Per the knowledge document's 'Null / None Dereference on Fallible Return' (CWE-476) and 'Generic Fallback Value Hiding Errors' rules: `os.getenv()` returns `None` when the environment variable does not exist. The f-string silently interpolates this as the string `'Bearer None'`, which will cause authentication to fail at the endpoint. The resulting error message (line 629) will blame the capability check rather than the missing environment variable, making debugging extremely difficult.

💡 **Suggestion:** Check the return value of `os.getenv()` before using it. For example:
```python
api_key = os.getenv(evaluation.target.api_endpoint.api_key_name)
if api_key is None:
    raise MisconfigurationError(
        f"Environment variable '{evaluation.target.api_endpoint.api_key_name}' is not set "
        f"but is required for capability verification."
    )
headers["Authorization"] = f"Bearer {api_key}"
```

packages/nemo-evaluator/src/nemo_evaluator/api/api_dataclasses.py
```diff
@@ -36,7 +37,6 @@ class EndpointType(str, Enum):
     UNDEFINED = "undefined"
     CHAT = "chat"
     COMPLETIONS = "completions"
```
pribalta, line 39: [Power Review] 🔴 **Breaking change: removal of VLM enum value from public EndpointType** · 95% confidence

The `EndpointType.VLM = "vlm"` enum value is removed from a public interface that is exported via `__init__.py`. Any existing callers referencing `EndpointType.VLM` or serialized configurations containing `"vlm"` as an endpoint type will break. Per the knowledge document's 'Breaking Change to Public Interface' rule: 'A public function signature, return type, serialisation format, or endpoint URL is changed in a way that breaks existing callers... changed enum values.' The test changes confirm this was a previously valid and tested value.

💡 **Suggestion:** Either deprecate `VLM` with a warning instead of removing it outright, or add a migration path (e.g., a model_validator that maps `"vlm"` to an equivalent using the new capabilities system). If VLM is being replaced by the `image_url` capability, consider keeping the enum value with a deprecation warning for at least one release cycle.

packages/nemo-evaluator/src/nemo_evaluator/core/input.py
pribalta: [Power Review] 🔴 **No timeout on requests.post() — hangs indefinitely on unresponsive endpoint** · 95% confidence

CWE-400 (Uncontrolled Resource Consumption): The `requests.post()` call has no `timeout` parameter. If the endpoint being tested for capability hangs or is slow, `verify_capabilities` blocks the entire evaluation pipeline indefinitely. The knowledge document explicitly lists 'set timeouts' as the required mitigation under Uncontrolled Resource Consumption.

💡 **Suggestion:** Add a timeout parameter, e.g. `requests.post(url, json=payload, headers=headers, timeout=30)`. Consider making the timeout configurable.

pribalta: [Power Review] ⚠️ **Missing timeout on requests.post() risks indefinite hang** · 95% confidence

Per the knowledge document's 'Missing Timeout on Resource Acquisition' rule: 'A blocking acquire (connect, lock, queue.get) has no timeout, risking indefinite hangs if the resource is unavailable.' The `requests.post()` call on line 626 has no `timeout` parameter, so if the endpoint is unresponsive, the entire evaluation pipeline will hang indefinitely with no way to recover. This is a network connection — a blocking resource acquisition.

💡 **Suggestion:** Add an explicit timeout parameter, e.g. `requests.post(url, json=payload, headers=headers, timeout=30)`. Consider making the timeout configurable. Handle `requests.exceptions.Timeout` with a clear error message indicating which capability verification timed out.

packages/nemo-evaluator/tests/unit_tests/core/test_capabilities.py
pribalta: [Power Review] ⚠️ **Test expression is not asserted — boundary check is a no-op** · 95% confidence

The expression `"model_id" not in BENCHMARK_CAPABILITIES[capability].payload` evaluates to a boolean but the result is discarded. This test is meant to verify that payloads don't pre-contain `model_id` (which would be silently overwritten in `verify_capabilities`), but without `assert`, the test always passes regardless of the actual content. This leaves a critical boundary condition (payload key collision) unverified.

💡 **Suggestion:** Add `assert` keyword: `assert "model_id" not in BENCHMARK_CAPABILITIES[capability].payload`.

packages/nemo-evaluator/src/nemo_evaluator/api/capabilities.py
pribalta, line 24: [Power Review] ⚠️ **Type annotation mismatch: non-optional fields default to None** · 92% confidence

The `Capability` model declares `description: str` and `payload: Dict[str, Any]` (non-optional types), but both have `default=None`. This violates the Pydantic data contract — the type annotation says the field is always a `str`/`Dict`, but it can actually be `None`. Per the knowledge document's 'Serialisation / Deserialisation Field Mismatch' rule, this creates a mismatch between the declared model schema and actual runtime values. Callers relying on the type annotation (e.g., `capability.description.upper()`) will get `AttributeError` at runtime. The project also mandates MyPy strict mode per patterns.md, which would flag this.

💡 **Suggestion:** Change the annotations to `Optional[str]` and `Optional[Dict[str, Any]]` to match the `None` defaults, or remove the defaults if these fields are always required.

packages/nemo-evaluator/src/nemo_evaluator/core/input.py
pribalta: [Power Review] ⚠️ **Authorization header may contain `Bearer None` if env var is unset** · 92% confidence

When `api_key_name is not None`, the code calls `os.getenv(api_key_name)` which returns `None` if the environment variable doesn't exist. This produces `Bearer None` as the Authorization header value — a malformed token that will either silently fail authentication or produce confusing error messages. Per the knowledge document's 'Missing Required Parameter or Header' rule: 'An API call omits a required field... The call may succeed in dev (lenient server) but fail in production.' Also relates to 'Violated Precondition or Invariant' — the precondition that the env var is actually set is not validated.

💡 **Suggestion:** Check the result of `os.getenv()` for `None` and raise a `MisconfigurationError` with a clear message if the API key environment variable is not set, e.g.:
```python
api_key = os.getenv(evaluation.target.api_endpoint.api_key_name)
if api_key is None:
    raise MisconfigurationError(f"Environment variable '{evaluation.target.api_endpoint.api_key_name}' is not set...")
headers["Authorization"] = f"Bearer {api_key}"
```

pribalta: [Power Review] ⚠️ **os.getenv returns None, producing 'Bearer None' auth header** · 90% confidence

Per the knowledge document rule 'Wrong Return Value or Assignment': when `api_key_name` is set but the corresponding environment variable is absent, `os.getenv(...)` returns `None`. The f-string interpolation then produces the string `"Bearer None"`, which is silently sent as the Authorization header. This will cause the capability verification request to fail with a misleading 401/403 error instead of a clear misconfiguration error about the missing API key.

💡 **Suggestion:** Resolve the environment variable before interpolation and raise a `MisconfigurationError` if it is `None`. For example:
```python
api_key = os.getenv(evaluation.target.api_endpoint.api_key_name)
if not api_key:
    raise MisconfigurationError(
        f"Environment variable '{evaluation.target.api_endpoint.api_key_name}' is not set"
    )
headers["Authorization"] = f"Bearer {api_key}"
```

pribalta: [Power Review] ⚠️ **os.getenv() returns None when env var is missing, producing 'Bearer None' header** · 90% confidence

CWE-20 (Missing Input Validation): When `api_key_name` is set but the referenced environment variable doesn't exist, `os.getenv()` returns `None`, and the Authorization header becomes the literal string `'Bearer None'`. This sends a meaningless credential to the endpoint, which would fail with a status error whose message obscures the real problem (missing env var). Data from outside the trust boundary (environment variables) should be validated at the entry point.

💡 **Suggestion:** Validate the env var result before use: `api_key = os.getenv(evaluation.target.api_endpoint.api_key_name)` followed by a guard `if api_key is None: raise MisconfigurationError(f'Environment variable {api_key_name} is not set...')`.

packages/nemo-evaluator/src/nemo_evaluator/api/capabilities.py
```diff
@@ -0,0 +1,128 @@
+from typing import Any, Dict
+
+from pydantic import BaseModel, Field
+
+
+class Capability(BaseModel):
+    name: str = Field(description="Name of the benchmark capability, e.g. 'logprobs'")
+    description: str = Field(
+        description="Description of the benchmark capability, e.g. 'Uses log probabilities to compute the score'",
+        default=None,
```
pribalta: [Power Review] ⚠️ **Capability model type hints disagree with defaults — description and payload typed as non-Optional but default to None** · 88% confidence

Per the knowledge document's 'Null / None Dereference on Fallible Return' principle: `description: str = Field(..., default=None)` and `payload: Dict[str, Any] = Field(..., default=None)` declare non-Optional types but accept None. This misleads static type checkers and downstream code into assuming these fields are always populated, masking potential None dereferences (as seen in `input.py` line 623). Pydantic will silently accept None despite the type hint.

💡 **Suggestion:** Either make the fields required (remove `default=None`) or use `Optional[str]` / `Optional[Dict[str, Any]]` to match the actual nullability.

packages/nemo-evaluator/src/nemo_evaluator/core/input.py
```diff
@@ -592,3 +593,42 @@ def validate_configuration(run_config: dict) -> Evaluation:
...
+            f"Bearer {os.getenv(evaluation.target.api_endpoint.api_key_name)}"
+        )
+    for capability in evaluation.config.required_capabilities:
+        payload = BENCHMARK_CAPABILITIES[capability].payload.copy()
```
pribalta: [Power Review] ⚠️ **Potential None dereference on Capability.payload — model allows None default** · 85% confidence

Per the knowledge document's 'Null / None Dereference on Fallible Return' (CWE-476) rule: The `Capability` model declares `payload: Dict[str, Any] = Field(..., default=None)`, meaning `.payload` can be `None`. Calling `.copy()` on `None` raises `AttributeError`. While all current capabilities define payloads, the model explicitly permits `None`, creating a latent defect for future capability additions.

💡 **Suggestion:** Either: (1) Remove the `default=None` from the `payload` field in `Capability` to make it required, or (2) add a guard: `if capability_obj.payload is None: raise MisconfigurationError(f"Capability '{capability}' has no test payload defined.")`

packages/nemo-evaluator/src/nemo_evaluator/core/input.py
```diff
@@ -592,3 +593,42 @@ def validate_configuration(run_config: dict) -> Evaluation:
...
+    model_id = evaluation.target.api_endpoint.model_id
+    headers = {
+        "Content-Type": "application/json",
+        "Accept": "application/json",
+    }
+    if evaluation.target.api_endpoint.api_key_name is not None:
+        headers["Authorization"] = (
+            f"Bearer {os.getenv(evaluation.target.api_endpoint.api_key_name)}"
+        )
+    for capability in evaluation.config.required_capabilities:
```
pribalta: [Power Review] ⚠️ **HTTP connections opened in loop without session or pool management** · 80% confidence

Per the knowledge document's 'Connection or Pool Exhaustion Risk' rule: 'Resources are acquired inside a loop or request handler without a pool or limit, risking exhaustion under load.' The `for` loop on line 622 calls `requests.post()` for each capability. Each call creates a new TCP connection, performs the TLS handshake, and tears it down rather than reusing connections. Using a `requests.Session` would provide connection pooling and reuse across the loop iterations.

💡 **Suggestion:** Use a `requests.Session` as a context manager to pool connections across capability checks:
```python
with requests.Session() as session:
    session.headers.update(headers)
    for capability in evaluation.config.required_capabilities:
        payload = BENCHMARK_CAPABILITIES[capability].payload.copy()
        payload['model_id'] = model_id
        response = session.post(url, json=payload, timeout=30)
```
This ensures proper connection pooling and deterministic cleanup when the `with` block exits.

packages/nemo-evaluator/src/nemo_evaluator/core/input.py
pribalta: [Power Review] ⚠️ **Shallow copy of nested payload dict risks mutating module-level constant** · 80% confidence

`.copy()` produces a shallow copy. The capability payloads (e.g., TOOLS, IMAGE_URL, AUDIO_URL) contain deeply nested structures (lists of dicts). While the current code only adds a top-level key `model_id`, any future modification to nested values through this copy would silently mutate the shared `BENCHMARK_CAPABILITIES` module-level constant. This is the 'Mutation of Shared Data Structure' pattern from the knowledge document.

💡 **Suggestion:** Use `copy.deepcopy(BENCHMARK_CAPABILITIES[capability].payload)` instead of `.copy()` to fully isolate the payload from the module-level constant.

packages/nemo-evaluator/src/nemo_evaluator/api/api_dataclasses.py
```diff
@@ -202,8 +202,28 @@ class EvaluationConfig(BaseModel):
...
 
+    @model_validator(mode="before")
+    @classmethod
+    def handle_benchmark_capabilities(cls, values):
+        """Handle benchmark capabilities."""
+        if isinstance(values, dict):
+            benchmark_capabilities = values.get("required_capabilities")
+            if benchmark_capabilities:
+                invalid_capabilities = []
+                for capability in benchmark_capabilities:
```
pribalta: [Power Review] ⚠️ **Before-validator iterates required_capabilities without type check — string input yields confusing error** · 80% confidence

CWE-20 (Missing Input Validation): The `model_validator(mode='before')` runs before Pydantic field type validation. If `required_capabilities` is a string (e.g., `'logprobs'` instead of `['logprobs']`), iterating over it yields individual characters (`'l'`, `'o'`, `'g'`, ...), each failing the lookup. The raised ValueError would list characters as invalid capabilities, which is confusing and unhelpful. The knowledge document requires validating type and shape at the entry point.

💡 **Suggestion:** Add a type guard before iterating: `if not isinstance(benchmark_capabilities, list): return values` (let Pydantic's field validator produce the proper type error).

packages/nemo-evaluator/src/nemo_evaluator/core/input.py
```diff
@@ -592,3 +593,42 @@ def validate_configuration(run_config: dict) -> Evaluation:
...
+        payload = BENCHMARK_CAPABILITIES[capability].payload.copy()
+        payload["model_id"] = model_id
```
pribalta: [Power Review] ⚠️ **Payload field `model_id` likely should be `model` for API compatibility** · 72% confidence

OpenAI-compatible chat/completions APIs universally use the field name `model` (not `model_id`) in request payloads. The internal `ApiEndpoint` dataclass uses `model_id` as an attribute name, but the outgoing HTTP request payload should use the API-expected field name. Using `model_id` means the endpoint may ignore the model specification entirely, leading to silent data loss or misconfigured requests. This falls under 'Wrong Variable or Field Referenced' in the knowledge document.

💡 **Suggestion:** Verify the target API's expected field name. If it follows OpenAI conventions, change to `payload["model"] = model_id`.

packages/nemo-evaluator/src/nemo_evaluator/core/input.py
```diff
@@ -592,3 +593,42 @@ def validate_configuration(run_config: dict) -> Evaluation:
...
+
+        response = requests.post(url, json=payload, headers=headers)
+        if response.status_code != 200:
+            raise MisconfigurationError(
+                f"Failed to verify capability {capability}. The endpoint returned {response.status_code} error code and response: {response.text}."
```
pribalta: [Power Review] ⚠️ **Full endpoint response body included in error message — information exposure risk** · 70% confidence

The error message includes the raw `response.text` from an external endpoint. Per the knowledge document's 'Information Exposure via Error Messages' (CWE-200) rule, error messages should not leak potentially sensitive information. The endpoint response could contain internal server details, stack traces, or configuration data that should not be surfaced to end users.

💡 **Suggestion:** Truncate the response text (e.g., `response.text[:500]`) and/or log the full response at DEBUG level while keeping the user-facing error message generic.

packages/nemo-evaluator/src/nemo_evaluator/core/input.py
pribalta: [Power Review] ⚠️ **Payload uses `model_id` but OpenAI-compatible APIs expect `model`** · 70% confidence

The `verify_capabilities` function injects `payload["model_id"] = model_id` into the request payload. However, the capability payloads in `capabilities.py` use OpenAI-compatible request formats (e.g., `messages`, `max_tokens`, `logprobs`, `echo`), where the standard field name is `model`, not `model_id`. Sending `model_id` to an OpenAI-compatible endpoint would likely be ignored, causing the request to fail or use a default model. Per the knowledge document's 'Wrong Argument Order or Type' and 'Missing Required Parameter or Header' rules, this sends an incorrectly named parameter.

💡 **Suggestion:** Verify whether the target API expects `model` or `model_id`. If targeting OpenAI-compatible endpoints, change to `payload["model"] = model_id`.

packages/nemo-evaluator/src/nemo_evaluator/api/api_dataclasses.py
```diff
@@ -202,8 +202,28 @@ class EvaluationConfig(BaseModel):
     supported_endpoint_types: Optional[list[str]] = Field(
         description="Supported endpoint types like chat or completions", default=None
     )
+    required_capabilities: Optional[List[str]] = Field(
```
pribalta: [Power Review] 💡 **Inconsistent generic syntax: `List[str]` vs `list[str]` three lines above** · 75% confidence

Line 202 uses `list[str]` (PEP 585 lowercase) for `supported_endpoint_types` while line 205 uses `List[str]` (typing module) for `required_capabilities`. While both work at runtime in modern Python, this inconsistent representation can confuse contributors about the project's conventions and cause type-checker behavior differences. This is the 'Inconsistent Normalization, Units, or Representation' pattern from the knowledge document applied to type annotations.

💡 **Suggestion:** Standardize on one style — preferably `list[str]` (PEP 585) since it's already used on the adjacent field.

packages/nemo-evaluator/src/nemo_evaluator/core/input.py
```diff
@@ -592,3 +593,42 @@ def validate_configuration(run_config: dict) -> Evaluation:
     check_type_compatibility(evaluation)
     _logger.info(f"User-invoked config: \n{yaml.dump(evaluation.model_dump())}")
     return evaluation
+
+
+def verify_capabilities(evaluation: Evaluation):
```
pribalta: [Power Review] 💡 **Inconsistent return contract: returns True or raises, but caller ignores return value** · 70% confidence

The `verify_capabilities` function returns `True` on success (lines 603, 634) and raises `MisconfigurationError` on failure. However, the caller in `evaluate.py` line 359 calls it as a statement (`verify_capabilities(evaluation)`) without capturing the return value. Per the knowledge document's 'Inconsistent Error Contract' rule, the mixed return-value-and-exception pattern is misleading — the function's contract is ambiguous about whether callers should check the return value or rely solely on exceptions.

💡 **Suggestion:** Either make the function return `None` (void) and rely purely on exceptions for error signaling, or document the return type. Adding a return type annotation (`-> None` or `-> bool`) would clarify the contract. The project enforces MyPy strict mode which requires return type annotations.

packages/nemo-evaluator/src/nemo_evaluator/core/input.py
```diff
@@ -592,3 +594,54 @@ def validate_configuration(run_config: dict) -> Evaluation:
...
+    for capability in evaluation.config.required_capabilities:
+        payload = copy.deepcopy(BENCHMARK_CAPABILITIES[capability].payload)
+        payload["model"] = model_id
+        response = requests.post(url, json=payload, timeout=30, headers=headers)
+        if response.status_code != 200:
+            should_raise = True
+            _logger.error(
+                f'Endpoint {url} responded with error status code when checking for capability "{capability}"',
+                response_status_code=response.status_code,
+                response_text=response.text,
```
AWarno, line 641: I wonder whether this should be an error or a warning (I am not sure how it is handled later). I’m not sure all of our examples are always valid. For example, a video might be provided in a format other than MP4, or there could be constraints on video/audio length or in the past we had issues with out dated vs new tool call payload or max_tokens is now becaming deprecated and max_completions_tokens will be used + timeout 30s may be not enough . These are just some edge cases I can think of right now but maybe some more realistic scenarios exist

packages/nemo-evaluator/src/nemo_evaluator/client/client.py
```diff
@@ -75,15 +75,10 @@ def __init__(
...
-                raise ValueError(
-                    f"API key environment variable '{self.api_key_name}' is not set. "
-                    f"Please set it or remove 'api_key_name' from the configuration."
-                )
-        else:
+        api_key_value = get_api_key_from_env(self.api_key_name)
+        if api_key_value is None:
             # No API key specified - use dummy for endpoints that don't require auth
+            # as AsyncOpenAI requires this parameter to be set
             api_key_value = "dummy_api_key"
```
wprazuch, line 83: I know it was not scoped in this PR, but maybe a `logger.debug` would be good there. Debug logs are pretty verbose already (I believe there could be some optimization done, i.e. for caching debug logs), but this might end up being nice to log in some cases

packages/nemo-evaluator/src/nemo_evaluator/core/input.py
wprazuch: Yeah, agree with the above. Or at least a path to force the evaluation regardless of the above status_code

AWarno: yes maybe some flag to skip

marta-sd: I will change this to warning - this will be closer to the current behavior (no verification at all) and would probably save our users some unpleasant surprises. In the future we can consider adding more strict validation

packages/nemo-evaluator/src/nemo_evaluator/client/client.py
marta-sd: Done in 53513ec890158e49cdfd8a068989157a8f50fb9f

packages/nemo-evaluator/src/nemo_evaluator/core/input.py
marta-sd: Done in e075ea8bccd01f58fd3ffb0e3da5cf2236e06f46

PR #799, opened 2026-03-04
feat: add watch mode for continuous checkpoint evaluation

gchlebus:
## Summary

Adds a new `nel-watch` CLI command (separate entrypoint, following Unix philosophy) that monitors a checkpoint directory and automatically triggers evaluation runs when new checkpoints appear. This enables "eval-and-sleep" workflows where evaluations run continuously as training produces new checkpoints.

`nel-watch` is installed as its own command — it does one thing (watch + dispatch) and composes with `nel run` under the hood.

## Motivation

Researchers doing posttraining want to automatically evaluate checkpoints as they are saved during training — similar to the `nseval watch` functionality in ns-eval. This eliminates manual intervention and ensures every checkpoint gets evaluated.

## Design

- **Separate entrypoint** (`nel-watch`) rather than a `nel watch` subcommand — Unix philosophy: do one thing well
- **CLI is a thin 1:1 wrapper** over the Python API (`api/watch.py`) — every flag maps directly to an API parameter
- **Composes with `nel run`** — watch detects checkpoints, the existing eval pipeline evaluates them
- Users manage their own tmux/screen sessions — not our job to wrap process management

### Core watch loop
- Polls checkpoint directory at configurable interval (default: 300s)
- Detects new checkpoint subdirectories matching configurable glob patterns (default: `iter_*,step_*`)
- Checks readiness via marker files (default: `metadata.json,config.yaml` — any match = ready)
- Overrides a configurable config field (default: `deployment.hf_model_handle`) with the checkpoint path
- Submits eval using the standard `run_eval()` pipeline

### Processing order
- `--order newest` (default): Process highest-step checkpoints first
- `--order oldest`: Process lowest-step checkpoints first
- Natural sort ensures `step_2 < step_10`

### State persistence
- Tracks submitted checkpoints in `.watch_state.yaml`
- Survives restarts — already-submitted checkpoints are skipped
- Custom state file path via `--state-file`

### Multi-directory watching
- `--watch-config watch.yaml` for watching multiple checkpoint directories simultaneously
- Per-directory `output_dir` and `checkpoint_field` overrides
- YAML format:
  ```yaml
  watch_dirs:
    - checkpoint_dir: /lustre/.../run-42/checkpoints
      output_dir: /lustre/.../run-42/evals
    - checkpoint_dir: /lustre/.../run-43/checkpoints
      output_dir: /lustre/.../run-43/evals
  checkpoint_field: deployment.hf_model_handle
  ```

### Additional flags
- `--dry-run`: Show what would be submitted without actually submitting
- `--once`: Scan once and exit (no looping)
- `-o` overrides: Same as `nel run` command

## CLI Usage

```bash
# Basic: watch a single directory
nel-watch --config eval_config.yaml \
  --watch-dir /lustre/checkpoints/training-run-42/ \
  --interval 300

# Dry run, scan once
nel-watch --config eval_config.yaml \
  --watch-dir /lustre/checkpoints/ \
  --dry-run --once

# Multi-directory via config file
nel-watch --config eval_config.yaml \
  --watch-config watch_dirs.yaml

# Run in tmux (user manages the session)
tmux new -d -s eval-watcher 'nel-watch --config eval_config.yaml --watch-dir /lustre/checkpoints/'
```

## Files changed

| File | Lines | Description |
|------|-------|-------------|
| `pyproject.toml` | +1 | Add `nel-watch` script entrypoint |
| `api/watch.py` | +344 | Core watch logic: checkpoint discovery, state management, submission loop |
| `cli/watch.py` | +130 | Thin CLI wrapper with `main()` entrypoint |
| `cli/main.py` | 0 | Unchanged — watch is not a nel subcommand |
| `tests/unit_tests/test_watch.py` | +700 | 38 unit tests covering all features |

## Testing

- **38 unit tests** covering: checkpoint discovery, pattern matching, multi-marker readiness, ordering, state persistence, watch config loading, multi-dir mode, dry-run, once mode
- All tests pass: `uv run --group dev pytest tests/unit_tests/test_watch.py -v`


pribalta (commented):
## Power Review

> Reviewed 4 files, left 30 suggestions: 4 errors, 22 warnings, 4 suggestions.

### Dimensions Analyzed

| Dimension | 🔴 | ⚠️ | 💡 |
| ai-code-hygiene/anti_patterns | 0 | 1 | 0 |
| ai-code-hygiene/architecture | 0 | 2 | 1 |
| ai-code-hygiene/patterns | 0 | 2 | 0 |
| code-correctness/api-contracts | 0 | 4 | 1 |
| code-correctness/boundary-conditions | 1 | 3 | 1 |
| code-correctness/concurrency | 0 | 1 | 0 |
| code-correctness/data-integrity | 0 | 1 | 1 |
| code-correctness/error-handling | 0 | 3 | 0 |
| code-correctness/logic-correctness | 0 | 1 | 0 |
| ultra-eval-infra/advanced-use-cases | 1 | 2 | 0 |
| ultra-eval-infra/code-quality | 2 | 2 | 0 |
| **Total** | **4** | **22** | **4** |

---
> *The diff brings change to add watch mode for ground,*
> *With 30 things our review has found.*
> *Machines can scan but miss the subtle thread,*
> *A human touch ensures the code is read.*

⏱️ Reviewed in **1.5 min** (est. manual review: ~59 min, **40x faster**)

---

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
```diff
@@ -0,0 +1,344 @@
...
+    OmegaConf.update(obj, keys[-1], value)
+
+
+def _load_watch_config(path: Path) -> tuple[list[WatchDirConfig], str]:
+    """Load a watch config YAML file.
+
+    Returns:
+        Tuple of (watch_dirs, global_checkpoint_field).
+    """
+    data = yaml.safe_load(path.read_text())
```
pribalta: [Power Review] 🔴 **Config Validation Missing: _load_watch_config does not validate YAML structure** · 92% confidence

Per the 'Config Validation Missing' ERROR rule, invalid configs should fail fast with a clear error message instead of producing a deep traceback. If a watch config entry is missing the required `checkpoint_dir` key, line 175 will raise a raw `KeyError` with no actionable context. Similarly, if `watch_dirs` is not a list or entries are not dicts, opaque TypeErrors will result.

💡 **Suggestion:** Validate the loaded YAML structure before accessing keys. Check that `watch_dirs` exists and is a list, and that each entry contains the required `checkpoint_dir` key. Raise a clear ValueError, e.g.: `raise ValueError(f"Watch config entry {i} missing required 'checkpoint_dir' key. See docs for format.")`

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
```diff
@@ -0,0 +1,344 @@
...
+            WatchDirConfig(
+                checkpoint_dir=Path(entry["checkpoint_dir"]),
+                output_dir=Path(entry["output_dir"]) if entry.get("output_dir") else None,
+                checkpoint_field=entry.get("checkpoint_field"),
+            )
+        )
+    return watch_dirs, global_field
+
+
+def watch_checkpoints(
```
pribalta: [Power Review] 🔴 **Config Validation Missing: `order` parameter not validated in API layer** · 88% confidence

Per the 'Config Validation Missing' ERROR rule, invalid configs should fail fast. The `order` parameter is only validated in the CLI layer (cli/watch.py line 131), but the public API function `watch_checkpoints` accepts any string. A caller passing an invalid order (e.g., `order='random'`) would silently process checkpoints in natural sort order with no error or warning, since the `if order == 'newest'` check at line 267 simply doesn't match.

💡 **Suggestion:** Add validation at the top of `watch_checkpoints`: `if order not in ('newest', 'oldest'): raise ValueError(f"order must be 'newest' or 'oldest', got '{order}'")`

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
```diff
@@ -0,0 +1,344 @@
...
+                    invocation_id = None
+                    if dry_run:
+                        logger.info(
+                            f"[dry-run] Would submit eval for checkpoint: {cp.name}",
+                            checkpoint_field=field_override,
+                            checkpoint_path=str(cp),
+                        )
+                    else:
+                        try:
+                            invocation_id = run_eval(cfg_copy)
```
pribalta: [Power Review] 🔴 **No timeout on run_eval invocation — watch loop can block indefinitely** · 85% confidence

Per REQ-ADV 'Missing Sandbox Timeout': `run_eval(cfg_copy)` is called without any timeout mechanism. If the evaluation hangs (e.g., sandbox stalls, GPU deadlock, network partition), the entire watch loop blocks, preventing subsequent checkpoints from being processed. There is no kill mechanism and no timeout failure marking.

💡 **Suggestion:** Wrap the `run_eval` call with a configurable timeout (e.g., `--eval-timeout` flag). On expiry, log the checkpoint as a timeout failure and continue the watch loop. Consider using `concurrent.futures` with a timeout or a signal-based alarm.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
```diff
@@ -0,0 +1,344 @@
...
+                continue
+        # Check if any ready marker exists
+        if not any((child / marker).exists() for marker in ready_markers):
+            continue
+        checkpoints.append(child)
+    checkpoints.sort(key=lambda p: _natural_sort_key(p.name))
+    return checkpoints
+
+
+def _set_nested_value(cfg: "OmegaConf", dotted_key: str, value: str) -> None:
```
pribalta: [Power Review] 🔴 **Unhandled empty or missing nested key in _set_nested_value causes AttributeError** · 80% confidence

Per the knowledge document's 'Unhandled Empty or Null Input' rule, `_set_nested_value` dereferences intermediate keys via `getattr(obj, key)` without any guard. If the dotted_key path refers to a non-existent intermediate key (e.g., user passes `checkpoint_field='foo.bar.baz'` but `foo` doesn't exist in the config), this raises an uncaught `AttributeError`. The `checkpoint_field` value comes from user input (CLI flag or watch config YAML) and is not validated against the config structure before use.

💡 **Suggestion:** Wrap the traversal in a try/except and raise a clear `ValueError` indicating which part of the dotted key path is missing, or validate the path before traversal.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
```diff
@@ -0,0 +1,344 @@
...
+        path.parent.mkdir(parents=True, exist_ok=True)
+        path.write_text(yaml.dump(data, default_flow_style=False, sort_keys=False))
+
+    @classmethod
+    def load(cls, path: Path) -> "WatchState":
+        if not path.exists():
+            return cls()
+        try:
+            data = yaml.safe_load(path.read_text()) or {}
+        except Exception:
```
pribalta: [Power Review] ⚠️ **Error message without root cause context** · 90% confidence

The knowledge document requires error messages to include the failing value and error details. The warning log here discards the caught exception entirely — there is no `error=str(e)` or `exc_info=True`. If a user sees 'Failed to load watch state from X, starting fresh', they have no information about *why* (corrupt YAML, permission denied, encoding issue, etc.).

💡 **Suggestion:** Capture the exception as `e` and include it in the log: `logger.warning(f"Failed to load watch state from {path}, starting fresh", error=str(e))`.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
pribalta: [Power Review] ⚠️ **Unhandled None/error from yaml.safe_load in _load_watch_config** · 90% confidence

If the watch config YAML file is empty or contains only comments, `yaml.safe_load()` returns `None`, and `data.get(...)` on line 170 would raise `AttributeError: 'NoneType' object has no attribute 'get'`. Unlike `WatchState.load()` which handles this with `or {}`, `_load_watch_config` does not. This is a null dereference on a fallible return (CWE-476).

💡 **Suggestion:** Add a None guard: `data = yaml.safe_load(path.read_text()) or {}`, consistent with the pattern used in `WatchState.load()`.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
pribalta: [Power Review] ⚠️ **Broad except swallows exception details** · 90% confidence

Per the knowledge doc: 'Never swallow exceptions without logging or re-raising.' While this does log a warning, the `except Exception` is overly broad (catches programming errors like KeyError/AttributeError) and the log message discards the exception details entirely, making debugging impossible.

💡 **Suggestion:** Catch a specific exception (e.g., `yaml.YAMLError, OSError`) and include the exception in the log: `logger.warning(f"Failed to load watch state from {path}, starting fresh", error=str(e))`

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
```diff
@@ -0,0 +1,344 @@
...
+    checkpoints = []
+    for child in watch_dir.iterdir():
+        if not child.is_dir():
+            continue
+        # Check if directory name matches any checkpoint pattern
+        if checkpoint_patterns is not None:
+            if not any(fnmatch.fnmatch(child.name, pat) for pat in checkpoint_patterns):
+                continue
+        # Check if any ready marker exists
+        if not any((child / marker).exists() for marker in ready_markers):
```
pribalta: [Power Review] ⚠️ **Checkpoint detection relies on file existence without stability verification** · 90% confidence

Per REQ-ADV 'Checkpoint Detection Fragile': the ready-marker check uses a bare `(child / marker).exists()` without verifying that the checkpoint directory has been fully written. On shared/network filesystems (e.g., Lustre, as shown in the PR examples), a marker file can appear before all checkpoint data (model weights, optimizer state) has been flushed. This can trigger evaluation on an incomplete checkpoint.

💡 **Suggestion:** Add a stability check — e.g., wait for the marker file's mtime to be stable for a configurable grace period (e.g., `--stable-seconds 30`), or verify a minimum expected file count/size within the checkpoint directory before declaring it ready.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
pribalta: [Power Review] ⚠️ **Monolithic function: `watch_checkpoints` exceeds 50 lines** · 85% confidence

The `watch_checkpoints` function spans lines 183–344 (~90 lines of logic after docstring). Per the knowledge document, functions exceeding ~50 lines are harder to test, review, and maintain. This function handles argument validation, state file resolution, signal handling, the polling loop, checkpoint discovery, config copying, eval submission, state persistence, and sleep logic — all in one flat block.

💡 **Suggestion:** Extract inner concerns into well-named helpers. For example: `_resolve_state_file(dirs_to_watch, config)` for lines 228–240, `_process_new_checkpoints(dir_cfg, config, state, ...)` for lines 283–330, and `_interruptible_sleep(interval, stop_requested)` for lines 336–339.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
```diff
@@ -0,0 +1,344 @@
...
+
+    state = WatchState.load(state_file)
+    session_submissions: list[SubmittedCheckpoint] = []
+
+    stop_requested = False
+
+    def _handle_sigint(signum, frame):
+        nonlocal stop_requested
+        stop_requested = True
+        logger.info("Received SIGINT, finishing current cycle...")
```
pribalta: [Power Review] ⚠️ **Signal handler performs non-signal-safe I/O (logging)** · 85% confidence

Per the knowledge document (Signal / Interrupt Handler Doing Non-Signal-Safe Work): 'A signal handler (SIGTERM, SIGINT) performs I/O, acquires locks, or allocates memory — all of which are unsafe in a signal context.' The `logger.info()` call inside `_handle_sigint` performs I/O and may internally acquire locks (e.g., on stream handlers). In Python, signal handlers run asynchronously in the main thread between bytecode instructions, so calling logging from a signal handler can deadlock if the signal arrives while the main thread already holds a logging lock.

💡 **Suggestion:** Set the flag only in the signal handler; move the log message to the main loop where `stop_requested` is checked. For example:
```python
def _handle_sigint(signum, frame):
    nonlocal stop_requested
    stop_requested = True
```
Then after `if stop_requested: break` in the main loop, add `logger.info("Received SIGINT, finishing current cycle...")`.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
pribalta: [Power Review] ⚠️ **_load_watch_config does not handle None or non-dict YAML content** · 85% confidence

Per 'Missing Input Validation (CWE-20)': `yaml.safe_load` can return `None` (empty file) or a non-dict type (e.g., a plain string or list). `data.get(...)` would then raise `AttributeError` on `None` or a non-dict object. The `WatchState.load` method handles this with `or {}` but `_load_watch_config` does not.

💡 **Suggestion:** Add `data = yaml.safe_load(path.read_text()) or {}` and/or validate that `data` is a dict before accessing keys.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
```diff
@@ -0,0 +1,344 @@
...
+            )
+        )
+    return watch_dirs, global_field
+
+
+def watch_checkpoints(
+    config: RunConfig,
+    watch_dir: Optional[Path] = None,
+    watch_dirs: Optional[list[WatchDirConfig]] = None,
+    interval: int = DEFAULT_INTERVAL,
```
pribalta: [Power Review] ⚠️ **No validation of 'order' parameter in watch_checkpoints API** · 85% confidence

Per 'Missing Input Validation (CWE-20)': the `order` parameter accepts any string but only `'newest'` is explicitly handled (line 267). Any other value (including typos like `'newst'`) silently falls through to the default natural sort order (oldest-first), which may not be the user's intent. The CLI validates this, but the public API function does not.

💡 **Suggestion:** Add validation at the top of `watch_checkpoints`: `if order not in ('newest', 'oldest'): raise ValueError(...)`.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
```diff
@@ -0,0 +1,344 @@
...
+                            checkpoint_path=str(cp),
+                        )
+                    else:
+                        try:
+                            invocation_id = run_eval(cfg_copy)
+                            logger.info(
+                                f"Submitted eval for checkpoint: {cp.name}",
+                                invocation_id=invocation_id,
+                            )
+                        except Exception as e:
```
pribalta: [Power Review] ⚠️ **Failed checkpoints are retried indefinitely with no backoff or failure tracking** · 85% confidence

Per REQ-ADV 'Failure Recovery Broken': when `run_eval` raises an exception, the checkpoint is not recorded in state (line 318: `continue`). On the next poll cycle, the same checkpoint is rediscovered and retried. A persistently failing checkpoint (e.g., corrupted weights) will be retried every cycle indefinitely, flooding logs and wasting resources. Failed runs are logged but never tracked in the state file.

💡 **Suggestion:** Record failed checkpoints in the state file with a failure count and/or status field. Implement a max-retry limit (e.g., `--max-retries 3`) after which the checkpoint is marked as permanently failed and skipped. This preserves the 'continue watching' behavior while preventing infinite retry loops.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
pribalta: [Power Review] ⚠️ **_set_nested_value fails with AttributeError on missing intermediate keys** · 85% confidence

The `_set_nested_value` function uses `getattr(obj, key)` to traverse intermediate keys (line 159). If any intermediate key does not exist in the OmegaConf config, this raises an `AttributeError` (or `ConfigAttributeError` from OmegaConf). There is no validation that the `dotted_key` path is valid within the config before traversal. Since `checkpoint_field` is user-supplied (via CLI or watch config), invalid paths will produce unhelpful errors. This is a violated precondition per the knowledge doc.

💡 **Suggestion:** Add validation or use `OmegaConf.update(cfg, dotted_key, value, merge=False)` directly, which handles nested key paths natively and provides better error messages. This would also eliminate the need for the helper function entirely.

pribalta: [Power Review] ⚠️ **Config Validation Missing: _set_nested_value produces opaque errors on invalid key paths** · 82% confidence

Per the 'Config Validation Missing' ERROR rule, invalid configs should produce clear errors. If `checkpoint_field` contains a dotted path that doesn't exist in the config (e.g., a typo like `deployment.hf_modle_handle`), `getattr(obj, key)` at line 159 will raise an opaque `AttributeError` or `omegaconf.errors.ConfigAttributeError` with no indication that the user-supplied `--checkpoint-field` value is the problem.

💡 **Suggestion:** Wrap the key traversal in a try/except and raise a clear error: `raise ValueError(f"Config field '{dotted_key}' does not exist. Check --checkpoint-field value.")`

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
```diff
@@ -0,0 +1,344 @@
...
+    @classmethod
+    def load(cls, path: Path) -> "WatchState":
+        if not path.exists():
+            return cls()
+        try:
+            data = yaml.safe_load(path.read_text()) or {}
+        except Exception:
+            logger.warning(f"Failed to load watch state from {path}, starting fresh")
+            return cls()
+        submitted = []
```
pribalta: [Power Review] ⚠️ **Unhandled KeyError when loading corrupted state file entries** · 82% confidence

The try/except on lines 89-92 only guards `yaml.safe_load()`. The subsequent loop (lines 94-103) accesses `entry["checkpoint"]` without validation. If the state file is valid YAML but a submitted entry is missing the `checkpoint` key (e.g., manual edit, partial write on crash), this raises an unhandled `KeyError` that crashes the watch loop. Per the knowledge doc, data arriving from a persistence boundary (file read) should be validated at the entry point.

💡 **Suggestion:** Wrap the deserialization loop (lines 94-103) inside the existing try/except, or add per-entry validation with `entry.get("checkpoint")` and skip malformed entries with a warning.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/cli/watch.py
```diff
@@ -0,0 +1,220 @@
...
+        default_factory=list,
+        action="append",
+        nargs="?",
+        alias=["-o"],
+        metadata={
+            "help": "Hydra override in the form some.param.path=value (pass multiple -o for multiple overrides).",
+        },
+    )
+
+    def execute(self) -> None:
```
pribalta, line 104: [Power Review] ⚠️ **Monolithic function: `Cmd.execute` exceeds 50 lines** · 80% confidence

The `execute` method spans lines 112–213 (~100 lines). It handles validation, env loading, config parsing, comma splitting, watch-config loading, logging, API dispatch, and summary printing. Per the knowledge document, functions exceeding ~50 lines should be decomposed.

💡 **Suggestion:** Extract the validation block (lines 123–134) into a `_validate()` method, and the summary printing (lines 204–213) into a `_print_summary()` method, keeping `execute` as thin orchestration.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
pribalta: [Power Review] ⚠️ **Error message not actionable on eval submission failure** · 80% confidence

Per the 'Error Messages Not Actionable' WARNING rule, error messages should include remediation hints. The error log at line 314-317 tells the user what failed but not what to do about it. Common causes include misconfigured API endpoints, missing credentials, or invalid checkpoint paths — none of which are hinted at.

💡 **Suggestion:** Add remediation context to the error log, e.g.: `logger.error(f"Failed to submit eval for checkpoint: {cp.name}. Check that the eval config is valid and the target endpoint is reachable.", error=str(e))`

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/cli/watch.py
```diff
@@ -0,0 +1,220 @@
...
+        watch_dirs = None
+        if self.watch_config:
+            watch_config_path = pathlib.Path(self.watch_config)
+            if not watch_config_path.is_file():
+                raise FileNotFoundError(
+                    f"Watch config file does not exist: {self.watch_config}"
+                )
+            watch_dirs, global_field = _load_watch_config(watch_config_path)
+            # Use global field from watch config if user didn't override on CLI
+            if self.checkpoint_field == "deployment.hf_model_handle":
```
pribalta, line 154: [Power Review] ⚠️ **Cannot distinguish explicit CLI default from unset — checkpoint_field override logic is fragile** · 80% confidence

Line 166 checks `if self.checkpoint_field == "deployment.hf_model_handle"` to determine whether the user explicitly provided --checkpoint-field. If a user explicitly passes `--checkpoint-field deployment.hf_model_handle`, the code incorrectly treats it as 'not overridden' and substitutes the watch config's global_field. This violates the documented CLI contract ('Dot-separated config field to override with checkpoint path'). This is a broken contract between the CLI interface and API behavior.

💡 **Suggestion:** Use `None` as the default for `checkpoint_field` and only fall back to `"deployment.hf_model_handle"` when neither the CLI flag nor the watch config provides a value. For example: `checkpoint_field: str | None = field(default=None, ...)` and then `checkpoint_field = self.checkpoint_field or global_field or "deployment.hf_model_handle"`.

pribalta, line 154: [Power Review] ⚠️ **Default value comparison used to detect CLI override** · 75% confidence

Checking `self.checkpoint_field == "deployment.hf_model_handle"` to determine whether the user explicitly passed `--checkpoint-field` is fragile. If the user explicitly passes the same value as the default, this logic incorrectly treats it as 'not overridden' and substitutes the watch config's global field instead. This is a subtle bug pattern where sentinel detection via default-value comparison fails.

💡 **Suggestion:** Use `None` as the default for `checkpoint_field` and set the actual default (`deployment.hf_model_handle`) only when neither CLI nor watch config provides a value. This correctly distinguishes 'user did not pass the flag' from 'user explicitly passed the default value'.

pribalta, line 154: [Power Review] ⚠️ **Wrong comparison target: default value used as sentinel to detect CLI override** · 75% confidence

The code checks `self.checkpoint_field == "deployment.hf_model_handle"` to decide whether the user explicitly set `--checkpoint-field` on the CLI. If the watch config file defines a different `checkpoint_field` (e.g., `model.path`), and the user explicitly passes `--checkpoint-field deployment.hf_model_handle` to override it, the code incorrectly treats the user's explicit choice as 'not overridden' and substitutes the watch config's `global_field` instead. This is a 'wrong comparison target' issue — comparing against a default value string is not a reliable way to detect explicit user intent.

💡 **Suggestion:** Track whether the flag was explicitly provided (e.g., via a sentinel default like `None`, or by checking `sys.argv` / parser internals) rather than comparing against the default value string. For example, change the dataclass default to `None` and set the actual default (`deployment.hf_model_handle`) only when neither the CLI nor the watch config provides a value.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
```diff
@@ -0,0 +1,344 @@
...
+    checkpoints.sort(key=lambda p: _natural_sort_key(p.name))
+    return checkpoints
+
+
+def _set_nested_value(cfg: "OmegaConf", dotted_key: str, value: str) -> None:
+    """Set a value in an OmegaConf config using a dot-separated key path."""
+    keys = dotted_key.split(".")
+    obj = cfg
+    for key in keys[:-1]:
+        obj = getattr(obj, key)
```
pribalta: [Power Review] ⚠️ **Null dereference on fallible getattr chain in _set_nested_value** · 70% confidence

The `getattr(obj, key)` call on line 159 will raise `AttributeError` if an intermediate key in the dotted path does not exist in the OmegaConf config. This produces a confusing error with no context about which key failed or what dotted_key was being set. The knowledge document flags null/none dereference on fallible returns and error messages without context.

💡 **Suggestion:** Wrap the traversal in a try/except that provides a clear error message, e.g.: `raise ValueError(f"Config path '{dotted_key}' does not exist: key '{key}' not found") from e`.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
```diff
@@ -0,0 +1,344 @@
...
+        try:
+            data = yaml.safe_load(path.read_text()) or {}
+        except Exception:
+            logger.warning(f"Failed to load watch state from {path}, starting fresh")
+            return cls()
+        submitted = []
+        for entry in data.get("submitted", []):
+            submitted.append(
+                SubmittedCheckpoint(
+                    checkpoint=entry["checkpoint"],
```
pribalta: [Power Review] ⚠️ **WatchState.load assumes 'checkpoint' key exists in each entry — KeyError on malformed state file** · 70% confidence

Per 'Unhandled Empty or Null Input': `entry["checkpoint"]` will raise `KeyError` if a state file entry is missing the `checkpoint` key (e.g., due to manual editing or corruption). Other fields use `.get()` with defaults, but `checkpoint` uses direct indexing.

💡 **Suggestion:** Use `entry.get('checkpoint')` with validation, or wrap in try/except and skip malformed entries with a warning, consistent with how corrupt files are already handled.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/cli/watch.py
```diff
@@ -0,0 +1,220 @@
...
+            checkpoint_patterns=checkpoint_patterns,
+            checkpoint_field=checkpoint_field,
+            order=self.order,
+            dry_run=self.dry_run,
+            once=self.once,
+            state_file=state_file_path,
+        )
+
+        # Print summary
+        print(bold(cyan("\n--- Watch Summary ---")))
```
pribalta, line 193: [Power Review] ⚠️ **Print statements instead of project logger** · 70% confidence

Per the knowledge doc: 'AI uses print() instead of the project's logging framework.' The project imports and uses `logger` from `nemo_evaluator_launcher.common.logging_utils` throughout the codebase, but the summary output uses raw `print()` calls.

💡 **Suggestion:** Use `logger.info()` for the summary output, consistent with the rest of the module. For example: `logger.info("Watch Summary", submitted_count=len(submissions))`

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
```diff
@@ -0,0 +1,344 @@
...
+                    logger.debug("No new checkpoints found", watch_dir=str(wd))
+
+                for cp in new_checkpoints:
+                    if stop_requested:
+                        break
+
+                    logger.info(f"Processing checkpoint: {cp.name}", path=str(cp))
+
+                    # Create a copy of the config with the checkpoint path overridden
+                    cfg_copy = OmegaConf.create(
```
pribalta: [Power Review] ⚠️ **Config copy loses RunConfig type — run_eval receives plain DictConfig instead of RunConfig** · 70% confidence

The function signature declares `config: RunConfig`, but at line 290-292, the config is copied via `OmegaConf.create(OmegaConf.to_container(config, resolve=True))`, which produces a plain `DictConfig`, not a `RunConfig`. This plain DictConfig is then passed to `run_eval(cfg_copy)`. If `run_eval` performs type checks, uses structured config validation, or expects `RunConfig`-specific methods, this will fail at runtime. This is a potential violated precondition (Wrong Argument Type per knowledge doc).

💡 **Suggestion:** Use `OmegaConf.structured(RunConfig(**OmegaConf.to_container(config, resolve=True)))` or `copy.deepcopy(config)` to preserve the structured config type. Alternatively, use `OmegaConf.merge(OmegaConf.structured(RunConfig), OmegaConf.to_container(config, resolve=True))` to maintain schema validation.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
```diff
@@ -0,0 +1,344 @@
...
+    session_submissions: list[SubmittedCheckpoint] = []
+
+    stop_requested = False
+
+    def _handle_sigint(signum, frame):
+        nonlocal stop_requested
+        stop_requested = True
+        logger.info("Received SIGINT, finishing current cycle...")
+
+    original_handler = signal.getsignal(signal.SIGINT)
```
pribalta: [Power Review] ⚠️ **signal.signal(SIGINT) raises ValueError if called from non-main thread** · 70% confidence

Line 252-253 calls `signal.signal(signal.SIGINT, _handle_sigint)` which is only valid in the main thread. If `watch_checkpoints` is called from a non-main thread (e.g., in a test runner, async framework, or embedded usage), this raises `ValueError: signal only works in main thread of the main interpreter`. This is a violated precondition — the function has an implicit requirement of running in the main thread that is not documented or checked.

💡 **Suggestion:** Guard the signal registration with a check: `if threading.current_thread() is threading.main_thread():` and skip signal handling otherwise. Document this constraint in the function docstring.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
pribalta: [Power Review] 💡 **Potential duplicated utility: `_set_nested_value` vs `OmegaConf.update`** · 85% confidence

The `_set_nested_value` function manually walks dot-separated keys and then calls `OmegaConf.update` on the leaf container. `OmegaConf.update(cfg, dotted_key, value)` already supports dot-separated key paths natively, making this wrapper unnecessary.

💡 **Suggestion:** Replace `_set_nested_value(cfg, key, val)` with `OmegaConf.update(cfg, key, val)` directly.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/cli/watch.py
```diff
@@ -0,0 +1,220 @@
...
+        metadata={
+            "help": "Hydra override in the form some.param.path=value (pass multiple -o for multiple overrides).",
+        },
+    )
+
+    def execute(self) -> None:
+        from dotenv import load_dotenv
+
+        from nemo_evaluator_launcher.api.functional import RunConfig
+        from nemo_evaluator_launcher.api.watch import (
```
pribalta, line 108: [Power Review] 💡 **CLI imports private function _load_watch_config from API module** · 75% confidence

The CLI layer imports `_load_watch_config` (prefixed with underscore, indicating a private/internal function) from `nemo_evaluator_launcher.api.watch`. This creates tight coupling to an implementation detail rather than a public contract. Per the knowledge doc, callers should depend on the public API, not internals that are subject to change.

💡 **Suggestion:** Either make `_load_watch_config` part of the public API by removing the underscore prefix (rename to `load_watch_config`), or move the watch config loading logic into the CLI layer since it's CLI-specific plumbing.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/cli/watch.py
```diff
@@ -0,0 +1,220 @@
...
+            load_dotenv(env_path, override=False)
+
+        # Load configuration via Hydra
+        config = RunConfig.from_hydra(
+            config=self.config,
+            hydra_overrides=self.override,
+        )
+
+        # Parse comma-separated args
+        ready_markers = [m.strip() for m in self.ready_markers.split(",") if m.strip()]
```
pribalta, line 136: [Power Review] 💡 **Empty ready_markers or checkpoint_patterns after split are not validated** · 70% confidence

Per 'Missing Input Validation': if a user passes `--ready-markers ''` or `--ready-markers ','`, the list comprehension filters empty strings but would produce an empty list. An empty `ready_markers` list means no checkpoint can ever be considered ready, causing the watch to silently do nothing forever.

💡 **Suggestion:** Validate that `ready_markers` and `checkpoint_patterns` are non-empty after parsing; raise a `ValueError` with a clear message if empty.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
pribalta: [Power Review] 💡 **Missing validation on watch config file structure** · 70% confidence

_load_watch_config accesses `entry["checkpoint_dir"]` (line 175) without validation. If the user-provided watch config YAML is valid but missing the required `checkpoint_dir` key in a watch_dirs entry, this raises an unhandled KeyError. Per the knowledge doc, data from external config files (system boundary) should be validated for required fields.

💡 **Suggestion:** Validate that each entry contains `checkpoint_dir` and raise a descriptive `ValueError` (e.g., "Each watch_dirs entry must have a 'checkpoint_dir' field") rather than letting a raw KeyError propagate.

packages/nemo-evaluator-launcher/pyproject.toml
```diff
@@ -51,6 +51,7 @@ all = ["mlflow", "wandb", "gsheets"]
 nemo-evaluator-launcher = "nemo_evaluator_launcher.cli.main:main"
 nv-eval = "nemo_evaluator_launcher.cli.main:main"
 nel = "nemo_evaluator_launcher.cli.main:main"
+nel-watch = "nemo_evaluator_launcher.cli.watch:main"
```
pribalta, line 55: [Power Review] 🔴 **New CLI command ships without user-facing documentation** · 95% confidence

Per the knowledge document: 'Features without docs do not exist from the user's perspective.' The PR adds a new `nel-watch` entrypoint with 11 flags, a watch config YAML schema, state file mechanics, and multi-directory mode, but no README, docs/, or usage guide is updated. The PR description contains rich examples, but that is not discoverable by users after merge.

💡 **Suggestion:** Add a documentation page (e.g., docs/watch-mode.md or a README section) covering: basic usage, watch config YAML schema, state file behavior, and at least one end-to-end example. Ship docs with the feature in the same MR.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
pribalta: [Power Review] 🔴 **Config Validation Missing for watch config YAML** · 92% confidence

Per the knowledge document, 'Invalid configs produce a deep traceback instead of failing fast with a clear error message.' The `_load_watch_config` function does no validation on the loaded YAML. A missing `watch_dirs` key silently returns an empty list. A missing `checkpoint_dir` key in an entry produces a raw `KeyError` traceback instead of a clear, actionable error.

💡 **Suggestion:** Validate the structure after loading: check that `watch_dirs` is a non-empty list, each entry contains a `checkpoint_dir` key, and raise a clear `ValueError` with remediation hints (e.g., 'Each entry in watch_dirs must have a checkpoint_dir key. See docs for watch config format.').

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/cli/watch.py
```diff
@@ -0,0 +1,220 @@
...
+        from nemo_evaluator_launcher.api.functional import RunConfig
+        from nemo_evaluator_launcher.api.watch import (
+            _load_watch_config,
+            watch_checkpoints,
+        )
+        from nemo_evaluator_launcher.common.printing_utils import bold, cyan
+
+        # Validate mutually exclusive args
+        if not self.watch_dir and not self.watch_config:
+            raise ValueError(
```
pribalta: [Power Review] 🔴 **Cryptic ValueError on missing required argument** · 90% confidence

Per the 'Cryptic Error Message' rule, errors must tell the user what went wrong and how to fix it. A bare `ValueError` raised at line 124-125 will surface as a Python traceback with 'ValueError: Either --watch-dir or --watch-config is required.' A junior user seeing a stack trace will not know this is a simple CLI usage issue. The error should be caught and printed cleanly, or the argument parser should enforce mutual exclusivity so the parser itself emits a clean usage message.

💡 **Suggestion:** Use `parser.error()` or `sys.exit('Error: ...')` instead of raising `ValueError`, so the user sees a clean message without a traceback. Alternatively, use argparse mutually exclusive groups to enforce this at the parser level.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/cli/watch.py
```diff
@@ -0,0 +1,220 @@
...
+                "Either --watch-dir or --watch-config is required."
+            )
+        if self.watch_dir and self.watch_config:
```
pribalta, line 117: [Power Review] 🔴 **Cryptic ValueError on mutually exclusive arguments** · 90% confidence

Same as above — `ValueError` at line 128-129 will produce a full Python traceback. A junior engineer will see internal module paths and exception types instead of a clear usage message.

💡 **Suggestion:** Handle this at the parser level or emit a clean error via `sys.exit`.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/cli/watch.py
```diff
@@ -0,0 +1,220 @@
...
+            print("No checkpoints were submitted.")
+        else:
+            print(f"Submitted {len(submissions)} checkpoint evaluation(s):")
+            for s in submissions:
+                inv_str = s.invocation_id or "(dry-run)"
+                print(f"  {s.checkpoint} -> {inv_str}")
+        print(bold(cyan("--- End Summary ---")))
+
+
+def main():
```
pribalta, line 204: [Power Review] 🔴 **Unhandled exceptions in main() will produce raw tracebacks** · 85% confidence

Per 'Cryptic Error Message', the `main()` function at lines 216-220 has no top-level exception handling. Any error from `execute()` (config loading failures, YAML parse errors, missing files) will surface as a full Python traceback. A junior user running `nel-watch` for the first time with a bad config will see internal module paths and stack frames.

💡 **Suggestion:** Wrap `args.cmd.execute()` in a try/except that catches common exceptions (ValueError, FileNotFoundError, yaml.YAMLError) and prints a clean error message with guidance, then exits with a non-zero code.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
pribalta: [Power Review] 🔴 **Watch config YAML parse failure produces cryptic error** · 85% confidence

Per 'Cryptic Error Message', `_load_watch_config` at line 169 calls `yaml.safe_load` and then accesses `data.get('watch_dirs', [])` without any error handling. A malformed watch config YAML will produce a raw `yaml.YAMLError` traceback or a `TypeError` if the file content is not a dict. A junior user will not know what is wrong with their YAML.

💡 **Suggestion:** Wrap the YAML loading and field access in try/except and raise a user-friendly error: 'Failed to parse watch config file {path}. Ensure it is valid YAML with a top-level watch_dirs list. See examples in the documentation.'

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
pribalta: [Power Review] 🔴 **Eval submission failure produces a cryptic error message** · 80% confidence

Per 'Cryptic Error Message', catching a bare `Exception` and logging only `str(e)` will often expose raw internal details (e.g., `KeyError`, `AttributeError`) without telling the user what to fix. A junior user seeing 'Failed to submit eval for checkpoint: step_1, error=KeyError: ...' cannot diagnose the issue.

💡 **Suggestion:** Wrap the error with actionable context: which config field was overridden, what the checkpoint path was, and suggest the user verify their eval config. For example: `f"Failed to submit eval for {cp.name}. Check that your eval config is valid and '{field_override}' is a valid field. Details: {e}"`.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
```diff
@@ -0,0 +1,344 @@
...
+    if checkpoint_patterns is None:
+        checkpoint_patterns = list(DEFAULT_CHECKPOINT_PATTERNS)
+
+    # Build the list of directories to watch
+    if watch_dirs is not None:
+        dirs_to_watch = watch_dirs
+    elif watch_dir is not None:
+        dirs_to_watch = [WatchDirConfig(checkpoint_dir=watch_dir)]
+    else:
+        raise ValueError("Either watch_dir or watch_dirs must be provided.")
```
pribalta: [Power Review] 🔴 **API-level ValueError lacks user-facing guidance** · 70% confidence

Per 'Cryptic Error Message', the error 'Either watch_dir or watch_dirs must be provided' exposes API-level parameter names. A junior user calling through the CLI would see a traceback referencing internal function parameters. This should not be reachable from the CLI (the CLI should catch it), but if it is, the message should reference CLI flags.

💡 **Suggestion:** Ensure the CLI layer catches this before it reaches the API, or include CLI flag names in the message: 'Provide --watch-dir or --watch-config.'

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
pribalta: [Power Review] ⚠️ **Failed checkpoints are retried infinitely — no failure tracking or back-off** · 90% confidence

Per 'Resource Leak at Scale': when `run_eval` raises an exception (line 313), the checkpoint is not recorded in state. On the next poll cycle, the same checkpoint is rediscovered and resubmitted. A permanently failing checkpoint (e.g., corrupted data) causes unbounded retry attempts, wasting GPU hours and log noise at scale.

💡 **Suggestion:** Track failed submissions in `WatchState` with a retry counter. After a configurable `--max-retries` (default: 3), mark the checkpoint as permanently failed and skip it. Log each retry with the attempt number.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/cli/watch.py
```diff
@@ -0,0 +1,220 @@
...
+import pathlib
+from dataclasses import dataclass
+
+from simple_parsing import ArgumentParser, field
+
+from nemo_evaluator_launcher.common.logging_utils import logger
+
+
+@dataclass
+class Cmd:
```
pribalta, line 27: [Power Review] ⚠️ **Config example missing for watch config YAML format** · 90% confidence

Per 'Config Example Missing or Outdated', new config formats must have working examples. The `--watch-config` flag accepts a YAML file with a specific schema (`watch_dirs`, `checkpoint_field`), but no example file is included in the diff. A junior user has no reference file to copy from — only the PR description documents the format.

💡 **Suggestion:** Add an example watch config file (e.g., `examples/watch_config.yaml`) with inline comments explaining each field, valid values, and defaults.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
```diff
@@ -0,0 +1,344 @@
...
+        for d in dirs_to_watch:
+            if d.output_dir is not None:
+                first_output = d.output_dir
+                break
+        if first_output is None:
+            if hasattr(config, "execution") and hasattr(config.execution, "output_dir"):
+                first_output = Path(str(config.execution.output_dir))
+            else:
+                first_output = Path(".")
+        state_file = first_output / ".watch_state.yaml"
```
pribalta: [Power Review] ⚠️ **Auto-generated state file location is unclear to the user** · 90% confidence

Per 'Unclear Output Location or Format', when `--state-file` is not provided, the state file is auto-generated at a path determined by complex fallback logic (first watch_dir output_dir → config output_dir → cwd). The user is never told where the state file was created. This makes debugging restart behavior difficult for a junior user.

💡 **Suggestion:** Log the resolved state file path at `info` level when the watch loop starts, e.g., `logger.info('State file: {state_file}')`.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
```diff
@@ -0,0 +1,344 @@
...
+    def _handle_sigint(signum, frame):
+        nonlocal stop_requested
+        stop_requested = True
+        logger.info("Received SIGINT, finishing current cycle...")
+
+    original_handler = signal.getsignal(signal.SIGINT)
+    signal.signal(signal.SIGINT, _handle_sigint)
+
+    try:
+        while True:
```
pribalta: [Power Review] ⚠️ **No upper bound on evaluations submitted per cycle — cost not controllable** · 85% confidence

Per the 'Cost Not Controllable' rule: unbounded operations with significant compute cost require a way to set limits. If a checkpoint directory accumulates hundreds of checkpoints (e.g., after a long training run or a restart with a cleared state file), all will be submitted in a single cycle with no cap. Each submission triggers a full GPU evaluation run.

💡 **Suggestion:** Add a `--max-evals` / `max_evals` parameter (defaulting to unlimited) that caps the number of evaluations submitted per session or per cycle. Log a warning when the cap is reached so the operator knows checkpoints were deferred.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
```diff
@@ -0,0 +1,344 @@
...
+
+                    invocation_id = None
+                    if dry_run:
+                        logger.info(
+                            f"[dry-run] Would submit eval for checkpoint: {cp.name}",
+                            checkpoint_field=field_override,
+                            checkpoint_path=str(cp),
+                        )
+                    else:
+                        try:
```
pribalta: [Power Review] ⚠️ **Missing timing telemetry for eval submissions** · 85% confidence

Per the 'Missing Operational Observability' rule: new execution paths require structured log events with timing and status. The `run_eval()` call has no duration measurement. At scale, operators need per-submission latency to monitor health, detect degradation, and capacity-plan.

💡 **Suggestion:** Record `time.monotonic()` before and after `run_eval()`, then include `duration_s` in the structured log event at line 309. Also emit a per-cycle summary log with total checkpoints processed, succeeded, failed, and cycle wall-time.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/cli/watch.py
```diff
@@ -0,0 +1,220 @@
...
+        # Validate mutually exclusive args
+        if not self.watch_dir and not self.watch_config:
+            raise ValueError(
+                "Either --watch-dir or --watch-config is required."
+            )
+        if self.watch_dir and self.watch_config:
+            raise ValueError(
+                "--watch-dir and --watch-config are mutually exclusive."
+            )
+        if self.order not in ("newest", "oldest"):
```
pribalta, line 119: [Power Review] ⚠️ **Invalid --order value produces a traceback instead of clean error** · 85% confidence

Per 'Cryptic Error Message', a `ValueError` traceback for an invalid `--order` value is unfriendly. The parser should use `choices=['newest', 'oldest']` so the user gets standard argparse error output with valid options listed.

💡 **Suggestion:** Define `order` with `choices=['newest', 'oldest']` in the argument metadata, or use `parser.error()` to emit a clean message.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
```diff
@@ -0,0 +1,344 @@
...
+                    state.submitted.append(record)
+                    session_submissions.append(record)
+
+                    if not dry_run:
+                        state.save(state_file)
+
+            if once or stop_requested:
+                break
+
+            logger.debug(f"Sleeping {interval}s before next poll")
```
pribalta: [Power Review] ⚠️ **No progress feedback during sleep interval** · 85% confidence

Per 'Missing Progress Feedback', a long-running operation with no output makes a junior user think the tool is stuck. With a default 300-second polling interval, the tool is silent for 5 minutes. The only output is a `logger.debug` message at line 335 which is hidden at default log levels.

💡 **Suggestion:** Log at `info` level instead of `debug`: e.g., `logger.info(f'No new checkpoints. Next poll in {interval}s...')`. This reassures the user the tool is still running.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/cli/watch.py
pribalta, line 27: [Power Review] ⚠️ **Debugging affordance missing: no --verbose/--debug flag** · 85% confidence

Per the knowledge document ('Debugging Affordance Missing'): 'When something goes wrong, there is no way to increase verbosity, inspect intermediate state, or reproduce the issue in isolation.' The `nel-watch` CLI is a separate entrypoint that defines its own `Cmd` dataclass but does not expose any `--verbose` or `--debug` flag. The existing `nel` CLI likely has verbosity controls, but since `nel-watch` is a standalone entrypoint, those controls are not inherited. A mid-level engineer debugging checkpoint discovery or submission failures has no way to increase log verbosity.

💡 **Suggestion:** Add a `--verbose` / `--debug` flag to the `Cmd` dataclass that configures the logger level (e.g., sets `logger` to DEBUG). This is consistent with the `nel` CLI and aids troubleshooting in long-running watch sessions.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
```diff
@@ -0,0 +1,344 @@
...
+            ]
+        }
+        path.parent.mkdir(parents=True, exist_ok=True)
+        path.write_text(yaml.dump(data, default_flow_style=False, sort_keys=False))
+
+    @classmethod
+    def load(cls, path: Path) -> "WatchState":
+        if not path.exists():
+            return cls()
+        try:
```
pribalta: [Power Review] ⚠️ **Warning on corrupt state file omits exception details** · 82% confidence

Per the knowledge document, error messages should be actionable. The warning when state loading fails swallows the exception details entirely via a bare `except Exception`, making it difficult for operators to diagnose why state loading failed (permissions? encoding? disk error?).

💡 **Suggestion:** Include the exception in the warning message: `logger.warning(f"Failed to load watch state from {path}, starting fresh", error=str(e))` by capturing the exception with `except Exception as e`.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
pribalta: [Power Review] ⚠️ **Checkpoint readiness check does not verify write completeness** · 82% confidence

REQ-ADV WARNING 'Checkpoint Detection Fragile': The readiness check only verifies marker file existence (`(child / marker).exists()`). A marker file may be created before the checkpoint data files are fully written (e.g., training framework writes marker first, then large model shards). There is no size/mtime stability verification to guard against partially-written checkpoints.

💡 **Suggestion:** Add an optional stability check: after detecting a marker, wait a short period and verify that the checkpoint directory's total size or newest file mtime has stabilized (e.g., no changes for N seconds). Alternatively, document that users must ensure their training framework writes the marker file last.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
```diff
@@ -0,0 +1,344 @@
...
+    # Build the list of directories to watch
+    if watch_dirs is not None:
+        dirs_to_watch = watch_dirs
+    elif watch_dir is not None:
+        dirs_to_watch = [WatchDirConfig(checkpoint_dir=watch_dir)]
+    else:
+        raise ValueError("Either watch_dir or watch_dirs must be provided.")
+
+    # Determine state file location
+    if state_file is None:
```
pribalta: [Power Review] ⚠️ **Debugging affordance missing: auto-generated state file path not logged** · 82% confidence

Per the knowledge document ('Debugging Affordance Missing'): 'no way to ... inspect intermediate state.' The state file location is auto-generated via a multi-step fallback (first watch_dir output_dir → config output_dir → cwd), but the resolved path is never logged. A mid-level engineer restarting `nel-watch` or debugging duplicate submissions cannot easily determine which state file is in use without reading the source code.

💡 **Suggestion:** Log the resolved `state_file` path at INFO level after the auto-generation logic completes (after line 240). Example: `logger.info('Using state file', state_file=str(state_file))`.

pribalta (commented):
## Power Review

> Reviewed 4 files, left 31 suggestions: 8 errors, 20 warnings, 3 suggestions.

### Dimensions Analyzed

| Dimension | 🔴 | ⚠️ | 💡 |
| nemo-evaluator-ux/junior-engineer-ux | 6 | 5 | 1 |
| nemo-evaluator-ux/mid-level-engineer-ux | 0 | 5 | 0 |
| nemo-evaluator-ux/product-management-ux | 1 | 1 | 1 |
| nemo-evaluator-ux/senior-principal-ux | 0 | 3 | 1 |
| ultra-eval-infra/advanced-use-cases | 0 | 3 | 0 |
| ultra-eval-infra/code-quality | 1 | 3 | 0 |
| **Total** | **8** | **20** | **3** |

---
> *The diff brings change to add watch mode for ground,*
> *With 31 things our review has found.*
> *Machines can scan but miss the subtle thread,*
> *A human touch ensures the code is read.*

⏱️ Reviewed in **1.3 min** (est. manual review: ~59 min, **46x faster**)

---

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
pribalta: [Power Review] ⚠️ **Incomplete error context on submission failure — no resume guidance** · 80% confidence

Per the knowledge document ('Incomplete Error Context in Multi-Step Workflow'): 'the error does not indicate which step failed, which input caused it, or how to resume from the failure point.' When `run_eval` raises an exception, the error is logged with only the checkpoint name and `str(e)`. The user is not told: (1) which internal step failed (config override, deployment, scoring, etc.), (2) that the checkpoint will be retried on next restart (since failed checkpoints are not recorded in state), or (3) how to manually re-run just that checkpoint. In a long-running watch session with many checkpoints, this makes debugging difficult.

💡 **Suggestion:** Enrich the error log to include: the config field being overridden, the resolved config path, and a note that the checkpoint will be retried on next scan cycle. Consider logging the full traceback at DEBUG level. Example: `logger.error(f'Failed to submit eval for {cp.name} (field={field_override}). Will retry on next cycle.', error=str(e))`.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
pribalta: [Power Review] ⚠️ **Silent fallback when `_set_nested_value` fails on invalid dotted key** · 78% confidence

Per the knowledge document, 'Failures proceed silently, hiding problems.' If a user provides an invalid `checkpoint_field` (e.g., a non-existent nested path), `_set_nested_value` will raise an opaque `AttributeError` from `getattr(obj, key)` with no actionable context about which config field was invalid.

💡 **Suggestion:** Wrap with a try/except that raises a clear `ValueError`, e.g., `f"Cannot set config field '{dotted_key}': key '{key}' not found. Verify --checkpoint-field matches your eval config structure."`.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
pribalta: [Power Review] ⚠️ **Failed checkpoints are silently retried on every poll cycle with no bound** · 78% confidence

REQ-ADV ERROR 'Failure Recovery Broken' requires that failed runs are logged and the watch loop continues — both are satisfied. However, since failed checkpoints are not recorded in state (line 318 `continue` skips state append), they will be rediscovered and retried indefinitely on every poll cycle. A permanently failing checkpoint (e.g., corrupted data) will generate unbounded retry attempts and log noise, potentially delaying processing of valid checkpoints.

💡 **Suggestion:** Track failed attempts in `WatchState` with a retry counter. After a configurable max retry count, mark the checkpoint as permanently failed and skip it in future cycles. Log a clear message when max retries are exhausted.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/cli/watch.py
pribalta, line 154: [Power Review] ⚠️ **Sentinel-value comparison makes --checkpoint-field non-overridable when using --watch-config** · 78% confidence

Per the knowledge document ('Config Override Not Supported'): fields should support CLI override. Line 166 checks `if self.checkpoint_field == "deployment.hf_model_handle"` to decide whether the user explicitly set `--checkpoint-field`. This comparison against the default value means if a user explicitly passes `--checkpoint-field deployment.hf_model_handle` (to force-override a watch config's global field), the CLI incorrectly treats it as 'not overridden' and uses the watch config's value instead. This is a subtle composability issue.

💡 **Suggestion:** Track whether `--checkpoint-field` was explicitly provided by the user (e.g., via `argparse`'s default sentinel or a separate boolean) rather than comparing against the default string value.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/cli/watch.py
```diff
@@ -0,0 +1,220 @@
...
+
+        from nemo_evaluator_launcher.api.functional import RunConfig
+        from nemo_evaluator_launcher.api.watch import (
+            _load_watch_config,
+            watch_checkpoints,
+        )
+        from nemo_evaluator_launcher.common.printing_utils import bold, cyan
+
+        # Validate mutually exclusive args
+        if not self.watch_dir and not self.watch_config:
```
pribalta, line 115: [Power Review] ⚠️ **Mutual exclusivity and order validation use raw ValueError instead of argparse error handling** · 75% confidence

Per the Inconsistent Product Surface rule: error reporting style should be consistent across the product. The existing `nel` CLI uses argparse/simple_parsing for argument validation, which produces standard usage messages with exit code 2. Here, `ValueError` is raised manually (lines 123-134), producing a Python traceback instead of a clean CLI error. This is a different error UX than the rest of the tool.

💡 **Suggestion:** Use `parser.error()` for mutually exclusive argument validation (or declare a mutually exclusive group in argparse) so that errors produce the standard usage message consistent with other `nel` commands.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
pribalta: [Power Review] ⚠️ **Error message not actionable on eval submission failure** · 75% confidence

Per the knowledge document, 'When something fails, the error tells the user what went wrong but not what to do about it.' The error log at line 315-317 only prints the exception string without remediation hints. Users seeing 'Failed to submit eval for checkpoint: step_1' with a raw exception have no guidance on how to fix the issue.

💡 **Suggestion:** Include remediation guidance, e.g., 'Check your eval config and API connectivity. The checkpoint will be retried on the next poll cycle. Use --once --dry-run to debug.'

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
```diff
@@ -0,0 +1,344 @@
...
+        path.write_text(yaml.dump(data, default_flow_style=False, sort_keys=False))
+
+    @classmethod
+    def load(cls, path: Path) -> "WatchState":
+        if not path.exists():
+            return cls()
+        try:
+            data = yaml.safe_load(path.read_text()) or {}
+        except Exception:
+            logger.warning(f"Failed to load watch state from {path}, starting fresh")
```
pribalta: [Power Review] ⚠️ **Corrupt state file warning lacks actionable guidance** · 75% confidence

Per 'Cryptic Error Message', the warning at line 92 says 'Failed to load watch state from {path}, starting fresh' but does not tell the user what to do. A junior user may not realize this means previously-submitted checkpoints will be re-evaluated.

💡 **Suggestion:** Expand the message: `f'Failed to load watch state from {path} — starting fresh. Previously submitted checkpoints may be re-evaluated. To fix, delete or repair {path}.'`

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
pribalta: [Power Review] ⚠️ **No results aggregation across evaluated checkpoints** · 70% confidence

REQ-ADV WARNING 'No Results Aggregation': After evaluating N checkpoints, results are not aggregated into a combined view (e.g., training curve, metric progression over steps). The watch loop tracks submission records but provides no mechanism to collect and present per-checkpoint evaluation results together.

💡 **Suggestion:** After completing a watch session (or periodically), aggregate per-checkpoint results into a combined report — e.g., a summary table or CSV mapping checkpoint step to key metrics, enabling training curve visualization.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/cli/watch.py
```diff
@@ -0,0 +1,220 @@
...
+        checkpoint_patterns = [
+            p.strip() for p in self.checkpoint_patterns.split(",") if p.strip()
+        ]
+
+        state_file_path = pathlib.Path(self.state_file) if self.state_file else None
+
+        # Determine watch mode
+        watch_dir = None
+        watch_dirs = None
+        if self.watch_config:
```
pribalta, line 146: [Power Review] ⚠️ **Config override not supported for watch config YAML fields** · 70% confidence

Per the knowledge document ('Config Override Not Supported'): 'A new config field cannot be overridden from the CLI via the `-o field.subfield=value` syntax, forcing users to create separate YAML files for small variations.' The `--watch-config` YAML file defines `watch_dirs[].checkpoint_dir`, `watch_dirs[].output_dir`, and `checkpoint_field`, but none of these can be overridden from the CLI via `-o`. A mid-level engineer wanting to temporarily change a single directory's output path must edit or duplicate the YAML file.

💡 **Suggestion:** Document this limitation explicitly in the `--watch-config` help text. Consider supporting CLI overrides for common watch-config fields (e.g., `--checkpoint-field` already exists as a global override, but per-directory overrides require YAML edits). At minimum, the `-o` flag help should clarify it only applies to the eval config, not the watch config.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/cli/watch.py
```diff
@@ -0,0 +1,220 @@
...
+from simple_parsing import ArgumentParser, field
+
+from nemo_evaluator_launcher.common.logging_utils import logger
+
+
+@dataclass
+class Cmd:
+    """Watch a checkpoint directory and trigger evaluations for new checkpoints."""
+
+    config: str = field(
```
pribalta, line 30: [Power Review] 💡 **CLI help text lacks a minimal getting-started example** · 80% confidence

Per 'Missing Getting Started Path', a junior user should be able to try the feature with a single copy-paste command. The `Cmd` docstring at line 28 is a bare description with no usage example. Adding an epilog or extended help with a minimal command would reduce time-to-first-success.

💡 **Suggestion:** Add an epilog or extended description to the `Cmd` dataclass docstring showing a minimal usage example, e.g.: `Example: nel-watch --config eval.yaml --watch-dir /path/to/checkpoints --once --dry-run`

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/cli/watch.py
pribalta, line 27: [Power Review] 💡 **CLI help text lacks usage examples for progressive disclosure** · 72% confidence

Per the No Progressive Disclosure suggestion: all complexity is exposed upfront. The `Cmd` docstring (line 28) is a single sentence. With 11 flags, a new user seeing `nel-watch --help` gets a wall of options without guidance on the simplest invocation. The existing `nel` CLI provides a rich help epilog with examples.

💡 **Suggestion:** Add an epilog or extended description showing the minimal invocation (`nel-watch --config eval.yaml --watch-dir /path --once`) and one advanced example (multi-dir via `--watch-config`), matching the pattern used by the existing `nel` CLI.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/watch.py
```diff
@@ -0,0 +1,344 @@
...
+@dataclass
+class WatchState:
+    """Persistent state for the watch loop."""
+
+    submitted: list[SubmittedCheckpoint] = field(default_factory=list)
+
+    def submitted_paths(self) -> set[str]:
+        return {s.checkpoint for s in self.submitted}
+
+    def save(self, path: Path) -> None:
```
pribalta: [Power Review] 💡 **State file has no file locking — concurrent watchers risk duplicate submissions** · 70% confidence

Per 'No Multi-Tenancy Consideration': the `.watch_state.yaml` file is read and written without any file lock. Two `nel-watch` processes watching overlapping directories (or the same directory) can race: both read state, both see a checkpoint as new, both submit, then both write — one overwriting the other's record. On shared infrastructure this is a realistic scenario.

💡 **Suggestion:** Use `fcntl.flock` or a `.lock` sidecar file to serialize state reads and writes. Alternatively, document that concurrent watchers on overlapping directories are unsupported.

marta-sd:
Feature added in #857

PR #805, opened 2026-03-05
refactor: clean up multi-node config and use num_instances

AdamRajfer:
Replace boolean multi-instance flag with `num_instances` count in slurm execution config. Set up node IP discovery and per-instance routing for multi-node deployments. Clean up multi-node examples and docs. Use `OmegaConf.select` for safe `num_instances` access across all executor types. Update tests.


marta-sd:
Closing as we've merged a more extended version in #757

PR #807, opened 2026-03-05
feat(multinode-multiinsnatnce): generic multiinstance multinode for slurm

AWarno:
## Changes

- **Generalize multi-instance support**  
  Move instance orchestration (node partitioning, `PROC_ID` / `MASTER_IP` injection, per-instance execution) from `vllm_ray.yaml` into the **executor** (currently SLURM). In the future, other executors can also use the vLLM Ray deployment as long as the `PROC_ID` and `MASTER_IP` environment variables are provided. 
  Multi-instance multi nodes deployments now work with any framework not just vllm ray (e.g., SGLang, TRT-LLM, Megatron-LM, Ray-less vLLM, etc.).

- **Decouple execution and deployment**  
  `vllm_ray.yaml` is now purely a **Ray cluster setup script** with no dependencies on the executor.

- **Improve Ray stability**  
  Add default value: `VLLM_ENGINE_READY_TIMEOUT_S=3600`.

- **Fix DeepSeek-R1 on Ray**  
  Add `--enforce-eager` to `extra_args`. CUDA graph capture fails with **DeepGEMM kernels** in multi-node Ray deployments for this model.

- **Fix failing tests**  
  Tests were failing due to the required presence of `execution.num_instances`.

### DIFF
--- a/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/__init__.py
+++ b/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/__init__.py
@@ -22,14 +22,7 @@ It automatically initializes logging and conditionally loads internal components
 import importlib
 import warnings
 
-from omegaconf import OmegaConf
-
 from nemo_evaluator_launcher.common.logging_utils import logger
-
-# Register integer-division resolver so YAML configs can express derived values.
-# Example: n_tasks: ${div:${execution.num_nodes},${execution.num_instances}}
-if not OmegaConf.has_resolver("div"):
-    OmegaConf.register_new_resolver("div", lambda a, b: int(a) // int(b))
 from nemo_evaluator_launcher.package_info import (
     __contact_emails__,
     __contact_names__,


packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/configs/execution/slurm/default.yaml
```diff
@@ -31,7 +31,7 @@ sbatch_comment: null # Optional comment for SLURM job (translates to #SBATCH --c
 
 # Deployment-specific SLURM configuration
 deployment:
-  n_tasks: ${execution.num_nodes}  # Number of tasks for deployment srun (default: num_nodes)
+  n_tasks: ${div:${execution.num_nodes},${execution.num_instances}}  # number of srun tasks per instance; defaults to nodes_per_instance (1 task per node). Override if your deployment requires more tasks per node (e.g. ntasks_per_node > 1).
```
AWarno: I changed the meaning, before it was for a whole srun (because we had one) now it is per srun deployment , some people using old syntax (some internal users) may have: n_tasks: ${execution.num_nodes} + multiple_instances: true, which by default will result in the same behavior (as default  num_instances is 1) but if they change num_instances (new flag)  + ${execution.num_nodes} + multiple_instances: true (deprectaed flag) it will result in a weird behavior

PR #811, opened 2026-03-06
feat(malformed-configs-validation): validation of malformed configs

AWarno:
Validate malformed NEL configs early

 - Pydantic check for evaluation section — unknown/typo'd keys raise ValueError
  immediately                                                                             
  - Pydantic check for execution.mounts section — unknown/typo'd keys raise ValueError
  immediately
  - Enriched error messages listing the invalid key and all allowed keys                 
  - Warn if nemo_evaluator_config.params contains keys not referenced in the task's      
  command template (checked against packaged IRs — run is not blocked, warning only)  detects typos in e.g temperature

Example configs which will fail now:
Typo in `pre_cmd` command:
```
defaults:
  - execution: local
  - deployment: generic
  - _self_

execution:
  output_dir: /tmp/nel-bad-test

deployment:
  image: python:3.12-slim
  command: echo hello
  port: 8000
  served_model_name: dummy-model

# BAD: 'precmd instead of pre_cmd'
evaluation:
  tasks:
    - name: lm-evaluation-harness.ifeval
      precmd: echo hi

```

output:

```
NFO log_dir=none (NEMO_EVALUATOR_LOG_DIR not set) logger=nemo_evaluator.logging.utils
✗ Job submission failed, see logs | Error: Invalid 'evaluation' config:
1 validation error for EvaluationModel
tasks.0.precmd
  Extra inputs are not permitted [type=extra_forbidden, input_value='echo hi', input_type=str]
    For further information visit https://errors.pydantic.dev/2.12/v/extra_forbidden
Allowed top-level evaluation keys: env_vars, nemo_evaluator_config, pre_cmd, tasks
Allowed task keys: container, dataset_dir, dataset_mount_path, endpoint_type, env_vars, name, nemo_evaluator_config, pre_cmd
[E 2026-03-09T13:53:42.687] Job submission failed error=Invalid 'evaluation' config:
```

Example configs which will raise warning (invalid params in nemo evaluator config):

```
  # bad_nemo_evaluator_params.yaml
  defaults:
    - execution: local
    - deployment: generic
    - _self_

  execution:
    output_dir: /tmp/nel-bad-test

  deployment:
    image: python:3.12-slim
    command: echo hello
    port: 8000
    served_model_name: dummy-model

  evaluation:
    tasks:
      - name: lm-evaluation-harness.ifeval
        container: nvcr.io/nvidia/eval-factory/lm-evaluation-harness:26.01
        nemo_evaluator_config:
          config:
            params:
              parallelism: 4
              bad_param_xyz: 999        # ← does not exist in ifeval command template
              another_fake_param: true  # ← same
```

Output:
```
[W 2026-03-09T13:51:40.743] Configuration contains parameter(s) that are not used in the command template: config.params.bad_param_xyz, config.params.another_fake_param. Valid params from command: ['config.params.limit_samples', 'config.params.max_new_tokens', 'config.params.max_retries', 'config.params.parallelism', 'config.params.request_timeout', 'config.params.task', 'config.params.temperature', 'config.params.top_p', 'config.params.extra.args', 'config.params.extra.downsampling_ratio', 'config.params.extra.num_fewshot', 'config.params.extra.tokenized_requests', 'config.params.extra.tokenizer', 'config.params.extra.tokenizer_backend']. Remove the unused parameters or update the command template to use them.
```

Check if user passed container is used in the validation (passed simple evals image instead of lm-eval) output:

```
[W 2026-03-12T09:44:05.128] Task not found or could not resolve task definition, skipping param validation task=lm-evaluation-harness.ifeval error=Harness 'lm-evaluation-harness' does not match container. Container supports: ['simple_evals']
```

**Note:** Only parameters are checked. For example, an invalid adapters config is not validated, as its correct definition may vary across different images.

### DIFF
--- a/packages/nemo-evaluator-launcher/tests/unit_tests/test_config_validation.py
+++ b/packages/nemo-evaluator-launcher/tests/unit_tests/test_config_validation.py
@@ -22,6 +22,7 @@ param validation (via _validate_config_sections).
 import copy
 import textwrap
 from typing import Any, Dict, Optional
+from unittest.mock import patch
 
 import pytest
 import yaml
@@ -317,19 +318,21 @@ INVALID_PARAM_CONFIGS = [
 
 
 _IFEVAL_CONTAINER = "nvcr.io/nvidia/eval-factory/lm-evaluation-harness:26.01"
+SIMPLE_EVALS_CONTAINER = "nvcr.io/nvidia/eval-factory/simple-evals:26.01"
 
 
 class TestNemoEvaluatorParamValidation:
-    def _make_cfg(self, task_name: str, params: dict):
+    def _make_cfg(self, task_name: str, params: dict, container: Optional[str] = None):
+        """Build config for param validation. Omit container to use packaged IRs (default mapping)."""
+        task = {
+            "name": task_name,
+            "nemo_evaluator_config": {"config": {"params": params}},
+        }
+        if container is not None:
+            task["container"] = container
         return make_config(
             evaluation={
-                "tasks": [
-                    {
-                        "name": task_name,
-                        "container": _IFEVAL_CONTAINER,
-                        "nemo_evaluator_config": {"config": {"params": params}},
-                    }
-                ],
+                "tasks": [task],
                 "nemo_evaluator_config": {},
             }
         )
@@ -358,13 +361,12 @@ class TestNemoEvaluatorParamValidation:
 
     def test_global_valid_task_invalid_emits_warning(self, caplog):
         # Given global nemo_evaluator_config has valid params
-        # but task-level adds an invalid param on top
+        # but task-level adds an invalid param on top (no container -> use packaged IRs)
         cfg = make_config(
             evaluation={
                 "tasks": [
                     {
                         "name": "lm-evaluation-harness.ifeval",
-                        "container": _IFEVAL_CONTAINER,
                         "nemo_evaluator_config": {
                             "config": {"params": {"bad_task_param": 1}}
                         },
@@ -380,13 +382,12 @@ class TestNemoEvaluatorParamValidation:
 
     def test_global_invalid_task_valid_emits_warning(self, caplog):
         # Given global nemo_evaluator_config has an invalid param
-        # but task-level only sets valid params
+        # but task-level only sets valid params (no container -> use packaged IRs)
         cfg = make_config(
             evaluation={
                 "tasks": [
                     {
                         "name": "lm-evaluation-harness.ifeval",
-                        "container": _IFEVAL_CONTAINER,
                         "nemo_evaluator_config": {
                             "config": {"params": {"parallelism": 4}}
                         },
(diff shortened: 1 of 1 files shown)


packages/nemo-evaluator-launcher/tests/unit_tests/test_config_models.py
```diff
@@ -0,0 +1,109 @@
...
+from nemo_evaluator_launcher.common.config_models import (
+    EvaluationModel,
+    MountsModel,
+    TaskModel,
+)
+
+
+class TestMountsModel:
+    def test_valid(self):
+        MountsModel(deployment={"/host/model": "/model"}, mount_home=False)
```
marta-sd: maybe we could add assertions to all the tests checking if the fields have expected values?

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/functional.py
```diff
@@ -54,6 +60,114 @@ def get_tasks_list() -> list[list[Any]]:
     return data
 
 
+def _validate_config_sections(cfg: Any) -> None:
```
marta-sd: I think this can be `RunConfig` and we can simplify the function body to expect this object. We use this function only inside `run_eval`, I don't think we'll ever apply it to completely raw inputs

packages/nemo-evaluator-launcher/tests/unit_tests/test_config_models.py
AWarno: ok

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/functional.py
AWarno: ok

AWarno: do you mean typing should be fixed?

marta-sd: yes, I mean we should have typing `cfg: RunConfig` and then in the function body we can assume it belongs to this class (unless I'm missing something and we actually want to run this function on inputs that are not RunConfig type)

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/functional.py
```diff
@@ -54,6 +66,98 @@ def get_tasks_list() -> list[list[Any]]:
...
+) -> None:
+    """Warn if nemo_evaluator_config params are not referenced in the task's command template.
+
+    Uses real packaged IRs — no network calls required. Needs cfg to merge task config via get_eval_factory_config.
+    """
+    if not evaluation.tasks:
+        return
+
+    try:
+        mapping = load_tasks_mapping()
```
marta-sd: What if the user specified a custom container for their evaluation? I think we need to check that when we loop over the tasks (L126) and load a custom mapping with `container_mapping = load_tasks_mapping(from_container=task.container)`

PR #822, opened 2026-03-09
feat(exporters): add Hugging Face Hub exporter

viviennezhangx:
## Summary

Adds a new **`huggingface`** exporter to NeMo Evaluator Launcher that pushes evaluation results as `.eval_results/` YAML files to Hugging Face Hub model repos, enabling automatic benchmark leaderboard integration.

This pairs with ongoing work to register `nemo-evaluator` as a supported evaluation framework on the HF Hub and add `eval.yaml` configs to benchmark datasets like `nvidia/compute-eval` and `livecodebench/code_generation_lite`.

### What's included

- **`huggingface.py`** — New exporter following the `BaseExporter` pattern, using `huggingface_hub.HfApi.upload_file` to push results
- **`test_huggingface_exporter.py`** — Unit tests covering helpers, config validation, metric selection, and export logic with mocked HfApi
- **`huggingface.yaml`** — Config template for `auto_export` integration
- **`__init__.py`** — Updated to register `HuggingFaceExporter`
- **`pyproject.toml`** — Added `huggingface_hub` as optional dependency (`pip install nemo-evaluator-launcher[huggingface]`)

### Key features

- Supports **PR mode** (default, safer) and **direct push** mode
- Auto-selects metrics with **substring matching** (e.g., `pass_at_1` matches `compute_eval_pass_at_1`)
- Includes **container info** in notes for reproducibility
- Follows the HF Evaluation Results spec: https://huggingface.co/docs/hub/eval-results

### Usage

```yaml
export:
  huggingface:
    dataset_id: nvidia/compute-eval
    task_id: compute_eval
    model_repo: nvidia/Llama-3.1-Nemotron-70B-Instruct-HF
    metric_name: pass_at_1
    create_pr: true
```

### Related PRs

- Framework registration on HF Hub: https://github.com/huggingface/huggingface.js/pull/2016
- `eval.yaml` for nvidia/compute-eval: https://huggingface.co/datasets/nvidia/compute-eval/discussions/6
- `eval.yaml` for livecodebench: https://huggingface.co/datasets/livecodebench/code_generation_lite/discussions/12

## Test plan

- [x] Unit tests pass for all helper functions (`_sanitize_filename`, `_select_metric`, `_build_eval_result_entry`, `_format_eval_results_yaml`)
- [x] Config validation tests (required fields, env var token fallback, explicit token override)
- [x] Export tests with mocked HfApi (success, upload failure, missing metric, multiple jobs, direct push mode, container notes)
- [x] Ruff format and lint checks pass
- [ ] Integration test with live HF Hub (manual, requires HF token)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/exporters/huggingface.py
+++ b/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/exporters/huggingface.py
@@ -1,0 +1,279 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+"""Export evaluation results to Hugging Face Hub leaderboards.
+
+Pushes results as `.eval_results/` YAML files to model repos on the Hub,
+enabling automatic leaderboard integration for benchmarks that define an
+`eval.yaml` with `evaluation_framework: nemo-evaluator`.
+
+See: https://huggingface.co/docs/hub/eval-results
+"""
+
+import os
+import re
+from datetime import datetime, timezone
+from typing import Any, Dict, List, Optional, Tuple
+
+import yaml
+from pydantic import Field, model_validator
+
+from nemo_evaluator_launcher.common.logging_utils import logger
+from nemo_evaluator_launcher.exporters.base import BaseExporter, ExportConfig
+from nemo_evaluator_launcher.exporters.registry import register_exporter
+from nemo_evaluator_launcher.exporters.utils import DataForExport
+
+
+class HuggingFaceExporterConfig(ExportConfig):
+    """Configuration for HuggingFaceExporter."""
+
+    dataset_id: str = Field(
+        ...,
+        description="HF Hub dataset ID of the benchmark (e.g., 'nvidia/compute-eval'). "
+        "Must have an eval.yaml defining it as a benchmark.",
+    )
+    task_id: str = Field(
+        ...,
+        description="Task ID as defined in the benchmark's eval.yaml (e.g., 'compute_eval').",
+    )
+    model_repo: str = Field(
+        ...,
+        description="HF Hub model repo to push results to (e.g., 'nvidia/Llama-3.1-Nemotron-70B-Instruct-HF').",
+    )
+    metric_name: Optional[str] = Field(
+        default=None,
+        description="Metric key from results.yml to use as the score value. "
+        "If not set, uses the first metric found. Supports substring matching.",
+    )
+    token: Optional[str] = Field(
+        default=None,
+        description="HF Hub token. Defaults to HF_TOKEN env var or cached login.",
+    )
+    create_pr: bool = Field(
+        default=True,
+        description="If True, submit results as a Pull Request (community contribution). "
+        "If False, push directly to main branch (requires write access).",
+    )
+    notes: Optional[str] = Field(
+        default=None,
+        description="Optional notes about the evaluation setup "
+        "(e.g., 'Pass@1, k=1', 'chain-of-thought').",
+    )
+    source_url: Optional[str] = Field(
+        default=None,
+        description="Optional URL linking to evaluation traces or logs.",
+    )
+    source_name: Optional[str] = Field(
+        default=None,
+        description="Optional display name for the source link.",
+    )
+
+    @model_validator(mode="before")
+    @classmethod
+    def _resolve_token(cls, data: Any) -> Any:
+        if isinstance(data, dict) and not data.get("token"):
+            data["token"] = os.environ.get("HF_TOKEN")
+        return data
+
+
+def _sanitize_filename(name: str) -> str:
+    """Convert a dataset ID or string into a safe filename component."""
+    return re.sub(r"[^a-zA-Z0-9_-]", "_", name)
+
+
+def _select_metric(
+    metrics: Dict[str, float], metric_name: Optional[str]
+) -> Tuple[str, float]:
+    """Select a metric from the results dict.
+
+    If metric_name is provided, finds the first key containing that substring.
+    Otherwise, returns the first metric.
+
+    Returns:
+        Tuple of (metric_key, metric_value).
+
+    Raises:
+        ValueError: If no matching metric is found or metrics dict is empty.
+    """
+    if not metrics:
+        raise ValueError("No metrics found in evaluation results")
+
+    if metric_name is None:
+        key = next(iter(metrics))
+        return key, metrics[key]
+
+    for key, value in metrics.items():
+        if metric_name in key:
+            return key, value
+
+    raise ValueError(
+        f"Metric '{metric_name}' not found in results. "
+        f"Available metrics: {list(metrics.keys())}"
+    )
+
+
+def _build_eval_result_entry(
+    dataset_id: str,
+    task_id: str,
+    value: float,
+    date: str,
+    notes: Optional[str] = None,
+    source_url: Optional[str] = None,
+    source_name: Optional[str] = None,
+) -> Dict[str, Any]:
+    """Build a single .eval_results/ entry conforming to the HF spec."""
+    entry: Dict[str, Any] = {
+        "dataset": {
+            "id": dataset_id,
+            "task_id": task_id,
+        },
+        "value": value,
+        "date": date,
+    }
+
+    if source_url:
+        source: Dict[str, str] = {"url": source_url}
+        if source_name:
+            source["name"] = source_name
+        entry["source"] = source
+
+    if notes:
+        entry["notes"] = notes
+
+    return entry
+
+
+def _format_eval_results_yaml(entries: List[Dict[str, Any]]) -> str:
+    """Serialize eval result entries to YAML matching the HF spec."""
+    return yaml.dump(
+        entries,
+        default_flow_style=False,
+        sort_keys=False,
+        allow_unicode=True,
+    )
+
+
+@register_exporter("huggingface")
+class HuggingFaceExporter(BaseExporter):
+    """Export evaluation results to Hugging Face Hub model repos.
+
+    Results are formatted as `.eval_results/` YAML files conforming to the
+    HF Evaluation Results specification. The Hub automatically aggregates
+    these into benchmark leaderboards.
+
+    Config keys:
+        dataset_id (str): HF benchmark dataset ID (e.g., 'nvidia/compute-eval')
+        task_id (str): Task ID from the benchmark's eval.yaml
+        model_repo (str): HF model repo to push results to
+        metric_name (str): Metric key from results.yml to use as score
+        token (str): HF Hub API token (defaults to HF_TOKEN env var)
+        create_pr (bool): Submit as PR (default True) vs direct push
+        notes (str): Optional evaluation setup details
+        source_url (str): Optional link to evaluation traces
+        source_name (str): Optional display name for source link
+    """
+
+    config_class = HuggingFaceExporterConfig
+
+    def export_jobs(
+        self, data_for_export: List[DataForExport]
+    ) -> Tuple[List[str], List[str], List[str]]:
+        """Export job results to HF Hub as .eval_results/ YAML files."""
+        try:
+            from huggingface_hub import HfApi
+        except ImportError:
+            logger.error(
+                "huggingface_hub is required for the HuggingFace exporter. "
+                "Install it with: pip install huggingface_hub"
+            )
+            return [], [d.job_id for d in data_for_export], []
+
+        api = HfApi(token=self.config.token)
+
+        success_jobs = []
+        failed_jobs = []
+        skipped_jobs = []
+
+        entries = []
+        job_ids_for_entries = []
+
+        for data in data_for_export:
+            try:
+                metric_key, metric_value = _select_metric(
+                    data.metrics, self.config.metric_name
+                )
+                logger.info(
+                    f"Job {data.job_id}: selected metric '{metric_key}' = {metric_value}"
+                )
+
+                eval_date = datetime.fromtimestamp(
+                    data.timestamp, tz=timezone.utc
+                ).strftime("%Y-%m-%d")
+
+                notes_parts = []
+                if self.config.notes:
+                    notes_parts.append(self.config.notes)
+                if data.container:
+                    notes_parts.append(f"container: {data.container}")
+                notes = ", ".join(notes_parts) if notes_parts else None
+
+                entry = _build_eval_result_entry(
+                    dataset_id=self.config.dataset_id,
+                    task_id=self.config.task_id,
+                    value=round(metric_value, 4),
+                    date=eval_date,
+                    notes=notes,
+                    source_url=self.config.source_url,
+                    source_name=self.config.source_name,
+                )
+                entries.append(entry)
+                job_ids_for_entries.append(data.job_id)
+
+            except Exception as e:
+                logger.error(f"Failed to prepare result for job {data.job_id}: {e}")
+                failed_jobs.append(data.job_id)
+
+        if not entries:
+            return success_jobs, failed_jobs, skipped_jobs
+
+        yaml_content = _format_eval_results_yaml(entries)
+        filename = f".eval_results/{_sanitize_filename(self.config.dataset_id)}.yaml"
+
+        commit_message = (
+            f"Add {self.config.task_id} evaluation results (via NeMo Evaluator)"
+        )
+
+        try:
+            api.upload_file(
+                path_or_fileobj=yaml_content.encode("utf-8"),
+                path_in_repo=filename,
+                repo_id=self.config.model_repo,
+                repo_type="model",
+                commit_message=commit_message,
+                create_pr=self.config.create_pr,
+            )
+
+            action = "PR submitted" if self.config.create_pr else "pushed"
+            logger.info(
+                f"Results {action} to {self.config.model_repo}/{filename} "
+                f"({len(entries)} entries)"
+            )
+            success_jobs.extend(job_ids_for_entries)
+
+        except Exception as e:
+            logger.error(f"Failed to upload results to {self.config.model_repo}: {e}")
+            failed_jobs.extend(job_ids_for_entries)
+
+        return success_jobs, failed_jobs, skipped_jobs
--- a/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/resources/config_templates/export/huggingface.yaml
+++ b/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/resources/config_templates/export/huggingface.yaml
@@ -1,0 +1,14 @@
+execution:
+  auto_export:
+    destinations: ["huggingface"]
+
+export:
+  huggingface:
+    dataset_id: ???  # HF Hub dataset ID of the benchmark (required)
+    task_id: ???     # Task ID from the benchmark's eval.yaml (required)
+    model_repo: ???  # HF Hub model repo to push results to (required)
+    # metric_name: pass_at_1  # Metric key (substring match); omit to use first metric
+    create_pr: true  # Submit as PR (true) or push directly (false)
+    # notes: "Pass@1, k=1"
+    # source_url: https://example.com/eval-logs
+    # source_name: Evaluation Logs
--- a/packages/nemo-evaluator-launcher/tests/unit_tests/exporters/test_huggingface_exporter.py
+++ b/packages/nemo-evaluator-launcher/tests/unit_tests/exporters/test_huggingface_exporter.py
@@ -1,0 +1,425 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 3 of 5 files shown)


packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/exporters/huggingface.py
```diff
@@ -0,0 +1,279 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION. All rights reserved.
```
pribalta, line 1: [Power Review] 🔴 **New user-facing exporter ships without documentation** · 95% confidence

The knowledge document states: 'Features without docs do not exist from the user's perspective.' This PR adds a new exporter with multiple config options (dataset_id, task_id, model_repo, metric_name, create_pr, notes, source_url, source_name), a new optional dependency, and a new auto_export destination — but no documentation updates (README, docs site, or usage guide) are included. The config template (huggingface.yaml) and PR description contain usage details, but these are not user-discoverable documentation.

💡 **Suggestion:** Add documentation covering: (1) what the HuggingFace exporter does and why a user would enable it, (2) prerequisite setup (HF account, token, eval.yaml on the benchmark dataset), (3) a complete worked example from config to leaderboard result, (4) explanation of PR mode vs direct push and when to use each. This should live alongside existing exporter documentation.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/exporters/huggingface.py
```diff
@@ -0,0 +1,279 @@
...
+    )
+    source_name: Optional[str] = Field(
+        default=None,
+        description="Optional display name for the source link.",
+    )
+
+    @model_validator(mode="before")
+    @classmethod
+    def _resolve_token(cls, data: Any) -> Any:
+        if isinstance(data, dict) and not data.get("token"):
```
pribalta, line 86: [Power Review] 🔴 **Missing token produces no actionable error message** · 92% confidence

Per the 'Undocumented Required Setup Step' rule: if a required env var is missing, the error message should say exactly what to set and where to get the value. When neither `token` config nor `HF_TOKEN` env var is set, `self.config.token` is `None`, which gets passed silently to `HfApi(token=None)`. The eventual failure will come from the HF Hub API as an opaque 401/403 HTTP error, not a clear message telling the user to set `HF_TOKEN` or pass `token` in config. A junior user will not know how to obtain or configure a HF token.

💡 **Suggestion:** Add a validation check (e.g., in the `model_validator` or at the start of `export_jobs`) that raises a clear error when token is None: `"HF Hub token not found. Set the HF_TOKEN environment variable or add 'token' to your huggingface export config. Get a token at https://huggingface.co/settings/tokens"`

packages/nemo-evaluator-launcher/tests/unit_tests/exporters/test_huggingface_exporter.py
```diff
@@ -0,0 +1,425 @@
...
+
+
+# ---------------------------------------------------------------------------
+# HuggingFaceExporterConfig
+# ---------------------------------------------------------------------------
+
+
+class TestHuggingFaceExporterConfig:
+    def test_required_fields(self):
+        with pytest.raises(Exception):
```
pribalta, line 217: [Power Review] 🔴 **Test asserts only that an exception is raised, not which one** · 90% confidence

Per the knowledge document, AI produces tests that verify too little. This test catches any `Exception` without checking the type or message, meaning it would pass even if the constructor raised an unrelated error (e.g., ImportError, TypeError). Other tests in this file correctly use `match=` parameter.

💡 **Suggestion:** Use `pytest.raises(ValidationError)` (from pydantic) or at minimum add a `match=` parameter to verify the error is about missing required fields, not an unrelated crash.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/exporters/huggingface.py
```diff
@@ -0,0 +1,279 @@
...
+            )
+
+            action = "PR submitted" if self.config.create_pr else "pushed"
+            logger.info(
+                f"Results {action} to {self.config.model_repo}/{filename} "
+                f"({len(entries)} entries)"
+            )
+            success_jobs.extend(job_ids_for_entries)
+
+        except Exception as e:
```
pribalta, line 275: [Power Review] 🔴 **Upload failure error message exposes raw exception without guidance** · 88% confidence

Per the 'Cryptic Error Message' rule: errors should not expose internal implementation details without telling the user what went wrong or how to fix it. The catch-all `except Exception as e` at line 275 logs the raw exception (e.g., '403 Forbidden') without explaining common causes (bad token, no write access, repo doesn't exist) or how to fix them. A junior user seeing 'Failed to upload results to nvidia/model: 403 Forbidden' won't know whether the issue is their token, permissions, or repo name.

💡 **Suggestion:** Catch specific `huggingface_hub` exceptions (e.g., `HfHubHTTPError`) and provide targeted guidance. For 401/403: 'Authentication failed. Check that your HF_TOKEN has write access to the model repo.' For 404: 'Model repo not found. Verify model_repo is correct.' Fall back to a generic message with a link to troubleshooting docs.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/exporters/huggingface.py
```diff
@@ -0,0 +1,279 @@
...
+
+    config_class = HuggingFaceExporterConfig
+
+    def export_jobs(
+        self, data_for_export: List[DataForExport]
+    ) -> Tuple[List[str], List[str], List[str]]:
+        """Export job results to HF Hub as .eval_results/ YAML files."""
+        try:
+            from huggingface_hub import HfApi
+        except ImportError:
```
pribalta, line 196: [Power Review] ⚠️ **ImportError silently fails all jobs without raising or printing install command for extras** · 90% confidence

Per the 'Undocumented Required Setup Step' rule: when a dependency is missing, the error should tell the user exactly how to install it. The error message says `pip install huggingface_hub` but the canonical install path via pyproject.toml is `pip install nemo-evaluator-launcher[huggingface]`. A junior user following the logged message may install a version that conflicts with the project's pinned range. Additionally, the failure is only logged (not raised), so the exporter silently marks all jobs as failed with no obvious indication of why.

💡 **Suggestion:** Update the error message to: `pip install nemo-evaluator-launcher[huggingface]` (matching the optional dependency in pyproject.toml). Consider raising the error or printing it to stderr so it's visible even if logging is not configured at the right level.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/resources/config_templates/export/huggingface.yaml
```diff
@@ -0,0 +1,14 @@
+execution:
```
pribalta, line 1: [Power Review] ⚠️ **Config template missing token/auth setup instructions** · 90% confidence

Per the 'Undocumented Required Setup Step' rule: the config template does not mention that an HF token is required for uploading results. A junior user copying this template will hit an auth error on first run. The `token` field and `HF_TOKEN` env var fallback are not documented in the template.

💡 **Suggestion:** Add a commented-out `token` field with instructions: `# token: ...  # HF Hub token. Set HF_TOKEN env var or provide here. Get one at https://huggingface.co/settings/tokens`

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/exporters/huggingface.py
```diff
@@ -0,0 +1,279 @@
...
+        for data in data_for_export:
+            try:
+                metric_key, metric_value = _select_metric(
+                    data.metrics, self.config.metric_name
+                )
+                logger.info(
+                    f"Job {data.job_id}: selected metric '{metric_key}' = {metric_value}"
+                )
+
+                eval_date = datetime.fromtimestamp(
```
pribalta, line 221: [Power Review] ⚠️ **Missing dataset version pinning breaks reproducibility** · 85% confidence

The knowledge document flags 'Reproducibility Broken' as an ERROR when dataset versions are unpinned and software versions are not recorded in output artifacts. The exported YAML records dataset_id and date but does not capture any version identifier for the benchmark dataset (e.g., a commit SHA or dataset revision). Two runs against different revisions of the same dataset_id would produce results that appear comparable but are not, constituting silent cross-version comparison. The HF Hub datasets support revision pinning via commit SHA.

💡 **Suggestion:** Add an optional `dataset_revision` config field (default to resolving the current HEAD SHA of the dataset repo via `HfApi.dataset_info`). Include the revision in the exported YAML entry so results are traceable to a specific dataset snapshot. At minimum, log a warning when no revision is pinned.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/exporters/huggingface.py
```diff
@@ -0,0 +1,279 @@
...
+        if metric_name in key:
+            return key, value
+
+    raise ValueError(
+        f"Metric '{metric_name}' not found in results. "
+        f"Available metrics: {list(metrics.keys())}"
+    )
+
+
+def _build_eval_result_entry(
```
pribalta, line 127: [Power Review] ⚠️ **No NeMo Evaluator version recorded in exported artifact** · 82% confidence

The knowledge document lists 'missing software version recording in output artifacts' as a reproducibility error. The exported YAML includes date and optional notes/container but does not record the nemo-evaluator-launcher version or the huggingface_hub library version. A senior engineer running evaluations at scale needs to trace results back to exact tool versions.

💡 **Suggestion:** Include the `nemo-evaluator-launcher` package version (e.g., via `importlib.metadata.version`) and optionally the `huggingface_hub` version in the YAML entry metadata or notes field. This makes every exported result fully traceable.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/exporters/huggingface.py
```diff
@@ -0,0 +1,279 @@
...
+        ValueError: If no matching metric is found or metrics dict is empty.
+    """
+    if not metrics:
+        raise ValueError("No metrics found in evaluation results")
+
+    if metric_name is None:
+        key = next(iter(metrics))
+        return key, metrics[key]
+
+    for key, value in metrics.items():
```
pribalta, line 117: [Power Review] ⚠️ **Ambiguous substring metric matching with no warning** · 82% confidence

Per the knowledge doc's 'Debugging Affordance Missing' rule: `_select_metric` uses substring matching (line 118: `if metric_name in key`) which silently returns the first match. If a user specifies `pass_at_1`, it will match both `compute_eval_pass_at_1` and `compute_eval_pass_at_10` (whichever appears first in dict iteration order). The user gets no warning about multiple matches, making it hard to diagnose incorrect metric values in results.

💡 **Suggestion:** When multiple keys match the substring, either (a) log a warning listing all matches and which one was selected, or (b) require an exact match first and fall back to substring only if no exact match exists. At minimum, log the full list of candidates at debug level.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/exporters/huggingface.py
```diff
@@ -0,0 +1,279 @@
...
+                    value=round(metric_value, 4),
+                    date=eval_date,
+                    notes=notes,
+                    source_url=self.config.source_url,
+                    source_name=self.config.source_name,
+                )
+                entries.append(entry)
+                job_ids_for_entries.append(data.job_id)
+
+            except Exception as e:
```
pribalta, line 244: [Power Review] ⚠️ **Broad except catches all exceptions without specificity** · 80% confidence

Per the knowledge document, AI often generates 'except Exception' hiding real failures. The inner except on line 244 catches all exceptions during metric selection and entry building. Errors like TypeError or AttributeError from bugs in the code itself would be silently swallowed into the failed_jobs list rather than surfacing as real bugs. Only ValueError is expected from _select_metric.

💡 **Suggestion:** Catch `ValueError` specifically (the only documented exception from `_select_metric`) instead of bare `Exception`. If other specific exceptions are expected, catch those explicitly. Let unexpected exceptions propagate.

packages/nemo-evaluator-launcher/tests/unit_tests/exporters/test_huggingface_exporter.py
```diff
@@ -0,0 +1,425 @@
...
+            "compute_eval_pass_at_1": 0.61,
+            "compute_eval_pass_at_3": 0.74,
+        },
+        harness="compute-eval",
+        task="compute-eval.CUDA",
+        container="nvcr.io/nvidia/eval-factory/compute-eval:26.01",
+        executor="local",
+        invocation_id="abc123",
+        job_id="abc123.0",
+        timestamp=time.time(),
```
pribalta, line 71: [Power Review] ⚠️ **Flaky test pattern: timestamp uses wall-clock time** · 75% confidence

The `sample_data_for_export` fixture uses `time.time()` for the timestamp, making the test dependent on wall-clock time. Per the knowledge document, tests that depend on `time.time()` are flagged as flaky test patterns. If date-boundary logic is ever tested, this would produce different results depending on when the test runs.

💡 **Suggestion:** Use a fixed timestamp value (e.g., `timestamp=1709424000.0`) to make tests fully deterministic and reproducible.

packages/nemo-evaluator-launcher/tests/unit_tests/exporters/test_huggingface_exporter.py
```diff
@@ -0,0 +1,425 @@
...
+        assert len(success) == 2
+        assert len(failed) == 0
+
+        uploaded_yaml = mock_api.upload_file.call_args[1]["path_or_fileobj"].decode(
+            "utf-8"
+        )
+        parsed = yaml.safe_load(uploaded_yaml)
+        assert len(parsed) == 2
+
+    @patch("nemo_evaluator_launcher.exporters.huggingface.HfApi")
```
pribalta, line 379: [Power Review] ⚠️ **Test name implies verification but has no return-value assertions** · 75% confidence

Per the knowledge document, test descriptions should match behavior. `test_direct_push_mode` only asserts on the mock call argument (`create_pr is False`) but doesn't verify the return values (success/failed/skipped). The test calls `export_jobs` but discards its return value, so it doesn't verify the export actually succeeded in direct-push mode.

💡 **Suggestion:** Capture the return value and assert that the job appears in `success` and not in `failed`: `success, failed, skipped = exporter.export_jobs([sample_data_for_export]); assert len(success) == 1`.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/exporters/huggingface.py
```diff
@@ -0,0 +1,279 @@
...
+
+    if notes:
+        entry["notes"] = notes
+
+    return entry
+
+
+def _format_eval_results_yaml(entries: List[Dict[str, Any]]) -> str:
+    """Serialize eval result entries to YAML matching the HF spec."""
+    return yaml.dump(
```
pribalta, line 160: [Power Review] ⚠️ **yaml.dump used instead of yaml.safe_dump for serialization** · 75% confidence

Per the knowledge document (Deserialization of Untrusted Data, CWE-502), safe YAML functions should be preferred. While `yaml.dump` is used here for serialization (not deserialization), it can emit Python-specific YAML tags (e.g., `!!python/object`) if the data structure inadvertently contains non-primitive Python objects. Since `entries` is built from user-provided config values and metric data that flow through without strict type enforcement, a non-primitive value could leak into the output. Using `yaml.safe_dump` constrains output to standard YAML types and is the safer default.

💡 **Suggestion:** Replace `yaml.dump(entries, ...)` with `yaml.safe_dump(entries, ...)` to ensure only standard YAML types are emitted.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/exporters/huggingface.py
```diff
@@ -0,0 +1,279 @@
...
+    model_repo: str = Field(
+        ...,
+        description="HF Hub model repo to push results to (e.g., 'nvidia/Llama-3.1-Nemotron-70B-Instruct-HF').",
+    )
+    metric_name: Optional[str] = Field(
+        default=None,
+        description="Metric key from results.yml to use as the score value. "
+        "If not set, uses the first metric found. Supports substring matching.",
+    )
+    token: Optional[str] = Field(
```
pribalta, line 60: [Power Review] ⚠️ **Token handling lacks clear error messaging for auth failures** · 75% confidence

The knowledge document notes 'Onboarding Regression' concerns — new features should not create friction for first-time users. The token resolution silently falls back from explicit token → HF_TOKEN env var → None (which relies on cached huggingface-cli login). If none are configured, the user will get a cryptic 403 or authentication error from the HF API at upload time (line 259), not a clear message about how to authenticate.

💡 **Suggestion:** Add an early validation step that checks token availability before attempting upload, with a user-friendly error message like: 'No HF Hub token found. Set HF_TOKEN environment variable, pass token in config, or run `huggingface-cli login`. See [docs link].'

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/resources/config_templates/export/huggingface.yaml
```diff
@@ -0,0 +1,14 @@
...
+  auto_export:
+    destinations: ["huggingface"]
+
+export:
+  huggingface:
+    dataset_id: ???  # HF Hub dataset ID of the benchmark (required)
```
pribalta, line 7: [Power Review] ⚠️ **Config template uses `???` placeholder without explaining the convention** · 75% confidence

Per the 'Complex Config Without Inline Comments' rule and 'Config Example Missing or Outdated': the `???` placeholder is an OmegaConf convention for required fields, but a junior user unfamiliar with OmegaConf will not understand that `???` means 'you must replace this'. The comments describe the fields but don't include example values that would help the user understand the expected format.

💡 **Suggestion:** Add concrete example values as comments. For example: `dataset_id: ???  # Required. Example: nvidia/compute-eval`. This helps users understand both the format and that the field must be replaced.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/exporters/huggingface.py
```diff
@@ -0,0 +1,279 @@
...
+from nemo_evaluator_launcher.common.logging_utils import logger
+from nemo_evaluator_launcher.exporters.base import BaseExporter, ExportConfig
+from nemo_evaluator_launcher.exporters.registry import register_exporter
+from nemo_evaluator_launcher.exporters.utils import DataForExport
+
+
+class HuggingFaceExporterConfig(ExportConfig):
+    """Configuration for HuggingFaceExporter."""
+
+    dataset_id: str = Field(
```
pribalta, line 42: [Power Review] ⚠️ **High prerequisite knowledge required without onboarding guidance** · 72% confidence

The knowledge document warns about 'Accessibility Barrier': features requiring specialized knowledge without a simpler alternative path. This exporter requires users to understand HF Hub concepts (dataset_id, task_id, eval.yaml spec, model repos, PR workflows) and have a pre-configured benchmark dataset with an eval.yaml file. There is no validation or guidance when these prerequisites are not met — e.g., if the benchmark dataset doesn't have an eval.yaml, the upload will succeed but nothing will appear on any leaderboard, with no warning to the user.

💡 **Suggestion:** Consider: (1) adding a preflight check that verifies the target dataset has an eval.yaml with 'nemo-evaluator' as a supported framework, (2) providing a clear error or warning when prerequisites are missing, (3) linking to setup instructions in error messages. At minimum, document the prerequisite chain in user-facing docs.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/exporters/huggingface.py
```diff
@@ -0,0 +1,279 @@
...
+            data["token"] = os.environ.get("HF_TOKEN")
+        return data
+
+
+def _sanitize_filename(name: str) -> str:
+    """Convert a dataset ID or string into a safe filename component."""
+    return re.sub(r"[^a-zA-Z0-9_-]", "_", name)
+
+
+def _select_metric(
```
pribalta, line 96: [Power Review] ⚠️ **Substring metric matching is surprising and undiscoverable** · 70% confidence

The knowledge document warns about 'Inconsistent Product Surface' — UX patterns that may confuse users. The metric_name field uses substring matching (e.g., 'pass_at_1' matches 'compute_eval_pass_at_1'), which is non-obvious behavior. If a user has metrics like 'pass_at_1' and 'pass_at_10', specifying 'pass_at_1' would match 'pass_at_1' first but could surprise users expecting exact matching. The fallback to 'first metric found' when metric_name is omitted is also order-dependent and potentially surprising.

💡 **Suggestion:** Document the substring matching behavior prominently. Consider: (1) logging which metric was matched and why, (2) warning when multiple metrics match the substring, (3) supporting exact match as the default with an explicit 'substring:' prefix for fuzzy matching.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/exporters/huggingface.py
```diff
@@ -0,0 +1,279 @@
...
+            api.upload_file(
+                path_or_fileobj=yaml_content.encode("utf-8"),
+                path_in_repo=filename,
+                repo_id=self.config.model_repo,
+                repo_type="model",
+                commit_message=commit_message,
+                create_pr=self.config.create_pr,
+            )
+
+            action = "PR submitted" if self.config.create_pr else "pushed"
```
pribalta, line 268: [Power Review] 💡 **Success log doesn't include URL to the created PR or uploaded file** · 82% confidence

Per the 'Unclear Output Location or Format' rule: after a successful export, the log says 'Results PR submitted to nvidia/model/.eval_results/file.yaml (2 entries)' but doesn't include the actual HF Hub URL where the user can view or approve the PR. The `upload_file` API returns a `CommitInfo` object with a `pr_url` field that could be logged.

💡 **Suggestion:** Capture the return value of `api.upload_file()` and log the PR URL (available as `commit_info.pr_url` when `create_pr=True`) so the user can immediately click through to review their submission.

packages/nemo-evaluator-launcher/tests/unit_tests/exporters/test_huggingface_exporter.py
```diff
@@ -0,0 +1,425 @@
...
+        timestamp=time.time(),
+    )
+
+
+# ---------------------------------------------------------------------------
+# _sanitize_filename
+# ---------------------------------------------------------------------------
+
+
+class TestSanitizeFilename:
```
pribalta, line 80: [Power Review] 💡 **Missing boundary tests for _sanitize_filename** · 70% confidence

Per the knowledge document, AI tests typical values but skip boundaries. `_sanitize_filename` is not tested with boundary inputs such as an empty string, a string of only special characters, or extremely long strings.

💡 **Suggestion:** Add tests for `_sanitize_filename('')`, `_sanitize_filename('///...')`, and a very long input string to verify the function handles edge cases correctly.

marta-sd:
Closing as we discussed that it's better to add this feature as a dedicated `nel publish` command

PR #825, opened 2026-03-09
Make cache-related resources group writable

janekl:
Goal: allow modifying results by group members when running jobs using dedicated service accounts. 

Details: SQLite creates database files with hardcoded 0644 permissions (`SQLITE_DEFAULT_FILE_PERMISSIONS`), preventing other group members from sharing or modifying the disk cache later. Similarly, the cache directory was created with 0755, blocking group writes. 

Changes:
* Set cache directory permissions to 0775 (group-writable) in `os.makedirs`
* Add `os.chmod` to 0664 for cache.db and its WAL/SHM companion files after each new SQLite connection (`sqlite3.connect()` does not expose a file mode parameter)

This ensures users under the same group can share and modify the cache. The WAL/SHM files are also handled to support crash recovery by other group members.

### DIFF
--- a/packages/nemo-evaluator/src/nemo_evaluator/adapters/caching/diskcaching.py
+++ b/packages/nemo-evaluator/src/nemo_evaluator/adapters/caching/diskcaching.py
@@ -470,7 +470,7 @@ class Cache:
 
         if not op.isdir(directory):
             try:
-                os.makedirs(directory, 0o755)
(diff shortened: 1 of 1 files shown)


marta-sd:
We discussed offline that the change would introduce a security risk and adviced to use a cleanup job (`post_cmd`) to modify permissions and achieve the same outcome

PR #830, opened 2026-03-10
feat: deploy auxiliary endpoints

wprazuch:


packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
```diff
@@ -1785,17 +1963,258 @@ def _generate_deployment_srun_command(
...
+    env_var_names = list(aux.env_vars.keys()) if aux.env_vars else []
+    # Always add MASTER_IP to the environment variables
+    master_ip_var = f"{prefix}_MASTER_IP"
+    if master_ip_var not in env_var_names:
+        env_var_names.append(master_ip_var)
+
+    if aux.num_instances > 1:
+        # Multi-instance mode: loop over instances, collect PIDs
+        nodes_per_instance = aux.num_nodes // aux.num_instances
+        n_tasks = aux.cfg.get("n_tasks", nodes_per_instance)
```
agronskiy: `bug`: fallback to nodes_per_instance seems like a bug.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
```diff
@@ -1785,17 +1963,258 @@ def _generate_deployment_srun_command(
...
+        s += f'export {all_node_ips_var}=$(IFS=,; echo "${{{prefix}_NODES_IPS_ARRAY[*]}}")\n'
+        if all_node_ips_var not in env_var_names:
+            env_var_names.append(all_node_ips_var)
+
+        env_setup = f"export PROC_ID=${{SLURM_PROCID:-0}} NODES_PER_INSTANCE={nodes_per_instance}"
+        script = (
+            f"{env_setup} && {create_script_cmd.cmd} && bash {name}_deployment_cmd.sh"
+        )
+
+        if pre_cmd:
```
agronskiy, line 2062: `question`: how is this different from the one above around L2000? Seems like two times called _str_to_echo_command wtih the same arg, no?

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
```diff
@@ -747,23 +769,60 @@ def _create_slurm_sbatch_script(
...
+        # Split nodes: model first, then each auxiliary deployment
+        s += "# Split nodes between model deployment and auxiliary deployments\n"
+        s += f"MODEL_NUM_NODES={cfg.execution.num_nodes}\n"
+        for aux in aux_deployments:
+            s += f"{aux.env_prefix}_NUM_NODES={aux.num_nodes}\n"
+        s += 'MODEL_NODES=("${ALL_NODES[@]:0:$MODEL_NUM_NODES}")\n'
+        offset_expr = "MODEL_NUM_NODES"
+        for aux in aux_deployments:
+            s += f'{aux.nodes_var}=("${{ALL_NODES[@]:${offset_expr}:${aux.env_prefix}_NUM_NODES}}")\n'
+            offset_expr = f"{offset_expr}+{aux.env_prefix}_NUM_NODES"
```
agronskiy, line 796: `question/suspicious`: I think bash would require `(( ... ))`  to apply arithmetic. Is it the case here w/ slurm?

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
```diff
@@ -747,23 +769,60 @@ def _create_slurm_sbatch_script(
...
-    s += "else\n"
-    s += '  nodes=( "$(hostname)" )\n'
-    s += "fi\n"
-    s += 'nodes_array=("${nodes[@]}")\n'
-    s += "if [[ ${#nodes_array[@]} -eq 0 ]]; then\n"
-    s += '  nodes_array=( "$(hostname)" )\n'
-    s += "fi\n"
-    s += 'export PRIMARY_NODE="${nodes_array[0]}"\n'
-    s += 'echo "PRIMARY_NODE: ${PRIMARY_NODE}"\n'
+    if has_aux_deployments:
```
agronskiy, line 773: `concern`: I have concern that this and the other branch (else one) look very very similar. I didn't skim deep but maybe the agent (not the same one) could check if that could be simplified.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
```diff
@@ -1785,17 +1963,258 @@ def _generate_deployment_srun_command(
...
+    s += f"proxy_config_file={proxy_cfg_path}\n"
+    s += f'for i in "${{!{prefix}_NODES_IPS_ARRAY[@]}}"; do\n'
+    s += f'    ip="${{{prefix}_NODES_IPS_ARRAY[$i]}}"\n'
+    s += f'    sed -i "s/{{{prefix}_IP_$i}}/$ip/g" "$proxy_config_file"\n'
+    s += "done\n"
+    s += "\n"
+
+    proxy_image = cfg.execution.get("proxy", {}).get("image", "haproxy:latest")
+    s += "srun --mpi pmix --overlap "
+    s += '--nodelist "${PRIMARY_NODE}" --nodes 1 --ntasks 1 '
```
agronskiy, line 2183: This is primary node's haproxy, is this coincidental?

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
wprazuch: Yes, that is for sure a bug, thanks for spotting it, I will fix it!

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
wprazuch: It works, so it still does this implicit math operation, but it's better to add this (( )) to become obvious

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
wprazuch: True, we should re-use the output of the command instead of calling it twice

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
wprazuch: That is true - I already tried to simplify it in the refactor [PR](https://github.com/NVIDIA-NeMo/Evaluator/pull/839/changes#diff-6d61bcd34d91ffbb01e7f8608dc9bfad157463ff2e6332ce6594ea756bb8c940), but could not properly test it. I will try to squeeze it in today so it's in a nice shape

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
wprazuch: Please keep me checked on this one, but if eval container runs there, then in order to ensure communication with the haproxy image for aux it should also run there, right?

wprazuch:
/ok to test [7501aa4](https://github.com/NVIDIA-NeMo/Evaluator/pull/830/commits/7501aa452b65cdb7ab4d44b86d7d56049ae92a74)

wprazuch:
/ok to test [f3a3950](https://github.com/NVIDIA-NeMo/Evaluator/pull/830/commits/f3a39501b5cbbe89f7264c2ac74558f83dcc85ad)

wprazuch:
/ok to test [51ec4b1](https://github.com/NVIDIA-NeMo/Evaluator/pull/830/commits/51ec4b18618807e75cba72a96c9c373cb5f39772)

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
wprazuch: Unfortunately I won't be able to accomplish it today 🙈 we can either leave it as it is or find someone to refactor this in my absence

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
wprazuch: Modified

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
wprazuch: changed

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
wprazuch: Fixed

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
```diff
@@ -1792,18 +1974,256 @@ def _generate_deployment_srun_command(
...
+    s += f"--container-image {proxy_image} "
+    s += f"--container-mounts {proxy_cfg_path}:/usr/local/etc/haproxy/haproxy.cfg:ro "
+    s += f"--output {remote_task_subdir}/logs/{name}-proxy-%A.log "
+    s += "haproxy -f /usr/local/etc/haproxy/haproxy.cfg &\n"
+    s += f"{aux.proxy_pid_var}=$!  # capture the PID of the {name} proxy background srun process\n"
+    s += f'echo "{name} proxy started with PID: ${aux.proxy_pid_var}"\n\n'
+
+    # Wait for proxy to be ready
+    s += _get_wait_for_server_handler(
```
agronskiy, line 2193: `bug`: positional here for timeout missing. Please instruct the agent to _always_ use kv arguments in Python if more than two arguments. It helps everyone. Did I already tell war story about switching the bounding box signatur in some tensorflow code from 4 ints to 4 ints but from YOLO format (x_center, y_center, height, width) to SSD (left, top, right, bottom) and subtle bugs popping all over? :D

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/helpers.py
```diff
@@ -283,8 +299,56 @@ def get_eval_factory_command(
     commands.append(create_post_script_cmd.cmd)
     debug.append(create_post_script_cmd.debug)
 
+    # Resolve auxiliary_deployments.NAME.FIELD references in the config.
+    # model_id is resolved to a literal; endpoint URLs use shell vars
+    # that envsubst resolves at runtime.
+    config_yaml = yaml.safe_dump(merged_nemo_evaluator_config)
+    for aux_name, aux_cfg_raw in cfg.get("auxiliary_deployments", {}).items():
```
agronskiy, line 308: `suggestion`: this and L332 might benefit from merging, or at least talking to agent about the logic there

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
AdamRajfer: Good catch! Fixed. I think that the agent might actually simply mirror current design as we have not used kwargs in any of the `_get_wait_for_server_handler` calls. I fixed it everywhere.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/helpers.py
AdamRajfer: Nice catch, refactored!

PR #832, opened 2026-03-10
feat: add --platform flag for BYOB container builds

laszkiewiczp:
Resolves EVAL-775

Adds an optional `--platform` flag to the BYOB CLI (`nemo-evaluator-byob`)
that allows users to target a specific platform when containerizing benchmarks                                                                                                                                                                                                                                                                                               (e.g. `linux/amd64`). This feature was necessary when running BYOB on local
machine and executing evaluation on a cluster with different architecture.

  - When `--platform` is set, uses `docker buildx build --platform <X> --load`
    to build for the specified architecture and import into the local daemon
  - When omitted, uses plain `docker build` for the host architecture (unchanged behavior)
  - Multi-platform builds (comma-separated values) are rejected with a clear error message,
    since `buildx --load` does not support multi-arch manifests and would require an immediate push
  - Extracts command construction into a `_get_build_command` helper for testability

### DIFF
--- a/docs/project.json
+++ b/docs/project.json
@@ -1,1 +1,1 @@
-{"name": "nemo-evaluator", "version": "0.2.2"}
+{"name": "nemo-evaluator", "version": "0.1.0"}
--- a/docs/uv.lock
+++ b/docs/uv.lock
@@ -1,5 +1,5 @@
 version = 1
-revision = 3
+revision = 2
 requires-python = ">=3.10, <3.14"
 resolution-markers = [
     "python_full_version >= '3.13'",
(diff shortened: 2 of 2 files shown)


packages/nemo-evaluator/src/nemo_evaluator/contrib/byob/containerize.py
```diff
@@ -254,12 +254,42 @@ def prepare_build_context(
...
+
+    Returns:
+        Command tokens ready for :func:`subprocess.run`.
+    """
+    base = (
+        ["docker", "buildx", "build", "--platform", platform, "--load"]
+        if platform
+        else ["docker", "build"]
+    )
+    return [*base, "-t", tag, context_dir]
```
prokotg, line 283: `nit`: if we do not allow multi-platform build (it's because IIUC manifest lists are not exportable in multiplatform and output type docker?) maybe we could bake in platform into the tag? I imagine people might still be interested in multiple platforms and images would be overwritten

PR #838, opened 2026-03-11
fix(deprecated-multiple-instances-flag): fix deprecated multiple instances

AWarno:
We need to decide what to do with deprecated multiple_instances flag

### DIFF
--- a/packages/nemo-evaluator-launcher/examples/nemotron/nemotron-3-super/local_nemotron-3-super-120b-a12b-base.yaml
+++ b/packages/nemo-evaluator-launcher/examples/nemotron/nemotron-3-super/local_nemotron-3-super-120b-a12b-base.yaml
@@ -1,0 +1,111 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
```diff
@@ -642,14 +642,14 @@ def _create_slurm_sbatch_script(
...
+    # deployment.multiple_instances is deprecated — use execution.num_instances and execution.num_nodes
     if cfg.deployment.get("multiple_instances") is not None:
-        logger.warning(
-            "deployment.multiple_instances is deprecated and will be "
-            "removed from config — use execution.num_instances instead."
+        raise ValueError(
+            "deployment.multiple_instances is deprecated and no longer supported. "
+            "Use execution.num_instances (number of instances) and execution.num_nodes "
+            "(total nodes) instead — num_nodes must be divisible by num_instances, so "
+            "one instance uses num_nodes // num_instances nodes."
```
piojanu, line 651: Don't you have to use `vllm_ray` when `num_nodes // num_instances` > 1?

piojanu: Also, please update the docs: https://github.com/NVIDIA-NeMo/Evaluator/blob/main/docs/libraries/nemo-evaluator-launcher/configuration/executors/slurm.md#multi-node-and-multi-instance

AWarno: Hi, no , it should work with any deployment, we have many other multinode deployments , even for vllm it does not need to be ray

AWarno: I updated the docs

PR #854, opened 2026-03-17
fix: byob readme example

laszkiewiczp:
Fixes EVAL-823

In the feedback regarding BYOB provided in https://docs.google.com/document/d/1y_rJAWLEmEFl82TUduHB0hlV0Yhxe5DptbXx9cxHEgs/edit?tab=t.0 it was mentioned that the example of BYOB usage should refer to a real example + some minor issue regarding how the installation is documented

### DIFF
--- a/docs/libraries/nemo-evaluator/extending/byob/index.md
+++ b/docs/libraries/nemo-evaluator/extending/byob/index.md
@@ -107,7 +107,7 @@ nemo-evaluator run_eval \
 ```
 
 :::{tip}
-Use `nemo-evaluator-byob my_benchmark.py --dry-run` to validate your benchmark without installing it.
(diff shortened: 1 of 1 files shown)


packages/nemo-evaluator/examples/byob/README.md
```diff
@@ -1,35 +1,89 @@
 # BYOB (Bring Your Own Benchmark)
 
-Create custom evaluation benchmarks for NeMo Evaluator in ~12 lines of Python.
+Create custom evaluation benchmarks for NeMo Evaluator with minimal Python.
```
marta-sd: nit: I think this is misleading - it's unclear that we refer to an amount of code / coding effort ("minimal Python" sounds like 3.10 to me 😂 )

laszkiewiczp: haha thats fair, I will change it

PR #855, opened 2026-03-17
fix: use normalized name in BYOB FDF evaluation entry

laszkiewiczp:
Fixes EVAL-824

The evaluation name and config type in the compiled BYOB FDF used the original benchmark name (e.g. `test-benchmark`) while the registry keyed it by the normalized form (`test_benchmark`). This mismatch causes `--eval_type byob_test_benchmark.test-benchmark` (a flag suggested in the output of BYOB) to fail with "Benchmark not found" at runtime.

Source: https://docs.google.com/document/d/1y_rJAWLEmEFl82TUduHB0hlV0Yhxe5DptbXx9cxHEgs/edit?tab=t.0

### DIFF
--- a/packages/nemo-evaluator/src/nemo_evaluator/contrib/byob/compiler.py
+++ b/packages/nemo-evaluator/src/nemo_evaluator/contrib/byob/compiler.py
@@ -127,7 +127,7 @@ def _build_fdf(
         "defaults": defaults,
         "evaluations": [
             {
-                "name": normalized_name,
+                "name": bench.name,
                 "description": f"BYOB benchmark: {bench.name}",
                 "defaults": {
                     "config": {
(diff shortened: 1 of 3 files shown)


laszkiewiczp:
@copilot There should be some extra changes made within `docs/libraries/nemo-evaluator/extending/byob/cli.md`

packages/nemo-evaluator/src/nemo_evaluator/contrib/byob/compiler.py
```diff
@@ -127,11 +127,11 @@ def _build_fdf(
         "defaults": defaults,
         "evaluations": [
             {
-                "name": bench.name,
```
marta-sd, line 130: are you sure name needs to be normalized? we often have dashes in benchmark names - this is the human-readable version of the benchmark name. I think only the `type` needs to be normalized

laszkiewiczp: Yeah I wasn't sure about how to approach this one, I will revert this change

laszkiewiczp: Done

PR #857, opened 2026-03-17
feat: add watch mode for continuous checkpoint evaluation

marta-sd:


docs/references/api/nemo-evaluator-launcher/api.md
```diff
@@ -176,6 +176,64 @@ print(f"Status: {status['status']}")
 print(f"Output directory: {status['data']['output_dir']}")
 ```
 
+## Watching Checkpoints
```
marta-sd, line 179: TODO: I forgot to update docs after taking over the task

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/container_metadata/loading.py
```diff
@@ -758,6 +758,11 @@ def extract_framework_yml(
         Tuple of (framework_yml_content, container_digest) or (None, None) if failed
     """
     container_digest = None
+    if container.endswith(".sqsh"):
```
marta-sd, line 761: This is not related to the main topic of the PR but something I found annoying while working on it

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/execdb.py
```diff
@@ -24,8 +25,13 @@
 from nemo_evaluator_launcher.common.logging_utils import logger
 
 # Configuration constants
-EXEC_DB_DIR = pathlib.Path.home() / ".nemo-evaluator" / "exec-db"
-EXEC_DB_FILE = EXEC_DB_DIR / "exec.v1.jsonl"
+EXEC_DB_FILE = os.getenv("NEMO_EVALUATOR_EXEC_DB_FILE", None)
```
marta-sd, line 28: This change is also something not needed for this task, but something I added for my convenience while working on it. TODO: add documentation

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/watcher/watchdb.py
```diff
@@ -0,0 +1,89 @@
...
+from pathlib import Path
+from typing import Any, Optional
+
+from pydantic import BaseModel, ConfigDict, Field
+
+from nemo_evaluator_launcher.common.logging_utils import logger
+
+WATCH_STATE_FILE = os.getenv(
+    "NEMO_EVALUATOR_WATCH_STATE_FILE",
+    Path.home() / ".nemo-evaluator" / "watch-state" / "watch-state.v1.jsonl",
```
agronskiy: `bug`: when the env var is set, the getenv would get `str`, not Path, and down at the calling site it at line 38 I think it will crash because `str` does not have those methods.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/ssh_utils.py
```diff
@@ -0,0 +1,195 @@
...
+        hostname: SSH hostname.
+
+    Raises:
+        RuntimeError: If rsync fails.
+    """
+    for local_source in local_sources:
+        assert local_source.is_dir()
+    remote_destination_str = f"{username}@{hostname}:{remote_target}"
+    local_sources_str = " ".join(map(str, local_sources))
+    rsync_upload_command = f"rsync -qcaz {local_sources_str} {remote_destination_str}"
```
agronskiy: `question`: on mac, oftentimes the local paths have paths w/ spaces (yeah we linux users should think about those mac ppl), would not that break? I suggest hacvin command as a list with sub-parts - this way the `.Run` below will pass them to bash properly tokenized.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/ssh_utils.py
```diff
@@ -0,0 +1,195 @@
...
+        username: SSH username.
+        hostname: SSH hostname.
+        socket: Path to the Unix socket file. No-op if None.
+
+    Raises:
+        RuntimeError: If the connection could not be closed.
+    """
+    if socket is None:
+        return
+    ssh_command = f"ssh -O exit -S {socket} {username}@{hostname}"
```
agronskiy: `question`: I would recommand to check on all the commands to be passed as lists -- shlex.split later will break if socket path on Mac contains spaces

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/ssh_utils.py
```diff
@@ -0,0 +1,195 @@
...
+    completed_process = subprocess.run(args=shlex.split(ssh_command))
```
agronskiy: `bug(?)`: the default of `stderr` param is `None`, I think this would lead to double-raise later when `stderr.decode` is called

agronskiy (approved):
Couple of small things, generally LGTM after those fixed -- unblocking with one-shot approve

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/watcher/watchdb.py
marta-sd: Good catch!

marta-sd: fixed in a9855d6f

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/ssh_utils.py
marta-sd: ...and it's inconsistent with how we open it 👍 Fixed in 6fda9686

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/ssh_utils.py
marta-sd: Actually there's no reason for us to throw an error here. I replaced with similar log.error message to the one we have when opening failed (6fda9686)

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/ssh_utils.py
marta-sd: This code was just moved from one place to another, so I assume it's functional, but better be on the safe side 👍 Done in 59dba76e

PR #859, opened 2026-03-17
feat: BYOB add output_parser parameter to judge_score()

laszkiewiczp:
Fixes EVAL-842

Allow users to plug in custom parsing logic for LLM judge responses. As for now, we were assuming JSON format or a regex search. This was another feature suggested in the feedback.

### DIFF
--- a/packages/nemo-evaluator/src/nemo_evaluator/contrib/byob/judge.py
+++ b/packages/nemo-evaluator/src/nemo_evaluator/contrib/byob/judge.py
@@ -309,7 +309,7 @@ def judge_score(
     score_mapping: Optional[Dict[str, float]] = None,
     judge_key: str = "judge",
     response_format: Optional[Dict[str, Any]] = None,
-    output_parser: Optional[Callable[[str], str]] = None,
+    judge_output_parser: Optional[Callable[[str], str]] = None,
     **template_kwargs: Any,
 ) -> dict:
     """Score a sample using an LLM judge.
@@ -336,7 +336,7 @@ def judge_score(
             decoding (e.g. ``{"type": "json_object"}``).  When set, it is
             passed to ``judge_call()`` and ``parse_grade()`` uses structured
             JSON parsing.
-        output_parser: Optional callable that receives the raw judge response
+        judge_output_parser: Optional callable that receives the raw judge response
             string and returns a grade string (e.g. ``"C"``, ``"3"``,
             ``"SAFE"``).  When provided, it replaces the default
             ``parse_grade()`` call.  The ``score_mapping`` step still runs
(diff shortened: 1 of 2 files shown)


packages/nemo-evaluator/src/nemo_evaluator/contrib/byob/judge.py
```diff
@@ -335,6 +336,11 @@ def judge_score(
             decoding (e.g. ``{"type": "json_object"}``).  When set, it is
             passed to ``judge_call()`` and ``parse_grade()`` uses structured
             JSON parsing.
+        output_parser: Optional callable that receives the raw judge response
```
marta-sd: `nit`: I think `grade_parser` or `judge_response_parser` would be a better name. we have output parsers for processing metrics

packages/nemo-evaluator/src/nemo_evaluator/contrib/byob/judge.py
```diff
@@ -395,12 +401,19 @@ def judge_score(
...
-        judge_response,
-        resolved_pattern,
-        structured=response_format is not None,
-    )
+    # Parse grade — use custom parser or default regex/JSON parsing
+    if output_parser is not None:
+        try:
+            grade = output_parser(judge_response)
+        except Exception:
+            logger.warning("Custom output parser failed", judge_key=judge_key)
```
marta-sd: I think it would be helpful if we showed the exception to the user. if we just except and move on it's difficult for the user to debug what went wrong

packages/nemo-evaluator/tests/unit_tests/byob/test_byob_judge.py
```diff
@@ -885,3 +885,74 @@ def test_custom_template_kwargs_passed_through(self, mock_call):
...
+    @patch("nemo_evaluator.contrib.byob.judge.judge_call")
+    def test_custom_output_parser_grade_mapped(
+        self, mock_call, parser_return, mapping, expected_score, expected_grade
+    ):
+        mock_call.return_value = "raw judge response"
+        sample = self._make_sample()
+        result = judge_score(
+            sample,
+            template="binary_qa",
+            output_parser=lambda resp: parser_return,
```
marta-sd: I don't think this test checks what we should be checking. I think  we should mock the judge response instead and use a real function for parsing it.

packages/nemo-evaluator/src/nemo_evaluator/contrib/byob/judge.py
laszkiewiczp: done

packages/nemo-evaluator/tests/unit_tests/byob/test_byob_judge.py
laszkiewiczp: sure, I updated the test

packages/nemo-evaluator/src/nemo_evaluator/contrib/byob/judge.py
laszkiewiczp: Improved

PR #864, opened 2026-03-18
feat: support arbitrary sbatch flags via sbatch_extra_flags

gchlebus:
## Summary

- Adds `sbatch_extra_flags` dict to the Slurm executor config, allowing users to inject any `#SBATCH` flag not natively exposed (e.g. `--switches`, `--constraint`, `--mem`, `--reservation`)
- Moves the previously hardcoded `--exclusive` into the default config as `exclusive: true`, so users can opt out by setting it to `false`
- Documents the feature in `slurm.md` with examples and a tip on `--switches=1` for multi-node deployments
- Adds 15 parametrized unit tests covering key-value flags, boolean flags, null/false skipping, ordering, and realistic multi-node scenarios

Ported from internal branch `crawlbert/EVAL-852-sbatch-extra-flags`.

## Test plan

- [x] `uv run pytest tests/unit_tests/test_slurm_executor.py::TestSbatchExtraFlags -v` — all 15 tests pass
- [x] Pre-commit (`ruff` lint + format) passes clean

### DIFF
--- a/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
+++ b/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
@@ -31,6 +31,12 @@ import yaml
 from jinja2 import Environment, FileSystemLoader
 from omegaconf import DictConfig, OmegaConf
 
+from nemo_evaluator_launcher.common.auxiliary_deployments import (
+    AuxDeploymentState,
+    build_aux_deployment_states,
+    resolve_deployment_command,
+    validate_auxiliary_deployments,
+)
 from nemo_evaluator_launcher.common.env_vars import (
     SecretsEnvResult,
     build_reexport_commands,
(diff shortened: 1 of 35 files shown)


packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
```diff
@@ -688,13 +688,17 @@ def _create_slurm_sbatch_script(
         s += "#SBATCH --gres {}\n".format(cfg.execution.gres)
     if cfg.execution.get("sbatch_comment"):
         s += "#SBATCH --comment='{}'\n".format(cfg.execution.sbatch_comment)
+    for flag_name, flag_value in cfg.execution.get("sbatch_extra_flags", {}).items():
+        if isinstance(flag_value, bool) and flag_value:
+            s += "#SBATCH --{}\n".format(flag_name)
+        elif not isinstance(flag_value, bool) and flag_value is not None:
+            s += "#SBATCH --{} {}\n".format(flag_name, flag_value)
```
agronskiy: `suggestion`: check what about values w/ spaces? in yaml, anything after `:` will be one single string and here it's not guarded. Another example: often disabling smth would require passing in yaml `key: ""` -- this will:
- flag_value is not None - correct
- it's recored without quotes like `#SBATCH --some_key ` instead of `--some_key ""`

gchlebus: Good point. Addressed it in c4a00c0cfedaa02a975df4a26d72461194c4786c.

PR #866, opened 2026-03-19
ci: Fix sso user check

chtruong814:
ci: Fix sso user check

The file name we used to check if a user is internal or not changed. Updating the workflow to pass in the file name via a Github variable so we can change it later if needed.


chtruong814:
Fast merging this to ensure CI is working as expected.

PR #873, opened 2026-03-19
feat(extra-params): export extra params

AWarno:


packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/cli/version.py
```diff
@@ -16,15 +16,22 @@
...
 
-from nemo_evaluator_launcher import __package_name__, __version__
+from nemo_evaluator_launcher import __package_name__
 from nemo_evaluator_launcher.common.logging_utils import logger
 
 
 def get_versions() -> dict:
     internal_module_name = "nemo_evaluator_launcher_internal"
-    res = {__package_name__: __version__}
+    try:
```
marta-sd: why we need this change?

PR #874, opened 2026-03-19
feat: allow duplicate task names in nel

laszkiewiczp:
## Summary

  - Add `get_unique_task_name()` pure function that generates positional directory names (`"{task_name}.{idx}"`) for evaluation tasks, enabling the same task to appear multiple times in a config (e.g., same benchmark with different temperatures)
  - Enhance `filter_tasks()` to accept both plain names (`"mmlu"` — all instances) and positional unique names (`"mmlu.0"` — one specific instance), with plain names taking priority to avoid shadowing
  - Extract `_build_task_dir_index()` in the exporter for clean directory-to-index resolution with backward compatibility for old-format directories

  ## Motivation

  Users want to run the same benchmark multiple times with different configurations (e.g., different temperatures, different prompts). Previously, task names had to be unique, which prevented this. Now a config like this works:

  ```yaml
  evaluation:
    tasks:
      - name: lm-evaluation-harness.ifeval
      - name: lm-evaluation-harness.ifeval
        nemo_evaluator_config:
          config:
            params:
              temperature: 0.7

  This creates directories lm-evaluation-harness.ifeval.0 and lm-evaluation-harness.ifeval.1, and users can filter with --tasks lm-evaluation-harness.ifeval (runs both) or --tasks lm-evaluation-harness.ifeval.0 (runs only the
  first).
  ```
  
  Fixes EVAL-56

### DIFF
--- a/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/functional.py
+++ b/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/functional.py
@@ -202,25 +202,22 @@ def _find_matching_task_indices(
     """Match each filter against the task list by plain name or unique name.
 
     A filter like ``"mmlu"`` matches every task with that plain name.
-    A filter like ``"mmlu.0"`` matches a single task by its unique name,
-    but only if no task is literally named ``"mmlu.0"`` (plain name takes
-    priority to avoid shadowing real task names).
+    A filter like ``"mmlu.0"`` matches a single task by its unique name
+    (unique names are generated as ``"{name}.{index}"`` and task names
+    cannot contain a dot-digit suffix, so there is no ambiguity).
 
     Returns ``(found_indices_sorted, unmatched_filters)``.
     """
-    by_name: defaultdict[str, list[int]] = defaultdict(list)
+    lookup: dict[str, list[int]] = defaultdict(list)
     for i, t in enumerate(tasks):
-        by_name[t.name].append(i)
-
-    by_unique_name = {get_unique_task_name(t.name, i): i for i, t in enumerate(tasks)}
+        lookup[t.name].append(i)
+        lookup[get_unique_task_name(t.name, i)].append(i)
 
     found: set[int] = set()
     unmatched_filters: list[str] = []
     for f in task_name_filters:
-        if f in by_name:
-            found.update(by_name[f])
-        elif f in by_unique_name:
-            found.add(by_unique_name[f])
+        if f in lookup:
+            found.update(lookup[f])
         else:
             unmatched_filters.append(f)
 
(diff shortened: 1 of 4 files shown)


packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/functional.py
```diff
@@ -192,42 +196,75 @@ def _validate_no_missing_values(cfg: Any, path: str = "") -> None:
...
 
 
-def filter_tasks(cfg: RunConfig, task_names: list[str]) -> RunConfig:
+def _find_matching_task_indices(
+    tasks: list, task_name_filters: list[str]
+) -> tuple[list[int], list[str]]:
+    """Match each filter against the task list by plain name or unique name.
+
+    A filter like ``"mmlu"`` matches every task with that plain name.
+    A filter like ``"mmlu.0"`` matches a single task by its unique name,
```
prokotg: In Nemo Evaluator Core we have a limitation on task name. Task invocations must be either `harness.task_name` or `task_name`. At the same time, we do not allow any other dots, so in this case I don't expect a task literally named `"task_name.index`

packages/nemo-evaluator-launcher/tests/unit_tests/exporters/test_base_exporter.py
```diff
@@ -462,32 +462,89 @@ def mock_get_jobs(invocation_id):
...
+                id="old_format_string_tasks",
+            ),
+            pytest.param(
+                [{"name": "mmlu"}, {"name": "mmlu"}],
+                {"mmlu.0": 0, "mmlu": 0, "mmlu.1": 1},
+                id="duplicate_names_first_wins_plain",
+            ),
+        ],
+    )
+    def test_build_task_dir_index(self, tasks, expected_index):
```
prokotg, line 485: What about `harness.task_name` invocation? In particular, if we have the same task in two harnesses. Let's say `lm_evaluation_harness.mmlu` , 'simple_evals.mmlu`. Would that work?

prokotg: Checked, works

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/functional.py
laszkiewiczp: Okay, in that case I will simplify this function

prokotg (approved):
Thank you!

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/api/functional.py
laszkiewiczp: Simplified

PR #876, opened 2026-03-20
feat: add generic heartbeat for periodic script execution during evaluation

AdamRajfer:
## Summary

Adds a heartbeat mechanism that runs a user-provided script periodically during evaluation jobs. Supported on all executors (SLURM, local, lepton).

### Configuration

```yaml
execution:
  heartbeat:
    script: 'ls -la artifacts/ 2>/dev/null || echo "no artifacts yet"'
    interval: 60          # seconds (default: 300)
    container: null       # optional container image
```

| Field | Default | Description |
|---|---|---|
| `script` | `null` | Command to run periodically. When `null`, heartbeat is disabled. |
| `interval` | `300` | Seconds between heartbeat executions. |
| `container` | `null` | Optional container image for the script. |

### How it works

When `script` is set, the executor starts a background loop that runs the command every `interval` seconds. The loop runs alongside the evaluation and stops automatically when the evaluation completes or fails. Script failures are logged but do not affect the evaluation.

### Changes

- `executor.py` (SLURM): heartbeat background loop in sbatch script, with optional container support via srun
- `run.template.sh` (local): heartbeat background loop in bash template
- `executor.py` (lepton): heartbeat background loop in launch script
- Config defaults added to all execution configs
- Example config: `examples/slurm_heartbeat.yaml`
- Docs: heartbeat section added to SLURM executor docs

### Removed

- `exporters/mlflow_live.py` and tests — replaced by the generic heartbeat

## Test plan

- [x] 8 heartbeat unit tests across all executors
- [x] All existing tests pass (926 passed)
- [x] Pre-commit clean
- [x] End-to-end SLURM test: heartbeat logged every 10s throughout job


packages/nemo-evaluator-launcher/tests/unit_tests/test_local_executor.py
```diff
@@ -865,3 +865,91 @@ def mock_get_task_def_side_effect(*_args, **kwargs):
...
+
+class TestHeartbeatInLocalTemplate:
+    def test_heartbeat_start_and_stop(self):
+        """Heartbeat start/stop appear when script is configured."""
+        from pathlib import Path
+
+        from jinja2 import Environment
+
+        template_text = (
+            Path(__file__).parent.parent.parent
```
marta-sd: Please check if we can use package resources instead

packages/nemo-evaluator-launcher/tests/unit_tests/test_local_executor.py
```diff
@@ -865,3 +865,91 @@ def mock_get_task_def_side_effect(*_args, **kwargs):
...
+        assert "kill" in rendered
+
+    def test_no_heartbeat_when_none(self):
+        """No heartbeat injected when script is None."""
+        from pathlib import Path
+
+        from jinja2 import Environment
+
+        template_text = (
+            Path(__file__).parent.parent.parent
```
marta-sd: Same here

packages/nemo-evaluator-launcher/tests/unit_tests/test_local_executor.py
AdamRajfer: We can. Fixed

packages/nemo-evaluator-launcher/tests/unit_tests/test_local_executor.py
AdamRajfer: We can. Fixed

marta-sd (approved):

PR #892, opened 2026-04-01
feat: parametrize exported metric name and set better defaults for W&B

marta-sd:


marta-sd:
This change turned out not needed

PR #901, opened 2026-04-09
fix: move auto-export to separate CPU-only sbatch job

AdamRajfer:
Auto-export (MLflow, wandb) was running inline in the GPU evaluation sbatch job via srun --overlap, wasting GPU time on I/O-bound uploads.

Changes:
- Generate a separate export.sbatch script with no GPU request
- Submit it via sbatch from the eval script on success
- Mount the full invocation dir + output_dir (fixes mount mismatch that caused MLflow export to fail because --job-dirs path was not accessible inside the export container)
- Support auto_export.partition config to target a CPU partition

The export job runs after the GPU job completes and releases GPUs immediately, instead of holding them during artifact uploads.


packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
```diff
@@ -1091,78 +1091,120 @@ def _generate_auto_export_section(
...
+    if isinstance(auto_export_cfg, dict) or OmegaConf.is_config(auto_export_cfg):
+        launcher_install_cmd = auto_export_cfg.get("launcher_install_cmd")
+    if not launcher_install_cmd:
+        launcher_install_cmd = "pip install nemo-evaluator-launcher[all]"
 
-    if secrets:
-        reexport_cmd = build_reexport_commands("export", secrets)
-        s += f"    {reexport_cmd}\n"
+    export_partition = None
+    if isinstance(auto_export_cfg, dict) or OmegaConf.is_config(auto_export_cfg):
```
marta-sd: we have same if in L1109, let's combine them

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
```diff
@@ -1091,78 +1091,120 @@ def _generate_auto_export_section(
     remote_task_subdir: Path,
     export_image: str = "python:3.12.7-slim",
 ) -> str:
-    """Generate simple auto-export section for sbatch script."""
+    """Generate auto-export as a separate CPU-only sbatch job.
+
+    Instead of running export inside the GPU job (wasting expensive GPU
+    time on I/O-bound MLflow uploads), we write a lightweight export
+    sbatch script and submit it with ``--dependency=afterok`` so it only
```
marta-sd: This doesn't seem to be true (and it shouldn't be - this wouldn't work with auto-resume)

marta-sd (commented):
Two small comments

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/configs/execution/slurm/default.yaml
```diff
@@ -42,6 +42,11 @@ mounts:
   auxiliary: {}  # keyed by auxiliary deployment name, e.g. auxiliary.judge: {/host/path: /container/path}
   mount_home: true
 
+auto_export:
```
marta-sd: You need to either put safe defaults (ie bunch og `.get("...", {})`) in the code if this section is missing or add it to all internal slurm configs

marta-sd: Ok, I think what we have in L1120 looks save, so we should be good

marta-sd (approved):

marta-sd (approved):

PR #909, opened 2026-04-14
fix(launcher): re-raise unrelated ModuleNotFoundError in container_metadata init

wprazuch:
## Summary
- The bare `except ModuleNotFoundError: pass` in `container_metadata/__init__.py` silently swallowed errors from transitive dependencies (pydantic, structlog, etc.) when `nemo_evaluator` was on `PYTHONPATH` as source but its deps were not installed in the active venv.
- This caused `extract_framework_yml` (and other `loading` symbols) to silently disappear from the namespace with no error, breaking downstream scripts like docs autogen.
- Now only suppresses the error when `nemo_evaluator` itself is the missing module; any other missing dependency is re-raised.

### DIFF
--- a/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/container_metadata/__init__.py
+++ b/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/common/container_metadata/__init__.py
@@ -56,6 +56,10 @@ try:
             "parse_framework_to_irs",
         ]
     )
-except ModuleNotFoundError:
+except ModuleNotFoundError as exc:
     # Allow importing this package for IR-only workflows (docs autogen, etc.)
-    pass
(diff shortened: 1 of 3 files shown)


wprazuch:
/ok to test [61172ba](https://github.com/NVIDIA-NeMo/Evaluator/pull/909/commits/61172bab1bbb7e60d78de7e0c1fec689743d63d3)

wprazuch:
/ok to test [46ceae1](https://github.com/NVIDIA-NeMo/Evaluator/pull/909/commits/46ceae120f116282caf5194596bedcf33d4ca23b)

wprazuch:
/ok to test [4d51bee](https://github.com/NVIDIA-NeMo/Evaluator/pull/909/commits/4d51beedb4d180d3efbf0a97670d1ee50b66afc0)

AWarno (approved):

PR #916, opened 2026-04-17
fix: force base-10 in bash _walltime_to_seconds

agronskiy:
Reported here in [slack](https://nvidia.slack.com/archives/C08DT9UC2V6/p1776384358707919): 
- I (we :claude:) think `max_walltime` was being silently ignored on job chains whose previous run landed on a zero-padded minute/second like `02:08:45` -- apparently `bash` interprets `08`/`09` as octal inside `$((...))` the arithmetic errors out, `_prev_elapsed_seconds` stays empty, and the accumulated walltime stays stuck at `00:00:00` forever, so the chain keeps resuming past the limit. 
- solution -- p[refixing each field with `10#` forces decimal parsing and restores the guard. Regression test runs the generated bash function against the offending inputs.

What checked
- addad a test  parametrized bash regression test covers `02:08:45`, `04:09:06` and observed it failing -- then applied the fix

### DIFF
--- a/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
+++ b/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
@@ -1671,7 +1671,7 @@ _walltime_to_seconds() {{
         seconds=${{BASH_REMATCH[1]}}
     fi
 
-    echo $((days * 86400 + hours * 3600 + minutes * 60 + seconds))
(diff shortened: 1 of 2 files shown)


prokotg:
/ok to test [1da24d3](https://github.com/NVIDIA-NeMo/Evaluator/pull/916/commits/1da24d3008c94045ebbe78c381e19635e5997cc0)

prokotg (approved):

agronskiy:
/ok to test [857ab03](https://github.com/NVIDIA-NeMo/Evaluator/commit/857ab03fbe93e85fb2086de1fe5fd258fd7eb191)

PR #933, opened 2026-04-23
fix(docs): use dynamic version from package_info in conf.py

ko3n1g:
<details><summary>Claude summary</summary>

Backports the `conf.py` patch from `main` to the `r0.1.0` release branch.

**Changes:**
- Inserts `../src` into `sys.path` so `nemo_eval` is importable at conf-load time
- Imports `__version__` from `nemo_eval.package_info` and uses it for `release` instead of the hardcoded `"0.1.0"` string
- Adds a `setup()` Sphinx hook that generates `project.json` (already declared in `html_extra_path`) with the current package version at build time

</details>

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -20,13 +20,17 @@
 import os
 import sys
 
+sys.path.insert(0, os.path.abspath("../src"))
+
+from nemo_eval.package_info import __version__
+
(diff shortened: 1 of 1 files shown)


ko3n1g:
Closing — wrong fix. r0.1.0 already has the correct `../versions1.json` path.

PR #945, opened 2026-04-28
feat: [EVAL-878] allow custom HTTP headers in payload_modifier

wprazuch:
## Summary

Adds three optional fields to `PayloadParamsModifierInterceptor.Params` — `headers_to_add`, `headers_to_remove`, `headers_to_rename` — mirroring the existing body-`params_*` surface. Unblocks NMP inference-gateway integration (Linear: [EVAL-878](https://linear.app/nvidia/issue/EVAL-878)), where eval clients need to inject headers like `X-NMP-Principal-Id: service:evaluator` on outgoing chat-completions requests.

- Header names are matched case-insensitively per RFC 7230 §3.2.
- Hop-by-hop headers (`Host`, `Content-Length`, `Connection`, `Transfer-Encoding`) are dropped at init with a warning; `Authorization` is intentionally allowed through so callers can override auth for inference-gateway deployments.
- Headers flow to upstream via the existing `EndpointInterceptor`, which already forwards `ar.r.headers` to `requests.request(...)`. No new request-flow code paths.

### Example config

```yaml
target:
  api_endpoint:
    adapter_config:
      interceptors:
        - name: payload_modifier
          config:
            headers_to_add:
              X-NMP-Principal-Id: "service:evaluator"
              X-Inference-Priority: "batch"
        - name: endpoint
```

## Test plan

- [x] 14/14 unit tests pass in `tests/unit_tests/adapters/interceptors/test_payload_modifier.py` (7 new + 7 pre-existing)
- [x] Full adapter test suite green (340/340)
- [x] `pre-commit run` clean
- [x] **End-to-end with real vLLM (Qwen3-1.7B):** spun up `AdapterServerProcess` standalone with `payload_modifier` injecting `X-Inference-Priority: batch` + `X-NMP-Principal-Id: service:evaluator`, posted a chat-completions request through it via a header-capturing sniffer proxy. Both injected headers were observed at the upstream side with original `Authorization` preserved; vLLM returned a real completion. (Companion change in nemo-evaluator-next coming via separate MR.)

### DIFF
--- a/packages/nemo-evaluator/src/nemo_evaluator/adapters/interceptors/payload_modifier_interceptor.py
+++ b/packages/nemo-evaluator/src/nemo_evaluator/adapters/interceptors/payload_modifier_interceptor.py
@@ -66,7 +66,7 @@ class PayloadParamsModifierInterceptor(RequestInterceptor):
                 "Dictionary of HTTP headers to add to the upstream request. "
                 "Existing headers with the same name (case-insensitive) are overridden. "
                 "Hop-by-hop headers (Host, Content-Length, Connection, Transfer-Encoding) "
-                "are silently dropped."
+                "are dropped with a warning."
             ),
         )
         headers_to_remove: Optional[List[str]] = Field(
--- a/packages/nemo-evaluator/src/nemo_evaluator/adapters/interceptors/payload_modifier_interceptor.py
+++ b/packages/nemo-evaluator/src/nemo_evaluator/adapters/interceptors/payload_modifier_interceptor.py
@@ -30,14 +30,22 @@ from nemo_evaluator.adapters.types import (
 )
 from nemo_evaluator.logging import BaseLoggingParams, get_logger
 
+# Transport-level headers that must not be set by user config — they are
+# managed by the HTTP layer (Werkzeug / requests) and overriding them breaks
+# message framing or routing. Authorization is intentionally NOT included so
+# users can override auth for inference-gateway style deployments.
+_HOP_BY_HOP_HEADERS: frozenset[str] = frozenset(
+    {"host", "content-length", "connection", "transfer-encoding"}
+)
+
 
 @register_for_adapter(
     name="payload_modifier",
-    description="Modifies request payload by removing, adding, and renaming parameters",
+    description="Modifies request payload by removing, adding, and renaming parameters and headers",
 )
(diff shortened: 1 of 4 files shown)


packages/nemo-evaluator/src/nemo_evaluator/adapters/interceptors/payload_modifier_interceptor.py
```diff
@@ -52,10 +60,36 @@ class Params(BaseLoggingParams):
             default=None,
             description="Dictionary mapping old parameter names to new names",
         )
+        headers_to_add: Optional[Dict[str, str]] = Field(
```
marta-sd, line 63: Please also expose these params in the legacy config similarly to what we do with `params_to_*` ([here](https://github.com/NVIDIA-NeMo/Evaluator/blob/cdbecdd9bddda845b309345fb64bbb8cfcb296dd/packages/nemo-evaluator/src/nemo_evaluator/adapters/adapter_config.py#L121))

packages/nemo-evaluator/src/nemo_evaluator/adapters/interceptors/payload_modifier_interceptor.py
```diff
@@ -52,10 +60,36 @@ class Params(BaseLoggingParams):
...
+            default=None,
+            description=(
+                "Dictionary of HTTP headers to add to the upstream request. "
+                "Existing headers with the same name (case-insensitive) are overridden. "
+                "Hop-by-hop headers (Host, Content-Length, Connection, Transfer-Encoding) "
+                "are silently dropped."
```
marta-sd:
```suggestion
                "are dropped with a warning."
```

wprazuch:
/ok to test [7da3237](https://github.com/NVIDIA-NeMo/Evaluator/pull/945/commits/7da32371b4cd8b91cdfb83e8b30cd4148d020035)

packages/nemo-evaluator/src/nemo_evaluator/adapters/interceptors/payload_modifier_interceptor.py
wprazuch: Fixed

marta-sd (approved):

PR #950, opened 2026-04-29
feat: add shared metric contract for scorer functions

SandyChapman:
Expose MetricInput -> MetricResult types and adapt decorated scorers via to_metric() so Evaluator OSS scorers can share a runtime contract with platform integrations while preserving BYOB scorer compatibility.


src/nemo_evaluator/environments/custom.py
```diff
@@ -235,37 +259,110 @@ async def verify(self, response: str, expected: str, sandbox: Sandbox | None = N
...
+                response=response,
+                metadata=meta,
+            )
+            result = await metric.compute_scores(metric_input)
+            return _metric_result_to_verify_result(
+                metric=metric,
+                result=result,
+                benchmark_name=self._defn.name,
+                response=response,
+            )
```
SandyChapman: This is just illustrative of how the `verify` func could use the `Metric` version of the `scorer`.

src/nemo_evaluator/environments/custom.py
```diff
@@ -59,21 +76,25 @@ def score(sample: ScorerInput) -> dict:
...
 
+ConfigT = TypeVar("ConfigT", bound=Mapping[str, object] | BaseModel)
+ConfigModelT = TypeVar("ConfigModelT", bound=BaseModel)
+
+
 @dataclass
 class BenchmarkDefinition:
     name: str
-    dataset: str | Callable[[], list[dict]]
+    dataset: str | Callable[..., list[dict[str, Any]]]
```
SandyChapman, line 97: There's a bit of diff noise in the PR as I address typechecking errors reported by `ty` and `pyright`.

src/nemo_evaluator/environments/custom.py
```diff
@@ -300,13 +397,133 @@ def __init__(self, num_examples: int | None = None):
...
+def scorer(
+    fn: None = None,
+    *,
+    metric_type: None = None,
+    outputs: None = None,
+    config_schema: None = None,
+) -> Callable[[ScorerCallable[ConfigT]], ScorerCallable[ConfigT]]: ...
+
+
+def scorer(
```
SandyChapman, line 508: The adjustments to the scorer decorator provide a couple additional features:
1. the ability to specify a schema (via a Pydantic BaseModel) that allows validation of inputted config objects.
2. static type safety of the Config type as well (such that the `to_metric` func will return a `ScorerFunctionMetric` with the generic parameter matching the passed config.
3. structured definition of outputs is required for supporting the `to_metric` function call as `score_names` is a required part of the `Metric` protocol.
4. `metric_type` is an optional label for the metric which is also needed to refer to it (for instance) via an API call or in the DB. By default it's generated if not specified, but we provide the ability to manually specify it in case the code gets moved or a module renamed (as the generated name uses the package name).

src/nemo_evaluator/scoring/types.py
```diff
@@ -17,14 +17,17 @@
...
 if TYPE_CHECKING:
     from nemo_evaluator.sandbox.base import Sandbox
 
 
+ConfigT = TypeVar("ConfigT")
+
+
 @dataclass
-class ScorerInput:
+class ScorerInput(Generic[ConfigT]):
```
SandyChapman, line 30: Genericize ScoreInput to allow specifying a strongly typed config object. `dict` is still accepted.

src/nemo_evaluator/environments/custom.py
wprazuch: `metric_type` to become required since it can introduce bugs

src/nemo_evaluator/scoring/metric.py
```diff
@@ -0,0 +1,395 @@
...
+        return value
+
+
+class MetricOutputSpec(BaseModel):
+    """Declared metric outputs."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    scores: list[ScoreOutputSpec] = Field(min_length=1)
+    annotations: list[AnnotationOutputSpec] = Field(default_factory=list)
```
wprazuch: todo: check if we can have one type

PR #953, opened 2026-04-30
feat(byob): add completions_logprob endpoint and extend scorers/datasets   

kanishks-23:
BYOB support for logprob-based multiple-choice evaluation and extends dataset loading/scoring utilities needed by the Sovereign benchmark suite.

- Added completions_logprob as a supported endpoint type
  - Pydantic validation now accepts target.api_endpoint.type: completions_logprob
  - CLI --model_type completions_logprob is allowed
  - Endpoint health checks treat completions_logprob like a /v1/completions endpoint
- Added BYOB logprob evaluation flow
  - Uses /v1/completions with max_tokens=0, echo=true, and logprobs=1
  - Scores candidate continuations using returned token logprobs
  - Supports multiple-choice ranking via one request per candidate answer
  - Supports nested choice fields such as choices.text
- Added/updated BYOB scorers
  - multiple_choice_acc: returns acc, acc_norm, and acc_greedy
  - mcq_letter_extract: supports A-J options and handles empty/None responses safely
  - Added task-oriented scorers for GSM8K-style numeric answers, yes/no tasks, chrF, and ROUGE
- Extended Hugging Face dataset URI support.
  - Parses extra query params beyond split
  - Supports trust_remote_code=true
  - Supports row filtering via filter_field/filter_value, including multiple filters with suffixes
  - This allows BYOB benchmarks to consume datasets where language is stored as a row field instead of a HF config

### DIFF
--- a/docs/libraries/nemo-evaluator/extending/byob/cli.md
+++ b/docs/libraries/nemo-evaluator/extending/byob/cli.md
@@ -97,6 +97,19 @@ Use `nemo-evaluator-byob --list` to see the exact `eval_type` for each installed
 benchmark. This avoids guessing the normalized name.
 :::
 
+For logprob-based multiple-choice benchmarks, use a completions endpoint that
+supports `echo` and `logprobs`:
+
+```bash
+nemo-evaluator run_eval \
+  --eval_type byob_<normalized_name>.<normalized_name> \
+  --model_url http://localhost:8000 \
+  --model_id my-model \
+  --model_type completions_logprob \
+  --output_dir ./results \
+  --api_key_name API_KEY
+```
+
 ## See Also
 
 - {ref}`byob` -- BYOB overview and quickstart
--- a/packages/nemo-evaluator/src/nemo_evaluator/contrib/byob/eval_logic.py
+++ b/packages/nemo-evaluator/src/nemo_evaluator/contrib/byob/eval_logic.py
@@ -16,6 +16,7 @@
 """Shared BYOB evaluation logic for both subprocess and native modes."""
 
 import importlib
+import random
 import sys
 import threading
 from concurrent.futures import ThreadPoolExecutor, as_completed
@@ -133,8 +134,9 @@ def import_benchmark(benchmark_module: str, benchmark_name: str) -> BenchmarkDef
 class EvalStrategy(Protocol):
     """Protocol for evaluation strategies.
 
-    Different evaluation modes (standard, judge, multi-turn, agentic)
-    implement this protocol to define how each sample is evaluated.
+    Different evaluation modes (standard, judge, multi-turn, agentic,
+    multiple-choice loglikelihood) implement this protocol to define how
+    each sample is evaluated.
     """
 
     def evaluate_sample(
@@ -142,9 +144,10 @@ class EvalStrategy(Protocol):
         idx: int,
         row: Dict,
         bench: BenchmarkDefinition,
-        model_call_fn: Callable[[str, str], str],
+        model_call_fn: Callable[..., Any],
         endpoint_type: str,
         request_timeout: Optional[float] = None,
+        fewshot_prefix: str = "",
     ) -> Tuple[Optional[Dict], Optional[SampleResult]]:
         """Evaluate a single sample.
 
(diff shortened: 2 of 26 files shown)


docs/libraries/nemo-evaluator/extending/byob/cli.md
```diff
@@ -97,6 +97,19 @@ Use `nemo-evaluator-byob --list` to see the exact `eval_type` for each installed
 benchmark. This avoids guessing the normalized name.
 :::
 
+For logprob-based multiple-choice benchmarks, use a completions endpoint that
+supports `echo` and `logprobs`:
+
+```bash
+nemo-evaluator run_eval \
+  --eval_type byob_<normalized_name>.<normalized_name> \
+  --model_url http://localhost:8000/v1/completions \
```
wprazuch: localhost might be confusing - usually it is the AdapterServer Proxy url, would be better to mock a production endpoint syntax

wprazuch:
@kanishks-23 Please amend sign-offs to the commits pushed and then launch pipeline with [following syntax](https://github.com/NVIDIA-NeMo/Evaluator/pull/876#issuecomment-4112279400) to make sure tests pass

docs/libraries/nemo-evaluator/extending/byob/cli.md
kanishks-23: @wprazuch In docs I see https://docs.nvidia.com/nemo/evaluator/latest/libraries/nemo-evaluator/extending/byob/cli.html# `--model_url http://localhost:8000` at many places. So should I change it everywhere to something like `https://integrate.api.nvidia.com/v1/chat/completions`?

kanishks-23: For now changed it to `http://localhost:8000` for consistency. We can modify it to https://integrate.api.nvidia.com/v1/chat/completions if required.

wprazuch:
/ok to test [bb06b86](https://github.com/NVIDIA-NeMo/Evaluator/pull/953/commits/bb06b86e099fe3cd92ed69ba7d039f58ea364e18)

wprazuch:
/ok to test [45e9484](https://github.com/NVIDIA-NeMo/Evaluator/pull/953/commits/45e9484d67e93d1de448aa0586d9f8760897c8e2)

docs/libraries/nemo-evaluator/extending/byob/cli.md
wprazuch: Oh, that's right. Let's leave it for consistency and we can change on a separate occassion

packages/nemo-evaluator/src/nemo_evaluator/contrib/byob/eval_logic.py
```diff
@@ -358,10 +367,330 @@ def evaluate_sample(
...
+            )
+
+        if fewshot_prefix:
+            prompt = fewshot_prefix + prompt
+
+        # Resolve choices: per-row field takes precedence over static list
+        choices: Optional[List[str]] = None
+        if bench.choices_field:
+            choices = _resolve_choices_from_row(row, bench.choices_field)
+        if not choices and bench.choices:
```
wprazuch, line 457: Let's log a warning that this fallback happens

wprazuch:
/ok to test [9556770](https://github.com/NVIDIA-NeMo/Evaluator/pull/953/commits/955677001b81c2f39a355af1bc4556c55d676ea0)

prokotg (approved):
Looks good. Overall, architecturally logprobs should be realized elsewhere than `endpoint_type` because `endpoint_type` only indicates the endpoint itself not behaviour (just like images, video, and audio can be realized through both chat and completions). I think logprobs would be better realized through a `Strategy` or other mechanism.  Similarily `choices`, `num_fewshots` and others should fall behind extras. Approved contingent we agree architecture is most likely going to change in the future.

wprazuch:
/ok to test [9556770](https://github.com/NVIDIA-NeMo/Evaluator/pull/953/commits/955677001b81c2f39a355af1bc4556c55d676ea0)

PR #993, opened 2026-05-07
feat(byob): add explicit few-shot dataset support

kanishks-23:
## Summary

Adds explicit BYOB few-shot controls for benchmarks where a split-only rewrite is not enough.

- Add `fewshot_dataset` to `@benchmark` so tasks can provide an exact few-shot dataset URI/path, including filters, configs, `data_files`, and other query params.
- Add `fewshot_prefix` to prepend static text before rendered few-shot examples.
- Fix `--num-fewshot 0` so it overrides non-zero benchmark defaults and enables true 0-shot validation.
- Save BYOB predictions by default from generated command templates.
- Expose `fewshot_dataset` and `fewshot_prefix` in generated FDF `config.params.extra.dataset`.
- Update docs and tests for precedence, fallback, prefix rendering, and explicit 0-shot behavior.

## Why

`fewshot_split` only works for simple datasets where changing `split` is enough. Some datasets also require `filter_field` / `filter_value`, `data_files`, configs, or other URI parameters. Reconstructing those generically is fragile and can produce mixed-language or wrong-source few-shot examples.

This change lets benchmark authors provide the exact few-shot source when needed while preserving existing `fewshot_split` behavior.

## Test Plan

```bash
cd packages/nemo-evaluator
uv run python -m pytest \
  tests/unit_tests/byob/test_byob_decorators.py::TestBenchmarkLogprobFields \
  tests/unit_tests/byob/test_byob_eval_logic.py::TestFewshotPrefix \
  tests/unit_tests/byob/test_byob_eval_logic.py::TestBuildFewshotExamples \
  tests/unit_tests/byob/test_byob_compiler.py::TestBuildFdfHelper::test_fdf_groups_dataset_config_under_extra_dataset \
  tests/unit_tests/byob/test_byob_runner.py::TestFewshotOverride

### DIFF
--- a/docs/libraries/nemo-evaluator/extending/byob/benchmark-decorator.md
+++ b/docs/libraries/nemo-evaluator/extending/byob/benchmark-decorator.md
@@ -30,7 +30,9 @@ def check(sample: ScorerInput) -> dict:
 | `choices` | `list[str]` | `None` | Static candidate continuations for `endpoint_type="completions_logprob"` |
 | `choices_field` | `str` | `None` | Dataset field containing per-row candidate continuations for `endpoint_type="completions_logprob"`; dotted paths such as `choices.text` are supported |
 | `num_fewshot` | `int` | `0` | Number of few-shot examples to prepend to each prompt |
-| `fewshot_split` | `str` | `None` | Optional split to sample few-shot examples from |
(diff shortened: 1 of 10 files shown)


wprazuch:
/ok to test [55af0dc](https://github.com/NVIDIA-NeMo/Evaluator/pull/993/commits/55af0dc1382c80464bf2862d9b01dd742d589d71)

wprazuch (approved):

PR #1010, opened 2026-05-13
ci: publish dev wheels to rolling GitHub pre-release on every dev/0.3.0 push

wprazuch:
## What this does

On every push to `dev/0.3.0` **or on-demand via `workflow_dispatch`** (Actions → Run workflow → pick any branch):

1. Stamps a **PEP 440 pre-release version** — `0.3.0.devN` where N = total commit count on that branch
2. Builds the wheel with `uv`
3. Uploads it to the rolling GitHub pre-release tag **`v0.3.0-dev`**, replacing the previous wheel

The `workflow_dispatch` trigger has an optional **"Upload wheel"** checkbox (default on) — uncheck to do a build-only dry run without publishing.

### Install latest dev build

```bash
pip install https://github.com/NVIDIA-NeMo/Evaluator/releases/download/v0.3.0-dev/nemo_evaluator-0.3.0.dev<N>-py3-none-any.whl
```

## Versioning scheme

| Branch / trigger | Version | Where |
|---|---|---|
| `dev/0.3.0` push (automatic) | `0.3.0.devN` | GitHub pre-release ← this PR |
| Any branch (manual dispatch) | `0.3.0.devN` | GitHub pre-release ← this PR |
| `main` | `0.3.0` | PyPI (existing workflow) |

## Why GitHub releases (not TestPyPI)

No `TWINE_USERNAME` / `TWINE_PASSWORD` secrets are configured in this repo. GitHub Releases require only the built-in `GITHUB_TOKEN`. When TestPyPI credentials are added, flip `no-publish` in `build-test-publish-wheel.yml` to also fire on `dev/0.3.0` and retire this workflow.

## Notes
- Only one wheel is kept in the pre-release at a time (old wheel deleted before uploading new one).
- The `v0.3.0-dev` tag is a pre-release — won't appear as "latest" on the releases page.
- Version stamp is ephemeral — applied in CI only, not committed back to the branch.

PR #1011, opened 2026-05-13
fix(dev-wheel): stamp pyproject.toml so wheel is named 0.3.0.devN

wprazuch:
## Problem

PR #1010 introduced the dev-wheel workflow but the wheel built as `nemo_evaluator-0.3.0-...` instead of `nemo_evaluator-0.3.0.devN-...`.

Root cause: `uv build` reads the version from `pyproject.toml`, not from `package_info.py` or `__init__.py`. The stamp step was patching the latter two but not `pyproject.toml`.

## Fix

Add one `sed` line to the stamp step:
```diff
+ sed -i "s/^version = \"0\.3\.0\"/version = \"0.3.0.dev${N}\"" pyproject.toml
  sed -i "s/^PRE_RELEASE = .*/PRE_RELEASE = \".dev${N}\"" src/nemo_evaluator/package_info.py
  sed -i "s/^__version__ = .*/__version__ = \"0.3.0.dev${N}\"" src/nemo_evaluator/__init__.py
```

After this, the wheel will be correctly named `nemo_evaluator-0.3.0.dev24-py3-none-any.whl` and installable as:
```bash
pip install https://github.com/NVIDIA-NeMo/Evaluator/releases/download/v0.3.0-dev/nemo_evaluator-0.3.0.dev24-py3-none-any.whl
```

### DIFF
--- a/.github/workflows/dev-wheel.yml
+++ b/.github/workflows/dev-wheel.yml
@@ -56,8 +56,10 @@ jobs:
         id: version
         run: |
           N=$(git rev-list --count HEAD)
+          # pyproject.toml is what uv build reads for the wheel filename
+          sed -i "s/^version = \"0\.3\.0\"/version = \"0.3.0.dev${N}\"/" pyproject.toml
+          # Keep package_info.py and __init__.py in sync for runtime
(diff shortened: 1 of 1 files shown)


wprazuch:
/ok to test [bac36c9](https://github.com/NVIDIA-NeMo/Evaluator/pull/1011/commits/bac36c90c201dfb8d693b035a508e00802b03435)

PR #1020, opened 2026-05-21
feat: add publish command to exporting NEL results to HF leaderboards

marta-sd:
This PR adds CLI command that allows to publish results generated with NEL to a HuggingFace benchmark leaderboard. It uploads NEL artifacts (required + logs) to a newly-created HF space:
Screenshot from 2026-05-22 12-10-49

and creates a PR to model repo with the leaderboard entry:
Screenshot from 2026-05-22 12-30-21


packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/exporters/utils.py
```diff
@@ -448,7 +448,7 @@ def copy_local_artifacts(
                             f"Failed to copy required artifact {artifact} from {original_file} to {copy_file}: {e}"
                         )
                     else:
-                        logger.warning(
+                        logger.debug(
```
daniilsorokin, line 451: Nit: do we need to touch this?

daniilsorokin (approved):

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/exporters/utils.py
marta-sd: We don't *need* but it's very annoying - our list of optional artifacts doesn't make much sense (see [here](https://github.com/NVIDIA-NeMo/Evaluator/blob/1c9488b6b6476dad929391c8e6d60444a710c3fc/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/exporters/utils.py#L47)) and this message is printed with every export / copy. Having this warning might be confusing for the users, especially when they are trying to publish scores to official leaderboard ("I'm missing some file, maybe my run is not fully correct and I shouldn't publish?")

My change doesn't address the root cause, ie incorrect artifacts' list, but I don't want to touch in this PR

daniilsorokin: Ah, okay, makes sense then!

daniilsorokin (approved):

PR #1021, opened 2026-05-21
fix: disable xtrace before setting up env vars holding secrets

marta-sd:


mjmikulski (approved):
LGTM

Flow in the executor.py is nontrivial - we should plan a task to write tests to detect leakage in the future.

PR #1035, opened 2026-05-27
fix: update installation instructions for NeMo FW multimodal evals

marta-sd:
The old instructions won't work starting from 26.04.01. The new instructions work both for 26.04.00 and newer containers.

I have also fixed the mentioned version of nemo fw container needed for multi-modal evals - 26.02 was not fully functional.


athitten (approved):
LG, thank you @marta-sd !

PR #1037, opened 2026-05-28
ci: moving skills to base folder for ci ingestion

ritaneves:
moving the skills to top /skills folder for ci pipeline


agronskiy (approved):
LG!

PR #1046, opened 2026-06-02
feat: Legacy container

kslowik:
Add `ContainerEnvironment` to run legacy eval-factory containers as opaque boxes via the `container://image#task` URI scheme.  NEL pre-renders a v1-format `run_config.yaml`, mounts it at `/config/run_config.yaml`, and dispatches the container's `nemo-evaluator run_eval` entrypoint either via local `docker run` or in-job `srun --container-image` (under `--submit`).

Includes:
- `ContainerEnvironment` + module-level helpers (`build_legacy_run_config`, `parse_legacy_results`) so `slurm_gen` can pre-render at submit time
- SLURM dispatch path for `container://` benchmarks + `service.pre_cmd` field for user-supplied service-side scripts
- `cluster.container_env` propagation on local-cluster path (was a no-op before — only SLURM / docker dispatch paths honored it)
- `ContainerSolverConfig` schema aligned with codebase
- Guard against nel-next adapter proxies on ContainerSolverConfig
- SSH executor: ship nested sidecar dirs on `--submit`
- Tests: legacy container BC dispatch, SSH executor, env-var propagation, CLI
- Walkthrough Section 14: generic `container://` schema example

### DIFF
--- a/src/nemo_evaluator/cli/eval.py
+++ b/src/nemo_evaluator/cli/eval.py
@@ -745,7 +745,7 @@ def eval_report(results_dir, run_id, fmt, output, all_formats, with_results):
     Accepts a local RESULTS_DIR or --run-id to fetch results from a remote
     SLURM cluster automatically.
     """
-    from nemo_evaluator.reports.eval import RENDERERS, build_table, load_bundles
+    from nemo_evaluator.reports.eval import RENDERERS, build_table, load_bundles, materialize_legacy_bundles
 
     if results_dir is None and run_id is None:
         raise click.ClickException("Specify a RESULTS_DIR or --run-id.")
@@ -757,6 +757,12 @@ def eval_report(results_dir, run_id, fmt, output, all_formats, with_results):
         results, cleanup_dir = _fetch_remote_results(run_id, include_results=with_results)
 
     try:
+        # Convert any raw legacy-harness results.yml into NEL bundles
+        # before scanning for eval-*.json.  No-op when no legacy results
+        # are present (native runs already emit bundles directly).
+        for bundle_path in materialize_legacy_bundles(results):
+            click.echo(f"  Materialized legacy bundle: {bundle_path}")
+
         bundle_files = sorted(results.rglob("eval-*.json"))
 
         if not bundle_files:
--- a/src/nemo_evaluator/environments/container.py
+++ b/src/nemo_evaluator/environments/container.py
@@ -41,8 +41,12 @@ Usage in NEL config:
 
 from __future__ import annotations
 
+import asyncio
 import json
 import logging
+import os
+import re
+import shutil
 import tempfile
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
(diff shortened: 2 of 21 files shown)


src/nemo_evaluator/cli/eval.py
```diff
@@ -84,6 +85,7 @@ def eval_cmd():
 @click.option("--submit", is_flag=True, help="Submit to cluster via SSH")
 @click.option("--background", is_flag=True, help="Run locally in background")
 @click.option("--resume", is_flag=True, help="Resume a partially completed evaluation")
+@click.option("--env-file", type=click.Path(exists=True), help="Load environment variables before parsing config")
```
Glorf: is this relevant? If so, why not dotenv?

src/nemo_evaluator/environments/container.py
```diff
@@ -136,55 +364,63 @@ async def run_batch(self, solver: Any = None, config: dict[str, Any] | None = No
...
-            except _aio.TimeoutError:
+                _stdout, stderr = await asyncio.wait_for(proc.communicate(), timeout=self._timeout)
+            except asyncio.TimeoutError:
                 proc.kill()
                 await proc.wait()
                 stderr = b"timeout"
 
             if proc.returncode != 0:
-                logger.error("Container failed (exit %d): %s", proc.returncode, (stderr or b"").decode()[:2000])
+                logger.error("Container failed (exit %d): %s", proc.returncode, (stderr or b"").decode()[-5000:])
```
Glorf: ? Basically either way - why do we strip that?

src/nemo_evaluator/environments/container.py
```diff
@@ -235,49 +476,76 @@ def _build_docker_cmd(self, results_dir: Path, config_path: Path) -> list[str]:
...
+        for mount in mounts:
             cmd.extend(["-v", mount])
-        return cmd
+        for k, v in env_vars.items():
+            cmd.extend(["-e", f"{k}={v}"])
+        cmd.append(self._image)
+        cmd.extend(inner_command)
+        return cmd, None
+
+    def _build_srun_cmd(
```
Glorf, line 487: qq: is this tested both for docker (local runs) and pyxis?

src/nemo_evaluator/executors/ssh.py
```diff
@@ -176,9 +183,25 @@ def copy_to_remote(
     else:
         _ssh(target, f"mkdir -p {shlex.quote(remote_dir)}/logs", timeout=15.0)
 
+    if local_base is not None:
```
Glorf, line 186: Why this is needed?

src/nemo_evaluator/orchestration/model_server.py
```diff
@@ -42,6 +42,7 @@ class DeployConfig:
     extra_env: dict[str, str] = field(default_factory=dict)
     cluster_env: dict[str, str] = field(default_factory=dict)
     extra_args: list[str] = field(default_factory=list)
+    container_mounts: list[str] = field(default_factory=list)
```
Glorf: I think we had container_mounts available before? 🤔

src/nemo_evaluator/orchestration/orchestrator.py
```diff
@@ -881,8 +883,16 @@ def _build_batch_config(
 
     protocol = getattr(svc, "protocol", "chat_completions") if svc else "chat_completions"
     cfg["endpoint_type"] = solver_cfg.endpoint_type or _NEL_PROTOCOL_TO_LEGACY_TYPE.get(protocol, "chat")
-    if solver_cfg.params:
-        cfg["params"] = solver_cfg.params
+    if solver_cfg.adapter_config:
```
Glorf, line 885: this modifies logic external to legacy containers, why do we need that?

Glorf (commented):
Briefly those changes, could you consider giving a diff (https://github.com/NVIDIA-NeMo/Evaluator/pull/1046) one more pass through claude to make sure only the relevant changes are kept. Then should be ok to accept

src/nemo_evaluator/cli/eval.py
kslowik: Dropped — agree this duplicates dotenv. Reverted in the latest force-push.

src/nemo_evaluator/environments/container.py
kslowik: Dropped — reverted to the original `[:2000]`.

src/nemo_evaluator/environments/container.py
kslowik: Yes — `test_environments_extended.py::TestContainerEnvironment` covers the Docker dispatch path; `test_orchestration/test_slurm_legacy_container.py` covers the `srun --container-image` (Pyxis) path.

src/nemo_evaluator/executors/ssh.py
kslowik: Needed for the SLURM dispatch of `container://` benchmarks — the BC bridge writes the per-benchmark `run_config.yaml` under `{staging}/{safe_name}/run_config.yaml`, and the sbatch script bind-mounts that nested path. Without `local_base`, scp flattens it and the mount target does not exist on the remote. See `slurm_executor.py:432-435` for the call site. Backwards-compatible — default `None` keeps the original flattening behavior, so non-container paths are unaffected.

src/nemo_evaluator/orchestration/model_server.py
kslowik: Dropped

src/nemo_evaluator/orchestration/orchestrator.py
kslowik: All gated on `isinstance(solver_cfg, ContainerSolverConfig)` — non-container paths are unchanged. Required because (a) `adapter_config` / `env_vars` / `mounts` need to flow into the container's `run_config.yaml` and `docker run` flags (otherwise the container is unconfigurable from nel-next), and (b) the container hosts its own internal adapter, so fronting it with NEL's proxy would double-wrap every request.

Glorf:
lgtm, merge when ready :)

PR #1066, opened 2026-06-09
fix: recover ECS Fargate sandbox tunnels and capture task-death cause

Glorf:
## Summary
Makes the ECS Fargate sandbox recover from SSH-tunnel deaths instead of failing `POST /exec` with connection-refused, and captures *why* a tunnel died so failures are diagnosable. Addresses two issues from a GPT-OSS-120B SWE-bench Verified run where 122 rollouts were misclassified as infra failures.

## Root cause
**Reconnect became terminal.** On reconnect, the sandbox reused the same `ssh -R <port>` while the sandbox sshd still held the reverse listener from the dead tunnel (kept up to the keepalive window). The bind failed with `remote port forwarding failed (address in use)`, which `SshTunnel.open()` treated as fatal, so `POST /exec` failed with connection-refused. Under high concurrency a burst of failing execs each spawned its own competing reconnect — the `bind [127.0.0.1]:<port>: Address in use` pileup across task streams.

**Tunnel-death cause was unknowable.** Post-mortems could not explain the original death: ECS stopped-task metadata expires (~1h) and CloudWatch retained no stop reason / OOM / container exit cause.

## Changes
- `SshTunnel.open()` retries remote-forward bind failures (previously connection-level errors only).
- Serialize reconnects behind a per-sandbox `asyncio.Lock` so a burst triggers a single reconnect, not a storm of competing `ssh -R` binds.
- Reclaim the stale reverse listen port via `fuser -k` (adds `psmisc` to the sidecar image) before rebinding, so recovery doesn't wait out the keepalive window.
- Capture ECS `lastStatus` / `stopCode` / `stoppedReason` / container `exitCode` at failure time; if the task is gone, abort reconnect instead of churning on a dead container.
- Move the blocking health-wait / cleanup off the event loop.

## Test plan
- [x] `pytest tests/test_sandbox/test_ecs_fargate.py` -> 84 passed; ruff check + format clean.
- [x] New regression tests confirmed to fail against pre-fix code.
- [x] **Live end-to-end on real ECS Fargate** (us-east-2): start -> exec -> simulated tunnel death -> reverse-port reclaim via real `fuser` (`rc=0`) -> reconnect -> exec; then out-of-band `stop_task` -> captured `stopCode=UserInitiated; stoppedReason=...; container(main=STOPPED exit=137)` -> reconnect aborted; task torn down.

## Notes
- `psmisc` changes the sidecar container definition, so the task-definition content hash changes once (one-time re-register; self-healing). Sidecars built before this change lack `fuser` -> graceful fallback to the bind-retry path.
- SSH keepalive tolerance is intentionally left unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resilience of the SSH tunnel used by the ECS Fargate sandbox, with smarter retry handling for tunnel startup and connection failures.
  * Added reliable recovery for “dead tunnel” exec sessions, including prevention of reconnect storms during concurrent requests.
  * Enhanced diagnostics capture and best-effort reclamation of stale reverse-port listeners to restore connectivity faster.
* **Tests**
  * Expanded coverage for reconnect/recovery scenarios, stale port reclamation, sidecar command composition, and failure-fast behavior when tasks are already stopped or confirmed dead.

### DIFF
--- a/src/nemo_evaluator/sandbox/ecs_fargate.py
+++ b/src/nemo_evaluator/sandbox/ecs_fargate.py
@@ -428,6 +428,19 @@ def download_secret_to_string(secret_arn: str, region: str | None = None) -> str
 
 # ── SSH tunnel ───────────────────────────────────────────────────────
 
+# ssh stderr substrings worth a tunnel-open retry: connection-level blips, plus a
+# reverse-forward bind failing because the sandbox sshd still holds the port from a
+# prior dead tunnel (frees once reclaimed/reaped). Matched case-insensitively.
+_SSH_RETRYABLE_STDERR: tuple[str, ...] = (
+    "connection refused",
+    "connection timed out",
+    "no route to host",
+    "connection reset",
+    "remote port forwarding failed",
+    "address already in use",
+    "cannot listen to port",
+)
+
 
 class SshTunnel:
     """Manages an ``ssh -N`` subprocess with ``-L`` / ``-R`` tunnels."""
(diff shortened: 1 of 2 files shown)


Glorf:
/ok to test https://github.com/NVIDIA-NeMo/Evaluator/commit/88d22ba37df9ff5a34d863adb0d03ab9ada8c3d6

piojanu (approved):

PR #1068, opened 2026-06-12
feat: guard chat-template kwargs on /v1/completions

teelinsan:
## Summary

Adds a model-agnostic guard that catches a silent, score-corrupting misconfiguration: chat-template controls (`chat_template_kwargs`, `chat_template`) set on the **text** completions endpoint (`/v1/completions`), where they have no effect.

## Background

`chat_template_kwargs` / `chat_template` only do something on a **chat** endpoint (`/v1/chat/completions`), where the server renders the prompt from `messages` and applies the chat template. The **text** completions endpoint (`/v1/completions`) receives an already-rendered `prompt` string and never applies a chat template, so any chat-template field is **inert**.

Concretely, a user trying to disable reasoning via the adapter on a text-completions endpoint:

```yaml
target:
  api_endpoint:
    adapter_config:
      params_to_add:
        chat_template_kwargs:
          enable_thinking: false   # never reaches the model on /v1/completions
```

…gets a run that _looks_ correctly configured but generates with reasoning still **ON**, and the reported scores are misleading. The correct place to set the toggle is wherever the prompt is rendered client-side (the exact key is model-specific), or to use a chat endpoint so the server applies the chat template.

This is easy to get wrong and impossible to notice from the results, so we want to fail loudly instead of scoring silently.

## What this PR does

A new shared module `nemo_evaluator/completions_guard.py` implements one tiered, model-agnostic policy (it never needs to know a model's specific reasoning key):

| Condition on `/v1/completions`                                                                                                  | Action                                                                          |
| ------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------- |
| **Boolean toggle** (`True` **or** `False`) anywhere inside `chat_template_kwargs` (e.g. `enable_thinking: true`, `thinking: false`) | **hard error** (`ValueError`) — fail fast                                       |
| Any other chat-template field present (a bare `chat_template` string, non-boolean `chat_template_kwargs` values like `thinking_mode: "chat"`) | strip it + log a **WARNING**                                                    |
| `NEMO_EVALUATOR_ALLOW_COMPLETIONS_CHAT_TEMPLATE=1` set                                                                           | opt out of the hard error: strip **every** chat-template field (booleans included) + **WARNING** |

Notes on the design:

* **Any boolean — not just `False` — is a hard error.** Model key names are not standardized (`enable_thinking`, `disable_thinking`, `thinking`, `high_reasoning`, …), so the value alone cannot distinguish "on" from "off" (`disable_thinking: true` is an *off* expressed as `True`). Treating any boolean as a deliberate toggle is the model-agnostic signal that someone tried to control generation in a place where it is silently ignored.
* Integers `0`/`1` are **not** treated as booleans (`isinstance(..., bool)`), so a value like `some_count: 0` degrades to strip+warn rather than failing.
* String-encoded values (e.g. `thinking_mode: "chat"`) and bare `chat_template` strings degrade to **strip+warn**, never silent and never fatal.
* Detection and stripping are **recursive** — nested `dict`/`list` structures (e.g. `extra_body.chat_template_kwargs`) are handled, not just top-level keys.
* `/v1/chat/completions` is explicitly excluded (it legitimately uses these fields).

### Where it is enforced (two chokepoints)

1. **`adapters/interceptors/endpoint.py`** — per request, on the outgoing body just before it is sent. This is the universal path and covers in-container adapters regardless of how the config was assembled.
2. **`environments/container.py`** (`build_legacy_run_config`) — when the legacy `adapter_config` is assembled into `run_config.yaml`, so the error surfaces early, before the container starts.

## Backward compatibility

The default is now **fail-fast on boolean toggles**. A config that inherits an _enabling_ `chat_template_kwargs` (e.g. `thinking: true`) from a shared default layer via config merge will **hard-error** on `/v1/completions` instead of running in the wrong mode. To keep such a run going when the toggle is known to be safely inert on this endpoint, set `NEMO_EVALUATOR_ALLOW_COMPLETIONS_CHAT_TEMPLATE=1` (strips + warns instead of failing). Non-boolean chat-template fields and bare `chat_template` strings continue to strip+warn by default.

> Note: this replaces the opt-in `NEMO_EVALUATOR_STRICT_COMPLETIONS` from earlier revisions of this PR with the opt-out `NEMO_EVALUATOR_ALLOW_COMPLETIONS_CHAT_TEMPLATE`, per review feedback (fail on any boolean, allow an explicit escape hatch).

## Testing

`tests/test_completions_guard.py` (49 tests) covers:

* path / endpoint-type / allow-env predicates,
* hard-fail on any boolean toggle, top-level and nested,
* allow-env downgrade to strip+warn,
* strip+warn for non-boolean values and bare `chat_template` strings,
* integer-vs-boolean handling (`0`/`1` are not toggles),
* recursive nested stripping,
* chat-endpoint and clean-body no-ops,
* `adapter_config` normalization (immutability of the input, `None` passthrough),
* `build_legacy_run_config` integration (hard-fail on boolean, allow-env strip, chat-endpoint passthrough).

```bash
ruff check src/ tests/
ruff format --check src/ tests/
pytest tests/test_completions_guard.py -q   # 49 passed
```

## Files changed

* `src/nemo_evaluator/completions_guard.py` (new) — shared tiered policy
* `tests/test_completions_guard.py` (new)
* `src/nemo_evaluator/adapters/interceptors/endpoint.py` — enforce on request body
* `src/nemo_evaluator/environments/container.py` — enforce when building `adapter_config`

## Release Notes

* **New Features**
  * Text completion requests now include automatic validation of request body parameters.
  * Added environment variable `NEMO_EVALUATOR_ALLOW_COMPLETIONS_CHAT_TEMPLATE` to control request validation behavior.

* **Bug Fixes**
  * Improved adapter configuration handling during environment setup to prevent incompatible parameter propagation.

* **Tests**
  * Added comprehensive test coverage for request validation and configuration normalization.

### DIFF
--- a/src/nemo_evaluator/completions_guard.py
+++ b/src/nemo_evaluator/completions_guard.py
@@ -1,0 +1,226 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Guard against chat-template controls leaking onto text completions.
+
+``chat_template_kwargs`` / ``chat_template`` only take effect on a chat
+endpoint, where the server renders the prompt from ``messages``. The text
+completions endpoint (``/v1/completions``) consumes an already-rendered
+``prompt`` and never applies a chat template, so these fields are inert: a
+toggle such as ``chat_template_kwargs.enable_thinking=false`` injected through
+the adapter never reaches the model. That produces a run that *looks* correctly
+configured but silently scores with the wrong generation mode.
+
+The policy is tiered and model-agnostic — it never needs to know a model's
+specific reasoning key:
+
+* **present (any value)** -> strip the field and log a warning. Removes the
+  silent inert field so it can never quietly change (or fail to change) a run.
+* **disable intent** (a boolean ``False`` anywhere inside
+  ``chat_template_kwargs``, e.g. ``enable_thinking: false`` / ``thinking:
+  false``) -> hard error. Someone is trying to turn model behavior *off* in a
+  place that is silently ignored, so the behavior would stay *on* and scores
+  would be misleading. Failing fast is safer than a misconfigured run.
+* ``NEMO_EVALUATOR_STRICT_COMPLETIONS=1`` -> escalate *any* present field
+  (even enabling ones) to a hard error, for CI.
+
+A boolean ``False`` is the model-agnostic signal for "turn this off": it does
+not matter whether the key is ``enable_thinking``, ``thinking``, ``reasoning``,
+etc. String-encoded disables (e.g. ``thinking_mode: "chat"``) are not treated
+as a hard error and degrade to strip+warn — still safe, never silent.
+"""
+
+from __future__ import annotations
+
+import logging
+import os
+from typing import Any
+
+logger = logging.getLogger(__name__)
+
+CHAT_TEMPLATE_ONLY_FIELDS: frozenset[str] = frozenset({"chat_template_kwargs", "chat_template"})
+
+STRICT_ENV_VAR = "NEMO_EVALUATOR_STRICT_COMPLETIONS"
+
+_TRUTHY = frozenset({"1", "true", "yes", "on"})
+
+
+def strict_completions_enabled() -> bool:
+    """True when ``NEMO_EVALUATOR_STRICT_COMPLETIONS`` requests hard failures."""
+    return os.environ.get(STRICT_ENV_VAR, "").strip().lower() in _TRUTHY
+
+
+def is_text_completions_path(path: str) -> bool:
+    """True for the OpenAI *text* completions endpoint (``/v1/completions``).
+
+    The chat endpoint ``/v1/chat/completions`` also ends in ``/completions``,
+    so it is explicitly excluded.
+    """
+    clean = path.rstrip("/")
+    return clean.endswith("/completions") and not clean.endswith("/chat/completions")
+
+
+def is_text_completions_endpoint_type(endpoint_type: str | None) -> bool:
+    """True for the legacy ``endpoint_type: completions`` selector."""
+    return (endpoint_type or "").strip().lower() == "completions"
+
+
+def _has_boolean_false(obj: Any) -> bool:
+    """True when a boolean ``False`` appears anywhere inside *obj*.
+
+    ``is False`` (not ``== False``) so the integer ``0`` is not mistaken for a
+    disable toggle.
+    """
+    if obj is False:
+        return True
+    if isinstance(obj, dict):
+        return any(_has_boolean_false(v) for v in obj.values())
+    if isinstance(obj, list):
+        return any(_has_boolean_false(item) for item in obj)
+    return False
+
+
+def _scan_chat_template_fields(obj: Any, prefix: str = "") -> tuple[list[str], list[str]]:
+    """Find chat-template-only keys inside *obj*.
+
+    Returns ``(locations, disabling_locations)`` as dotted paths.
+    ``disabling_locations`` are the ``chat_template_kwargs`` occurrences that
+    carry a boolean ``False`` (a disable intent).
+    """
+    locations: list[str] = []
+    disabling: list[str] = []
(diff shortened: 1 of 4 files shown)


src/nemo_evaluator/completions_guard.py
```diff
@@ -0,0 +1,226 @@
...
+import os
+from typing import Any
+
+logger = logging.getLogger(__name__)
+
+CHAT_TEMPLATE_ONLY_FIELDS: frozenset[str] = frozenset({"chat_template_kwargs", "chat_template"})
+
+STRICT_ENV_VAR = "NEMO_EVALUATOR_STRICT_COMPLETIONS"
+
+_TRUTHY = frozenset({"1", "true", "yes", "on"})
```
prokotg, line 63: `nit:` I like this, but unless applied to all variables, it might create an expectation it works across envvars

src/nemo_evaluator/completions_guard.py
```diff
@@ -0,0 +1,226 @@
...
+            key: _strip_chat_template_fields(value)
+            for key, value in obj.items()
+            if key not in CHAT_TEMPLATE_ONLY_FIELDS
+        }
+    if isinstance(obj, list):
+        return [_strip_chat_template_fields(item) for item in obj]
+    return obj
+
+
+def _build_message(
```
prokotg, line 142: beautifully clear, appreciate it!

src/nemo_evaluator/completions_guard.py
```diff
@@ -0,0 +1,226 @@
...
+    clean = path.rstrip("/")
+    return clean.endswith("/completions") and not clean.endswith("/chat/completions")
+
+
+def is_text_completions_endpoint_type(endpoint_type: str | None) -> bool:
+    """True for the legacy ``endpoint_type: completions`` selector."""
+    return (endpoint_type or "").strip().lower() == "completions"
+
+
+def _has_boolean_false(obj: Any) -> bool:
```
prokotg: We can stay with finding Falses, but I imagine, that "enabling" something that does not exists could be renderedinvalid as well. Any boolean IMO qualifies.

src/nemo_evaluator/completions_guard.py
```diff
@@ -0,0 +1,226 @@
...
+    if not locations:
+        return
+    strict = strict_completions_enabled() if strict is None else strict
+    message = _build_message(locations, context=f"a request to {path!r}", disabling=disabling, strict=strict)
+    if disabling or strict:
+        raise ValueError(message)
+    logger.warning(message)
+    for name in [key for key in body if key in CHAT_TEMPLATE_ONLY_FIELDS]:
+        body.pop(name, None)
+
```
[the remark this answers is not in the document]
prokotg: This comment is partially valid. In `_strip_chat_template_fields` you also look for `CHAT_TEMPLATE_ONLY_FIELDS` nesed inside body. It might be the case the you wll emit a warning about some chat template only fields that are nested inside body but will have no effect. For the main use case though, this code still works

tests/test_completions_guard.py
```diff
@@ -0,0 +1,219 @@
...
+    body = {"messages": [], "chat_template_kwargs": {"enable_thinking": False}}
+    enforce_text_completions_body(path, body)
+    assert body["chat_template_kwargs"] == {"enable_thinking": False}
+
+
+def test_body_noop_on_clean_request() -> None:
+    body = {"prompt": "rendered", "max_tokens": 16}
+    enforce_text_completions_body("/v1/completions", body)
+    assert body == {"prompt": "rendered", "max_tokens": 16}
+
```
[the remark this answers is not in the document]
prokotg: I don't think we want that, but Rabbit is right that code is written for nested dict as well. Can be ignored, or scanning be improved

tests/test_completions_guard.py
```diff
@@ -0,0 +1,219 @@
...
+)
+def test_strict_completions_enabled(monkeypatch: pytest.MonkeyPatch, value: str, expected: bool) -> None:
+    monkeypatch.setenv(STRICT_ENV_VAR, value)
+    assert strict_completions_enabled() is expected
+
+
+# ── request body (#2, in-container send path) ────────────────────────────────
+
+
+@pytest.mark.parametrize("disable", [{"enable_thinking": False}, {"thinking": False}, {"nested": {"x": False}}])
```
prokotg: I'll repeat my comment from the source itself -  I think we might have a different "enabling" field in the future that can affect behaviour too as they are not standardized. Imagine "disable_thinking", "high_resoning", etc

prokotg (approved):
Looks good! Please consider if we should only look for "False" values or for any boolean. There is also a discrepancy between how chat template fields are scanned and removed but IIUC the original intent still works. 

Thank yo!

src/nemo_evaluator/completions_guard.py
teelinsan: Solved in new revision

tests/test_completions_guard.py
teelinsan: Solved in new revision

prokotg (approved):

prokotg:
/ok to test [3304e50](https://github.com/NVIDIA-NeMo/Evaluator/pull/1068/commits/3304e501cbc26a9fbbed989ed4bc564589997dca)

PR #1074, opened 2026-06-16
fix: [FEP-1087] make reasoning_replay native mode work on vLLM >=0.19

iguehring-nvidia:
## Summary
- `reasoning_replay` `mode: native` silently dropped cross-turn chain-of-thought on vLLM ≥0.19: it wrote only the `reasoning_content` field, but vLLM's chat input parser (`chat_utils._parse_chat_message_content`) keys on the `reasoning` field and ignores `reasoning_content`, so the CoT was dropped before `apply_chat_template` → no cross-turn reasoning → ~2× longer agentic rollouts (Nemotron-Nano SWE-bench Verified: median 131 vs 69 turns, same accuracy).
- `intercept_response` likewise cached only from `reasoning_content`, which vLLM ≥0.19 never sets on responses.

## Fix
- `native`/`both` re-injection now writes **both** `reasoning` and `reasoning_content` (backends disagree on which key they read on input — vLLM ≥0.19 vs SGLang/older vLLM).
- Response caching falls back to `reasoning` when `reasoning_content` is absent, so `reasoning_replay` is self-sufficient without the companion `reasoning` normalizer for the vLLM field-rename case.
- Request-side skip-guard considers both fields. Default `think_tags` mode is unchanged.

## Testing
- `pytest tests/ -m "not network"` — full offline suite green (1653 passed); 4 new + 2 updated reasoning_replay tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reasoning replay compatibility across backends by reading, caching, and re-injecting reasoning from whichever field is provided.
  * In native/both modes, the interceptor now preserves both `reasoning` and `reasoning_content`, and wraps cached hidden chain with the configured think tags.
  * Request-side replay is now correctly skipped when reasoning is already present.

* **Tests**
  * Expanded interceptor tests to cover native and both-mode behaviors, including round-trip verification and fallback/caching scenarios across different reasoning-field representations.

### DIFF
--- a/src/nemo_evaluator/adapters/interceptors/reasoning_replay.py
+++ b/src/nemo_evaluator/adapters/interceptors/reasoning_replay.py
@@ -14,13 +14,18 @@
 # limitations under the License.
 """Cross-turn ``reasoning_content`` recovery.
 
-Caches ``reasoning_content`` from upstream responses (keyed by
-``tool_call.id``, with a content-hash fallback for text-only turns) and
+Caches the model's reasoning from upstream responses — the ``reasoning_content``
+field, falling back to ``reasoning`` (vLLM >=0.19 emits the latter) — keyed by
+``tool_call.id`` with a content-hash fallback for text-only turns, and
 re-injects it into matching assistant messages on the next outbound
-request.  Works around agent SDKs (OpenHands, pi-ai) that drop
-``reasoning_content`` from chat-completions assistant messages on
-replay, which would otherwise lose the model's chain-of-thought across
-turns.
+request.  Works around agent SDKs (OpenHands, pi-ai) that drop reasoning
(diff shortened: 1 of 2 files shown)


src/nemo_evaluator/adapters/interceptors/reasoning_replay.py
```diff
@@ -143,6 +148,8 @@ async def intercept_response(self, resp: AdapterResponse) -> AdapterResponse:
                 if not isinstance(msg, dict):
                     continue
                 rc = msg.get("reasoning_content")
```
AWarno: should we  also not follow here convention:  msg.get("reasoning_content") or msg.get("reasoning")?

AWarno (approved):

PR #1043, opened 2026-06-01
chore: Sync from gitlab main @ 986d0650

marta-sd:
## Summary by CodeRabbit

* **New Features**
  * HTTP request/response dump interceptor with status metrics.
  * PinchBench: multi-source support with per-source caching and manifest ordering.
  * Harbor solver: optional skill prompt injection.
  * OpenClaw: Tavily web-search support and multi-session/multi-turn PinchBench runs.
  * Added Terminal-Bench 2.1 playbook and Harbor registry override support.
  * CLI: legacy-result materialization for export/report commands.

* **Bug Fixes**
  * Authorization headers redacted from interceptor context.
  * Improved ECS Fargate async exec and task-definition caching.
  * More robust proxy shutdown and safer progress token handling.
  * Cleaner timestamped logging for certain backends.

PR #1002, opened 2026-05-08
chore: catch dev/0.3.0 up with gitlab main @ 20830504 (bulk sync)

wprazuch:
## Summary

Bulk catch-up sync of public-partition files from gitlab/main into public `dev/0.3.0`. Companion to four parallel feature PRs (Terraform stack, reasoning_replay interceptor, OracleSolver + TB-Hard, MLflow exporter overhaul).

After this PR + the four companions, `dev/0.3.0` is caught up with the work accumulated on the internal branch since the original 0.3.0 publication (2026-03-06).

## What's in this PR

64 files changed, +3420 / -460. Spans:

- **Orchestration / SLURM hardening**: `chmod a+rx` eval outputs, threaded `cluster.container_mounts`, new sharding-strategy + shuffling, `pipeline_parallel_size` to vLLM, sglang `--model-path` flag, sandbox refactor for multi-OutsideEndpoints (env-var keyed reverse tunnels).
- **Observability**: `scrape_metrics` Prometheus sidecar (FEP-831), engine_metrics.jsonl wiring.
- **Adapters / proxy**: hop-by-hop header drop, exception-class logging, endpoint timeout init log.
- **PinchBench scoring stability**: judge retry budget (5 attempts / 4–45s backoff), `grading_weights`, transcript enrichment via `_load_transcript` / `_summarize_transcript`.
- **Resume correctness**: ModelResponse restoration from `inference_log` cache, ArtifactCollector cached-step replay (EVAL-1248 / EVAL-1255 — refs stripped).
- **Quality-gate / paired-regression**: McNemar test, flip report, verdicts, clustered SE, sign test, permutation test, NVFP4 policy YAMLs.
- **Routine**: Python floor `>=3.11` → `>=3.12` (per consultation), pyproject `[tool.uv] conflicts` table for `skills` extra, copyright touches.

## Excluded (per partition contract)

- **Internal-only paths** that never travel to public: `internal/`, `.gitlab/`, `.gitlab-ci.yml`, `deploy/gitlab-ci-eval.yml`, `examples/configs/vpr/**`, `sonar-project.properties`, `docs/agents/**`, `CONTEXT.md` / `CONTEXT-MAP.md` / subpackage `CONTEXT.md`, the 5 stray `examples/configs/0[89]*` / `12a_*` configs.
[...]

PR #1077, opened 2026-06-17
fix: repair broken container builds

wprazuch:
## Problem

Two bulk syncs from the internal branch introduced `COPY` lines in Dockerfiles for files that were never committed to the public repo, breaking `container-build-base` and `container-build-components` (skills) on every push since `0.3.1`.

## Fix

**`docker/Dockerfile.base`** — remove premature `COPY docker/observability/otel-collector.yaml`. The otelcol-contrib binary is still installed, but the config file is not copied. The code that uses it hasn't been published yet; COPY and file go in together when that feature lands.

**`docker/skills-uv-overrides.txt`** — add the missing file. `Dockerfile.skills` COPYs it for `pip install --overrides`; the skills container is a live public feature.

**`pyproject.toml`** — bump to `0.3.2`. `0.3.1` never produced a working image; this is the first buildable release.

## Root cause

Partial bulk syncs (#1043, #1002) cherry-picked Dockerfile changes from the internal branch without including the files being COPYed. With the new GL→GH mirror setup this won't recur.

## Summary by CodeRabbit

* **Chores**
  * Version bumped to 0.3.2.


iguehring-nvidia (approved):

pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "nemo-evaluator"
-version = "0.3.1"
```
marta-sd, line 7: this should be dynamic and derived from `package_info.py`. See legacy version: https://github.com/NVIDIA-NeMo/Evaluator/blob/main-legacy/packages/nemo-evaluator-launcher/pyproject.toml

src/nemo_evaluator/__init__.py
```diff
@@ -18,7 +18,7 @@
 # can safely back-reference ``nemo_evaluator.__version__`` without triggering
 # a circular import while ``__init__.py`` is partially loaded. ``package_info``
 # remains the canonical source for the FW-CI-templates wheel patcher.
-__version__ = "0.3.1"
```
marta-sd, line 21: We should import it from `package_info`. See legacy code: https://github.com/NVIDIA-NeMo/Evaluator/blob/main-legacy/packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/__init__.py

pyproject.toml
wprazuch: Let me do that as follow up PR just to unblock the CI, reverted version change

src/nemo_evaluator/__init__.py
wprazuch: Let me do that as follow up PR just to unblock the CI, reverted version change

wprazuch:
/ok to test [cf0369f](https://github.com/NVIDIA-NeMo/Evaluator/pull/1077/commits/cf0369fdf9f41297306a3a9045330d23165cc1b1)

PR #1078, opened 2026-06-17
chore: bump version to 0.3.2

wprazuch:
Patch release to follow #1077.

`0.3.1` never produced a working container image (broken `Dockerfile.base` and `Dockerfile.skills` from two partial bulk syncs). `0.3.2` is the first buildable release.

**Merge order**: #1077 first, then this.


wprazuch:
Folding into #1077.

PR #1085, opened 2026-06-22
fix: [EVAL-1480] chunk Terminus 2 tmux send-keys by UTF-8 byte length

mlazuka:
Harbor's Terminus 2 agent drives the container through a tmux session and splits long keystrokes so each `tmux send-keys` command stays under tmux's maximum command length. The split measured length in Unicode code points (`len()`), but tmux enforces the limit in bytes. Multibyte UTF-8 payloads therefore passed the character-count check while exceeding the byte limit, so tmux rejected the command with "command too long". Terminus raised that as a runtime error, the agent crashed before finishing, verification was skipped, and the trial scored 0 (surfaced misleadingly as a model error).

Make the chunking byte-accurate: measure command and chunk lengths in UTF-8 bytes for both the whole-payload check and the per-key splitting (including the binary search that splits a single oversized key), and raise when the command prefix leaves no room for any payload. The fix is applied to Harbor's TmuxSession at solver setup, host-side, so it covers every sandbox backend and any agent that drives a tmux session.

Add unit tests (ASCII, quote-heavy, multibyte, many-key, and a real Lean proof fixture) plus an opt-in real-tmux smoke test behind a new `tmux` marker.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the Harbor solver’s tmux keystroke submission to chunk by UTF-8 **byte** length (matching tmux’s limit), including correct handling of quoted and multibyte payloads.

* **Tests**
  * Added dedicated coverage to verify the byte-based chunking behavior, idempotent installation, and end-to-end correctness (including real tmux smoke execution when available).
  * Added a new Lean4 proof fixture for the solver regression scenario.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -108,6 +108,7 @@ markers = [
     "integration: integration tests that wire real components together",
     "slow: tests that take > 30s (deselect via: -m 'not slow')",
     "e2e: end-to-end smoke tests requiring infrastructure",
+    "tmux: tests that require a real tmux binary",
 ]
 filterwarnings = [
     "ignore::DeprecationWarning",
--- a/src/nemo_evaluator/solvers/harbor.py
+++ b/src/nemo_evaluator/solvers/harbor.py
@@ -23,6 +23,7 @@ import json
 import logging
 import os
 import random
+import shlex
 import tempfile
 import time
 from pathlib import Path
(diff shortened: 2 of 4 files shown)


piojanu (approved):
LGTM

piojanu:
This commit d52dc6686a1ca2376cb18d9d63d16c13bcd54d31 lacks sign-off. Please amend it with the `git commit --ammend --no-edit --signoff` and force push it again.

PR #1087, opened 2026-06-23
fix: relabel cap-truncated finish_reason to length and flag empty generations

teelinsan:
## Summary
- Add a `finish_reason` adapter interceptor that reconciles `finish_reason` with token usage: when a generation reaches its token cap but the server still reports `finish_reason="stop"`, relabel it to `"length"`.
- Warn on empty generations (empty assistant content with no tool calls) so silent failures surface in the run logs.
- Enable it by default for every proxied run.

## Problem
Some serving stacks report `finish_reason="stop"` even when a generation exhausts its entire token budget. For reasoning models this happens when the budget is consumed inside the thinking phase, leaving empty content and no terminating token. The `"stop"` label then hides a truncated, unusable generation from every downstream consumer metrics, reports, and especially LLM judges, which can score an empty or degenerate output as correct. Because the mislabel originates in the serving stack, `finish_reason` on its own is not a reliable truncation signal.

## Solution
A small interceptor that sits in the proxy (`adapters/interceptors/finish_reason.py`). It runs in two steps: once on the outgoing request, and once on the response that comes back.

- On the request, it remembers the token limit that was asked for (`max_tokens` or `max_completion_tokens`).
- On the response, it checks how many tokens were actually generated. If the model used up the entire limit but the server still labeled it `finish_reason="stop"` (meaning "finished normally"), that label is wrong, because the output was really cut off, so it changes the label to `"length"`. It also logs a warning whenever a response comes back empty.

The interceptor is turned on automatically for every proxied run (registered and added in `orchestration/orchestrator._start_proxy`). It's added after `payload_modifier` so it sees the real token limit, and it won't be added twice if you've already configured it. It also runs after the `reasoning` step, so the "empty output" check looks at the final answer text rather than the model's thinking text.

## Testing
- `tests/test_adapters/test_finish_reason_interceptor.py`: relabel on cap hit (including `max_completion_tokens` and reasoning-token detection), no relabel under the cap or without a cap, non-`stop` reasons left untouched, non-dict bodies ignored, and empty-generation warning fires (but not for tool-call or non-empty responses).
- `tests/test_orchestration/test_orchestrator.py`: interceptor injected by default and ordered last, ordered after `payload_modifier`, and not duplicated when explicitly configured.
- `ruff check` / `ruff format` clean; adapter and orchestrator suites pass with no regressions.

## Summary by CodeRabbit

* **New Features**
  * Added automatic finish-reason fixup in proxy flows: when configured token caps are reached, eligible `stop` results are relabeled to `length`.
  * Added empty-generation detection with warnings for whitespace-only responses (tool calls considered non-empty).
  * Introduced a `finish_reason_fixup` proxy setting (default `true`) to enable/disable this behavior.
* **Tests**
  * Added coverage for token-cap relabeling, reasoning-token truncation detection, multi-choice behavior, and warning suppression/conditions.


src/nemo_evaluator/orchestration/orchestrator.py
```diff
@@ -799,6 +799,12 @@ def _start_proxy(
         if overrides:
             interceptor_specs.append({"name": "payload_modifier", "config": {"params_to_add": overrides}})
 
+    # Always-on safety net: relabel cap-truncated `stop` responses to `length`
+    # and warn on empty generations. Appended last so its request phase observes
+    # any token cap injected by payload_modifier above.
+    if not any(spec["name"] == "finish_reason" for spec in interceptor_specs):
```
marta-sd: I agree that this should be enabled by default, but I think we should provide an opt-out mechanisms. This interceptor masks the original server behavior which might be unwanted for some users.

marta-sd (commented):
LGTM, added one small comment. Please also make sure it's tested with vllm, sglang and build.nvidia.com.

src/nemo_evaluator/orchestration/orchestrator.py
teelinsan: Added opt-out mechanism

teelinsan:
> LGTM, added one small comment. Please also make sure it's tested with vllm, sglang and build.nvidia.com.

Verified end-to-end against all three backends with the `finish_reason` interceptor active (GSM8K, Qwen3-4B on an RTX 5090 for the self-hosted ones):

| Backend | below cap | at cap | interceptor |
|---|---|---|---|
| build.nvidia.com | `stop` | `length` | no-op ✅ |
| vLLM | `stop` | `length` | no-op ✅ |
| sglang | `stop` | `length` | no-op ✅ |

All three report `finish_reason` correctly (`length` at the token cap, `stop` below it).

teelinsan:
/ok to test [fde3b49](https://github.com/NVIDIA-NeMo/Evaluator/pull/1087/commits/fde3b4976869721814cf3c996bb5d98323f6c034)

teelinsan:
@marta-sd thanks for the review. I’ve implemented all requested changes. I'm waiting for final approval to merge 🙌

marta-sd (approved):

PR #1088, opened 2026-06-24
fix(trajectory report): wire alignment, default set up

AWarno:
Summary:
- Compare trajectory wire metrics against the selected last session per problem/repeat.
- Keep all-session retry traffic as separate context.
- Preserve non-success model traffic error_type/error_message/error_code/error_body.
- Make output.trajectories.enrich default true, with enrich: false as opt-out.

Tests:
- uv run --extra dev python -m pytest tests/test_config/test_config_schema.py tests/test_engine/test_model_traffic_capture.py tests/test_reports/test_trajectories_report.py -q

## Summary by CodeRabbit

* **New Features**
  * Trajectory enrichment now defaults to enabled (disable via `enrich=false`).
  * Model traffic capture now includes tool-call payloads, reasoning, and message content by default (opt-out available).
  * Trajectory reports expand token and wire-call breakdowns for both selected “last” sessions and all sessions, including retry/session summaries.

* **Bug Fixes**
  * Non-success responses now retain a compact, capped error summary in logs and reports, with new error fields (`error_message`, `error_body`, `error_code`).
  * Wire-session alignment now consistently uses the latest relevant session.

* **Tests**
  * Updated/added coverage for enrichment/capture defaults, error-summary forwarding, last-session selection, and report output expectations.

### DIFF
--- a/src/nemo_evaluator/reports/schemas.py
+++ b/src/nemo_evaluator/reports/schemas.py
@@ -1,6 +1,6 @@
 # SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
-"""Pydantic schemas for trajectories.jsonl and model_traffic.jsonl rows.
+"""Pydantic schemas for trajectory report coverage checks.
 
 Required fields (no default) mark the minimum a writer must emit.
 Optional fields (``= None``) are expected but may be absent in older logs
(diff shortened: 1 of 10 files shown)


daniilsorokin (approved):

src/nemo_evaluator/reports/schemas.py
```diff
@@ -137,6 +137,9 @@ class ModelTrafficRow(BaseModel):
     usage: ModelTrafficUsage | None = None
     token_provenance: str | None = None
     error_type: str | None = None
+    error_message: str | None = None
+    error_body: str | None = None
+    error_code: str | None = None
```
piojanu: **`ModelTrafficRow` is defined but unused** — the error fields added here won't be read by anything.

This hunk adds `error_message` / `error_body` / `error_code` to `ModelTrafficRow`, but the class isn't referenced anywhere outside its own definition:

```
$ grep -rn "ModelTrafficRow" src/ tests/
src/nemo_evaluator/reports/schemas.py:118:class ModelTrafficRow(BaseModel):
```

`trajectories.py` only imports `AgentStep` and `TrajectoryRow` for `fields_as_coverage_paths`; the model_traffic rows are consumed as raw dicts (`record.get(...)`) for token sums / dedup / non-200 / step-wire alignment, and these error fields are produced in `observability/model_traffic.py` + `engine/model_traffic_log.py` and read by dict access — all bypassing this schema. So the model is effectively dead, and the new fields won't surface through it (e.g. there's no per-wire-call field-coverage track the way there is for steps/trials).

Two options:
- **Wire it in** — validate/parse `model_traffic.jsonl` rows through `ModelTrafficRow`, e.g. a `per_wire_call_missing` coverage section parallel to `per_agent_step_missing` / `per_trial_missing`. That also gives these error fields a coverage signal.
- **Drop it** — if there's no near-term plan to consume it, remove `ModelTrafficRow` so `schemas.py` reflects what's actually validated.

Not blocking — just flagging so the schema doesn't read as authoritative when nothing consumes it.

piojanu (approved):
One comment, otherwise LGTM!

Also, Claude prepared [this HTML doc.](https://github.com/user-attachments/files/29300806/pr-1071-1088-explainer.html) explaining the value trajectory report/enrichment is providing to us. It's very nice!

PR #1090, opened 2026-06-24
fix(adapters): add Qwen-safe turn_counter reminder modes

fsiino-nvidia:
Adds configurable `turn_counter` reminder placement for Qwen-compatible SWE-bench runs:

- `position: new_user_message` appends threshold reminders as a new trailing user message instead of a mid-trajectory system message or mutating last user message
- `position: tool_message` appends periodic reminders to the trailing tool response
- `pre_threshold_tool_reminder_interval` appends pre-threshold environment reminders to the trailing tool response every N turns before threshold reminders begin

## PR Overview

`turn_counter` is a **request interceptor**. On every outgoing chat-completion request it increments a
per-session counter (`sess.count += 1`, one increment per agent generation), and — depending on
config — appends a "you are running out of turns" reminder into the request payload before it is
forwarded to the model. It also hard-rejects the session (`GracefulError`) once `count > max_turns`.

Before this PR there were **two configuration axes**:

- `position` in {`system_message`, `user_message`} — *where* the reminder lands.
- `trigger` in {`threshold` (fires at 80%/95% of `max_turns`), `periodic` (every `interval` turns)} — *when*.

This PR adds **three things** that are easy to conflate but are not the same kind of knob:

### 1. A third `position`: `new_user_message`

- `user_message` (pre-existing): **mutates** the content of the *last existing* `role:"user"` message.
  In agentic SWE runs there is typically one user message: the task prompt.
  So the reminder is concatenated onto the original task prompt, which sits at the top of the
  conversation. The conversation still ends on the `tool` result. No `[SYSTEM]` prefix.
- `new_user_message` (new): **appends a brand-new** `{"role":"user", ...}` message at the tail of
  the array, after the trailing tool message. For threshold reminders, it also receives a `[SYSTEM]`
  text prefix.

Invoked side-by-side at turn 8/10 (threshold WARN), realistic SWE payload:

```text
position=user_message      -> 16 msgs; MUTATED user msg[1] (the task):
   '...</issue_details>\n\nTurn 8/10 — 2 turns remaining. Begin wrapping up...'   (trailing role: tool)
position=new_user_message  -> 17 msgs; APPENDED new tail user msg:
   '[SYSTEM] Turn 8/10 — 2 turns remaining. Begin wrapping up...'                 (trailing role: user)
```

So `user_message` vs `new_user_message` differ on two dimensions, not one: mutate-in-place vs
append-new-turn, and no prefix vs `[SYSTEM]` prefix. The reminder text is not identical between the
two modes.

### 2. A periodic-only `position`: `tool_message`

`tool_message` appends the flat `ENVIRONMENT REMINDER:` template to the trailing `tool` message.
It is only supported with `trigger=periodic`; using it with `trigger=threshold` raises.

Example:

```yaml
trigger: periodic
position: tool_message
interval: 20
```

If the last message is not a `tool`, injection is skipped and logged.

Periodic mode now defaults to:

```yaml
trigger: periodic
position: tool_message
interval: 1
```

### 3. A new cadence knob `pre_threshold_tool_reminder_interval` (NOT a top-level position)

`pre_threshold_tool_reminder_interval: N` adds a second, separate reminder channel that:

- fires only under `trigger=threshold` and only in the sub-threshold (`NON_ACTIONABLE`) zone,
  i.e. below 80% of `max_turns`;
- on turns where `n % N == 0`, appends the plain `ENVIRONMENT REMINDER:` template to the trailing
  `tool` message via `_append_to_trailing_tool_message`, skipping (INFO log) if the last message is
  not a `tool`;
- ignores the top-level `position` entirely — it always targets the tool message;
- stops once the threshold is reached (the normal position-governed reminder takes over);
- is warned-and-ignored under `trigger=periodic`.

Empirical two-phase behavior (`max_turns=200`, threshold@160, `pre_threshold_tool_reminder_interval=20`):

```text
position='user_message'        position='new_user_message'
 turn 20 -> tool msg[3]          turn 20 -> tool msg[3]   (IDENTICAL — position ignored)
 turn 40 -> tool msg[3]          turn 40 -> tool msg[3]   (IDENTICAL)
 turn160 -> user msg[1]          turn160 -> NEW user @tail (position now matters)
 turn180 -> user msg[1]          turn180 -> NEW user @tail
```

So `pre_threshold_tool_reminder_interval` is effectively a periodic reminder that only runs below
threshold and always writes to the tool message. It is a third behavior that is neither a top-level
`position` nor a clean top-level `trigger`.

### `periodic` vs `pre_threshold_tool_reminder_interval` — both fire "every N turns", but are not the same

Both emit the same flat `ENVIRONMENT REMINDER: ...` text on a fixed cadence, so they are easy to
confuse. They differ on several axes. Side-by-side, both using tool-message placement, `max_turns=10`,
cadence 2:

```text
A) trigger=periodic, position=tool_message, interval=2
B) trigger=threshold, position=new_user_message, pre_threshold_tool_reminder_interval=2

  turn  2: tool[3]  "...8 turns left"      turn  2: tool[3]  "...8 turns left"
  turn  4: tool[3]  "...6 turns left"      turn  4: tool[3]  "...6 turns left"
  turn  6: tool[3]  "...4 turns left"      turn  6: tool[3]  "...4 turns left"
  turn  8: tool[3]  "...2 turns left"      turn  8: user tail "Turn 8/10 — wrap up"   <- escalates
  turn 10: tool[3]  "...0 turns left"      turn  9: user tail "Turn 9/10 — wrap up"
                                           turn 10: user tail "URGENT: ... answer NOW"
```

| axis | `trigger: periodic` | `pre_threshold_tool_reminder_interval: N` |
|---|---|---|
| what it is | a value of `trigger` (mutually exclusive with `threshold`) | an int add-on, only active under `trigger=threshold` (warned+ignored under periodic) |
| cadence param | `interval` (`n % interval`) | `pre_threshold_tool_reminder_interval` itself (`n % pre_threshold_tool_reminder_interval`) |
| active window | the entire run | only below the 80% threshold; stops once threshold hits |
| position | respects top-level `position`, including periodic-only `tool_message` | ignores top-level `position` — always the trailing `tool` message |
| escalation | none — flat reminders forever | phase 1 of an escalating run — then threshold takes over with WARN→URGENT |
| reminder text | `ENVIRONMENT REMINDER: ...` (flat) | `ENVIRONMENT REMINDER: ...` (flat) — identical while firing |

Mental model: `periodic` = one uniform single-phase cadence the whole run, at your chosen top-level
position, no urgency ramp. `pre_threshold_tool_reminder_interval` = light tool-message nudges early/mid-run,
then handoff to escalating WARN/URGENT threshold reminders at the end. Periodic can now target tool
messages, but periodic alone still cannot reproduce the two-phase handoff behavior.

## Summary by CodeRabbit

* **New Features**
  * Added `NEW_USER_MESSAGE` injection support for turn-counter logic, appending a trailing user reminder on the correct cadence for both periodic and threshold modes.
  * Added `pre_threshold_tool_reminder_interval` to optionally insert an environment reminder into the most recent trailing tool message before the threshold reminder.

* **Bug Fixes**
  * Improved reminder insertion behavior, including consistent prefixing and safer handling of tool vs non-tool targets and varied tool content formats.

* **Tests**
  * Expanded test coverage for new injection positions, pre-threshold tool reminders (including skip/no-change scenarios), and configuration validation/warning behavior.


iguehring-nvidia:
## PR Overview
@fsiino-nvidia can you confirm that this correctly summarizing what your PR does? Would be great to have that in the PR overview.

`turn_counter` is a **request interceptor**. On every outgoing chat-completion request it increments a
per-session counter (`sess.count += 1`, one increment per agent generation), and — depending on
config — appends a "you are running out of turns" reminder into the request payload before it is
forwarded to the model. It also hard-rejects the session (`GracefulError`) once `count > max_turns`.

Before this PR there were **two axes**, documented as orthogonal:

- `position` ∈ {`system_message`, `user_message`} — *where* the reminder lands.
- `trigger` ∈ {`threshold` (fires at 80%/95% of `max_turns`), `periodic` (every `interval` turns)} — *when*.

This PR adds **two independent things** that are easy to conflate but are not the same kind of knob:

### 1. A third `position`: `new_user_message`

- `user_message` (pre-existing): **mutates** the content of the *last existing* `role:"user"` message
  (`_append_to_last_user_message`, src ~L235). In agentic SWE runs there is only **one** user
  message: the task prompt.
  So the reminder is concatenated onto the **original task prompt**, which sits at the top of the
  conversation. The conversation still **ends on the `tool` result**. No `[SYSTEM]` prefix.
- `new_user_message` (new): **appends a brand-new** `{"role":"user", ...}` message at the **tail** of
  the array (src L210–211), after the trailing tool message. It also receives a `[SYSTEM] ` text prefix
  (grouped with `system_message`, src L198–201).

Invoked side-by-side at turn 8/10 (threshold WARN), realistic SWE payload:

```
position=user_message      -> 16 msgs; MUTATED user msg[1] (the task):
   '...</issue_details>\n\nTurn 8/10 — 2 turns remaining. Begin wrapping up...'   (trailing role: tool)
position=new_user_message  -> 17 msgs; APPENDED new tail user msg:
   '[SYSTEM] Turn 8/10 — 2 turns remaining. Begin wrapping up...'                 (trailing role: user)
```

So `user_message` vs `new_user_message` differ on **two** dimensions, not one: (a) mutate-in-place vs
append-new-turn, and (b) no prefix vs `[SYSTEM]` prefix. The reminder *text* is not identical between
the two modes.

### 2. A new cadence knob `remind_every` (NOT a position)

`remind_every: N` adds a **second, separate reminder channel** that:

- fires **only under `trigger=threshold`** and **only in the sub-threshold (`NON_ACTIONABLE`) zone**,
  i.e. *below* 80% of `max_turns` (src L192–196);
- on turns where `n % N == 0`, appends the plain `ENVIRONMENT REMINDER:` template to the **trailing
  `tool` message** via `_append_to_trailing_tool_message` (src ~L265), skipping (INFO log) if the last
  message is not a `tool`;
- **ignores `position` entirely** — it always targets the tool message;
- stops once the threshold is reached (the normal position-governed reminder takes over);
- is warned-and-ignored under `trigger=periodic` (src L123–128).

Empirical two-phase behavior (`max_turns=200`, threshold@160, `remind_every=20`):

```
position='user_message'        position='new_user_message'
 turn 20 -> tool msg[3]          turn 20 -> tool msg[3]   (IDENTICAL — position ignored)
 turn 40 -> tool msg[3]          turn 40 -> tool msg[3]   (IDENTICAL)
 turn160 -> user msg[1]          turn160 -> NEW user @tail (position now matters)
 turn180 -> user msg[1]          turn180 -> NEW user @tail
```

So `remind_every` is effectively *"a periodic trigger that only runs below threshold and always writes
to the tool message"* bolted onto the threshold branch. It is a third behavior that is neither a clean
`position` nor a clean `trigger`.

### `periodic` vs `remind_every` — both fire "every N turns", but are not the same

Both emit the **same flat** `ENVIRONMENT REMINDER: …` text on a fixed cadence, so they are easy to
confuse. They differ on four axes. Side-by-side, both pinned to `position=user_message`, `max_turns=10`,
cadence 2 (`/tmp/pvr.py`):

```
A) trigger=periodic, interval=2          B) trigger=threshold, remind_every=2
  turn  2: user[1]  "...8 turns left"      turn  2: tool[3]  "...8 turns left"
  turn  4: user[1]  "...6 turns left"      turn  4: tool[3]  "...6 turns left"
  turn  6: user[1]  "...4 turns left"      turn  6: tool[3]  "...4 turns left"
  turn  8: user[1]  "...2 turns left"      turn  8: user[1]  "Turn 8/10 — wrap up"   <- escalates
  turn 10: user[1]  "...0 turns left"      turn  9: user[1]  "Turn 9/10 — wrap up"
                                           turn 10: user[1]  "URGENT: ... answer NOW"
```

| axis | `trigger: periodic` | `remind_every: N` |
|---|---|---|
| what it is | a value of `trigger` (mutually exclusive with `threshold`) | an int add-on, only active under `trigger=threshold` (warned+ignored under periodic) |
| cadence param | `interval` (`n % interval`) | `remind_every` itself (`n % remind_every`) |
| active window | the **entire run** | **only below the 80% threshold**; stops once threshold hits |
| position | **respects `position`** (system/user/new_user) | **ignores `position`** — always the trailing `tool` message |
| escalation | **none** — flat reminders forever (turn 10 is still the bland template) | the gentle phase-1 of an escalating run — then threshold takes over with WARN→URGENT |
| reminder text | `ENVIRONMENT REMINDER: …` (flat) | `ENVIRONMENT REMINDER: …` (flat) — **identical** while firing |

Mental model: `periodic` = one uniform single-phase cadence the whole run, at your chosen position, no
urgency ramp. `remind_every` = light tool-message nudges early/mid-run (reasoning-safe, position-fixed),
then hand off to escalating WARN/URGENT threshold reminders at the end — a two-phase pattern that exists
only because the trigger model allows a single trigger per run. You cannot reproduce `remind_every` with
`periodic`: periodic has no escalation, fires the whole run rather than pre-threshold only, and has no
`tool` position to target. (This is the design awkwardness flagged in §R3.)

iguehring-nvidia:
@fsiino-nvidia If the description in https://github.com/NVIDIA-NeMo/Evaluator/pull/1090#issuecomment-4809632017 is accurate, I have a couple of questions / comments:

1. Are you aware that `new_user_message` will remove all prior reasoning for e.g., Qwen (and other models)? Often times, interleaved reasoning is only kept until the next user message. This could degrade performance as the model looses context.
2. Are you aware that with `user_message`, the Qwen chat template error that arises from using `system_message` can be resolved as well? It has a different behavior than `new_user_message`, but it also resolves the system message error. Just want to make sure that you are aware.
3. We add a user message but prefix it with `SYSTEM`: `[SYSTEM] Turn 8/10 — 2 turns remaining. Begin wrapping up...`. Just a heads-up that this is not consistent.
4. Once 1. - 3. are clarified, I don't have concerns with merging `new_user_message`.
5. I have more concerns with `remind_every: <int>` flag: 

- With the PR, we have ` trigger: periodic` and `remind_every: <int>` flags. Thats a bit of confusing user / config experience. One would assume these would do the same.
- Similarly, IIUC, it would be used like this:

```
 - name: turn_counter
        config:
          max_turns: 200
          trigger: threshold
          position: new_user_message   # 80%/95% escalation -> fresh trailing USER turn
          remind_every: 20             # before 80%: nudge every 20 turns on the TOOL message
```
That implies that `remind_every` configures when the `new_user_message` is applied. However, it rather configures an extra reminder. I am wondering if there is a cleaner way.

fsiino-nvidia:
@iguehring-nvidia Yes, the overview is accurate and looks good to me.

Regarding 1-3, these are downstream changes ported over from the runtime patch by the recipe PIC @sdevare-nv, who can weigh in here.

For `role=user` with `[SYSTEM]` prefix inconsistency, my understanding is that the patch intentionally preserved the original append-new-message placement, but changed the role from `system` to `user` because Qwen rejects mid-conversation system messages. The `[SYSTEM]` prefix was kept so the reminder still reads like an evaluator/framework directive. We can probably add a multi-line comment explaining this.

For the `remind_every` / config clarity, I agree the config shape is confusing now that we've added more complexity. We can rename it to `pre_threshold_tool_reminder_interval` and be sure any future run configs reflect that update.

Let me know your thoughts.

iguehring-nvidia:
@fsiino-nvidia I was thinking more in the direction of a refactor that would make `pre_threshold_tool_reminder_interval` its own position rather than a flag that additionally adds something. But to unblock, I am fine with the renaming. Three asks then:

- Please update the MR overview with the text from my comment sucht that the behavior is properly documented.
- Please drive the refactor that we discussed (renaming).
- Please drive the acknowledgement / discussion of 1-3. If its  @sdevare-nv who should comment here, pls bring it to his attention.

fsiino-nvidia:
@sdevare-nv can you please answer for items 1-3 from Ingo so we are all aligned?

> 1. Are you aware that `new_user_message` will remove all prior reasoning for e.g., Qwen (and other models)? Often times, interleaved reasoning is only kept until the next user message. This could degrade performance as the model looses context.
> 2. Are you aware that with `user_message`, the Qwen chat template error that arises from using `system_message` can be resolved as well? It has a different behavior than `new_user_message`, but it also resolves the system message error. Just want to make sure that you are aware.
> 3. We add a user message but prefix it with `SYSTEM`: `[SYSTEM] Turn 8/10 — 2 turns remaining. Begin wrapping up...`. Just a heads-up that this is not consistent.

src/nemo_evaluator/adapters/interceptors/turn_counter.py
```diff
@@ -33,9 +33,11 @@
 
 
 class InjectionPosition(str, Enum):
+    # ``tool_message`` is only supported with ``trigger=periodic``.
```
iguehring-nvidia, line 36: any reason why `tool_message` is only supported with `trigger=periodic`?

fsiino-nvidia: @iguehring-nvidia `trigger: periodic` has no escalation phase, so reminders go to the configured `position` every `interval` turns. If someone wants recurring tool-response reminders for the whole run, they can use:

```yaml
trigger: periodic
position: tool_message
interval: 1
```

`trigger: threshold` is different because before the escalation phase, `pre_threshold_tool_reminder_interval` already goes into the `tool_message` position. After that, it goes to a non-`tool_message` position so it's not redundant or confusing.

iguehring-nvidia (approved):

PR #1093, opened 2026-06-25
feat: in-container local sandbox + reasoning-model scoring fixes

hychiang-git:
## What
Fixes that enable faithful evaluation of reasoning + tool-using checkpoints on
SLURM/pyxis clusters (no Docker daemon / remote cloud needed).

## 1. Scoring/serving bugs that grade correct answers as wrong
Multiple-choice benchmarks scored far below a model's true accuracy because of bugs in the
scoring/request path — the model answered correctly, but the harness never credited it:

**(a) The answer was read from the wrong field.** Served with a reasoning parser, the final
`Answer: X` lands in `reasoning_content` and `content` is empty. `ChatSolver` and the ReAct
solver scored `content` only → ~98% of responses graded wrong (**GPQA pass@1 1.3%** despite
correct answers). *Fix:* score `reasoning_content` + `content`
([`chat.py`](https://github.com/hychiang-git/Evaluator/blob/94e92934/src/nemo_evaluator/solvers/chat.py), [`react.py`](https://github.com/hychiang-git/Evaluator/blob/94e92934/src/nemo_evaluator/solvers/react.py) → `_extract_last_text`).

**(b) Answer extraction was too strict.** [`multichoice_regex`](https://github.com/hychiang-git/Evaluator/blob/94e92934/src/nemo_evaluator/scoring/pattern.py) required a
*bare* letter right after "Answer:" and was hard-coded to `A-D`, missing wrapped answers
(`$B$`, `**B**`, `(B)`, `\boxed{B}`) and >4-choice benchmarks. *Fix:* a generic `letters`
arg (`"A-D"`, `"A-J"`, …) that builds a wrapper-tolerant pattern + takes the last match;
callers declare their set ([`gpqa.py`](https://github.com/hychiang-git/Evaluator/blob/94e92934/src/nemo_evaluator/benchmarks/gpqa.py) `A-D`,
[`mmlu_pro.py`](https://github.com/hychiang-git/Evaluator/blob/94e92934/src/nemo_evaluator/benchmarks/mmlu_pro.py) `A-J`). **MMLU-Pro 63% → 80%.**

**(c) The request timeout was too short.** The service `request_timeout` was applied to the
proxy→server hop but **not** the eval's `ModelClient` (built with no timeout → 120s default),
so long reasoning outputs under concurrency timed out → empty (0-token) responses scored 0.
*Fix:* thread `request_timeout` into the client in
[`orchestrator.py`](https://github.com/hychiang-git/Evaluator/blob/94e92934/src/nemo_evaluator/orchestration/orchestrator.py).

## 2. `local` sandbox backend
Expose `sandbox: {type: local}` ([`LocalSandbox` config](https://github.com/hychiang-git/Evaluator/blob/94e92934/src/nemo_evaluator/config/sandboxes.py);
[`_make_sandbox_manager`](https://github.com/hychiang-git/Evaluator/blob/94e92934/src/nemo_evaluator/orchestration/orchestrator.py) builds `backend="local"`). The
`SandboxManager` backend already existed, just wasn't reachable from config — so tool benchmarks
execute model code in-container (no Docker daemon, no nested pyxis, no remote cloud).

Why a `local` backend: on SLURM/pyxis clusters the other backends don't fit plain-python tool
benchmarks — `docker` needs a daemon (absent on compute nodes), the `slurm` sandbox must spawn a
*nested* pyxis container from inside the eval container, `apptainer` may not be installed, and
`ecs_fargate` is remote cloud (credentials, ECR, SSH sidecar) — overkill just to run a Python
snippet. For tools that only need an interpreter, running the model's code as a subprocess in the
eval container (which already ships python) is the zero-infrastructure option.

Hardening in [`local.py`](https://github.com/hychiang-git/Evaluator/blob/94e92934/src/nemo_evaluator/sandbox/local.py) → `LocalSandbox.exec`: SIGKILL the whole process
group on timeout (a wedged child can't hang the eval) + `cwd` falls back to the tempdir for
missing paths. [`slurm_gen.py`](https://github.com/hychiang-git/Evaluator/blob/94e92934/src/nemo_evaluator/orchestration/slurm_gen.py) and
[`manager.py`](https://github.com/hychiang-git/Evaluator/blob/94e92934/src/nemo_evaluator/sandbox/manager.py) skip the registry pre-pull for local `.sqsh` /
absolute-path images.

## 3. Export runtime token/step stats to MLflow
The MLflow exporter built metrics only from top-level score entries that were scalars or carried
a `value` key, so the benchmark `runtime` block (a nested dict of token/step/latency counters) was
dropped entirely — downstream dashboards had no tokens-per-sample signal for a run. *Fix:*
[`mlflow_export.py`](https://github.com/hychiang-git/Evaluator/blob/7d6159a7/src/nemo_evaluator/engine/exporters/mlflow_export.py)
`_build_metrics` flattens the `runtime` block into flat `runtime_*` metrics and derives per-sample
token averages (`runtime_avg_completion_tokens` / `runtime_avg_total_tokens` = totals ÷
`total_steps`, guarding zero steps); `bool` is excluded so flags aren't logged as 0/1 metrics.

## 4. gym:// native verify (NeMo-Gym resource servers)
Running a NeMo-Gym benchmark via `gym://…?protocol=native` (the evaluator drives the rollout and
POSTs to the gym resource server's `/verify`) failed or scored 0, for two reasons. *Fix:*
[`gym_protocol.py`](https://github.com/hychiang-git/Evaluator/blob/625f1275/src/nemo_evaluator/environments/gym_protocol.py)
+ [`gym.py`](https://github.com/hychiang-git/Evaluator/blob/625f1275/src/nemo_evaluator/environments/gym.py):

- **`wrap_text_as_gym_response` was missing required fields** → HTTP **422** at `/verify`. Recent
  NeMo-Gym servers validate `body.response` against the openai `Response` schema, which now requires
  `parallel_tool_calls` / `tool_choice` / `tools`. Added them.
- **`_verify_native` didn't forward the gold.** In native mode the evaluator holds the dataset and
  never calls `/seed_session`, so a dataset-graded server (e.g. mcqa) can't learn the expected answer
  any other way → every reward was 0. Now forwards `expected_answer` in the verify body.

Validated end-to-end (mcqa resource server, hosted policy): `/verify` returns 200 and a correct
single-letter answer scores 1.0.

## Impact
On a reasoning checkpoint: empty scored-responses **98% → 0%**; GPQA **~1.3% → realistic**,
MMLU-Pro **~63% → ~80%**. No behavior change for non-reasoning models; `local` is opt-in. Runtime token/step stats are now exported to MLflow (per-sample token averages) for dashboards.

## Tests
- [`test_benchmark_definitions.py`](https://github.com/hychiang-git/Evaluator/blob/94e92934/tests/test_scoring/test_benchmark_definitions.py) — parametrized
  `multichoice_regex` cases: `$B$`/`**B**`/`(B)`/`\boxed{D}`, last-match, placeholder
  reject, 10-choice via `letters="A-J"`, out-of-range.
- [`test_solvers.py`](https://github.com/hychiang-git/Evaluator/blob/94e92934/tests/test_solvers/test_solvers.py) / [`test_react_solver.py`](https://github.com/hychiang-git/Evaluator/blob/94e92934/tests/test_solvers/test_react_solver.py)
  — `ChatSolver` and ReAct `_extract_last_text` fold `reasoning_content` into the score.
- [`test_sandbox_backends.py`](https://github.com/hychiang-git/Evaluator/blob/94e92934/tests/test_sandbox/test_sandbox_backends.py) — `local` sandbox
  resolves + provisions, `exec` times out via process-group kill (no hang), `cwd` fallback.

- [`test_mlflow_exporter.py`](https://github.com/hychiang-git/Evaluator/blob/7d6159a7/tests/test_engine/test_mlflow_exporter.py)
  — the `runtime` block is flattened into `runtime_*` metrics, per-sample token averages are
  derived, and a zero `total_steps` is guarded (no divide-by-zero, no derived avg).

`ruff` clean; scoring/solver/sandbox suites pass (243 tests) + the MLflow exporter suite (incl. 2 new runtime-metric tests).

## Summary by CodeRabbit

* **New Features**
  * Added a local in-process sandbox backend for tool execution.
* **Bug Fixes**
  * Improved local sandbox timeout handling with process-group termination (exit code now 124).
  * Improved local execution when a requested working directory is missing (falls back to temp workdir).
  * Honor proxy request timeouts in evaluation model clients.
  * Reduced SLURM image pre-pull attempts for local/absolute image paths.
  * Solver and scoring now include reasoning text when it contains the final answer.
* **Tests**
  * Expanded coverage for local sandbox timeout/cwd behavior and scoring/solver reasoning-extraction edge cases.

### DIFF
--- a/src/nemo_evaluator/benchmarks/mmlu_pro.py
+++ b/src/nemo_evaluator/benchmarks/mmlu_pro.py
@@ -48,4 +48,5 @@ def _prepare(row, idx, rng):
 )
 @scorer
 def mmlu_pro_scorer(sample: ScorerInput) -> dict:
-    return multichoice_regex(sample, pattern=r"(?i)Answer\s*:\s*([A-J])")
+    # 10-choice; the wrapper-tolerant pattern is built from the letter set.
+    return multichoice_regex(sample, letters="A-J")
--- a/src/nemo_evaluator/config/__init__.py
+++ b/src/nemo_evaluator/config/__init__.py
@@ -48,6 +48,7 @@ from .sandboxes import (
     CustomSandbox,
     DockerSandbox,
     EcsFargateSandbox,
+    LocalSandbox,
     NoSandbox,
     SandboxConfig,
     SlurmSandbox,
@@ -123,6 +124,7 @@ __all__ = [
     "SlurmSandbox",
     "ApptainerSandbox",
     "NoSandbox",
+    "LocalSandbox",
     "CustomSandbox",
     "SandboxConfig",
     # Solvers
--- a/src/nemo_evaluator/config/sandboxes.py
+++ b/src/nemo_evaluator/config/sandboxes.py
@@ -161,6 +161,21 @@ class NoSandbox(_SandboxBase):
     stateful: bool = False
 
 
+class LocalSandbox(_SandboxBase):
+    """In-process sandbox: tool commands run as subprocesses in a temp dir
+    inside the eval container itself — no container isolation, no remote infra.
+
+    Ideal for plain-python tool benchmarks (e.g. gpqa/aime25 with tools) on
+    clusters where the other backends are unavailable: docker has no daemon,
+    the slurm backend needs nested pyxis, and apptainer may not be installed.
+    The eval image already ships python3/bash, so the model's tool code runs
+    right there. Backed by nemo_evaluator.sandbox.local.LocalSandbox (the
+    SandboxManager 'local' backend, which already exists)."""
+
+    type: Literal["local"] = "local"
+    concurrency: int = 4
+
+
 class CustomSandbox(BaseModel):
     """Plugin sandbox backend — dynamically imported from class_path."""
 
@@ -189,6 +204,7 @@ SandboxConfig = Annotated[
     | Annotated[SlurmSandbox, Tag("slurm")]
     | Annotated[ApptainerSandbox, Tag("apptainer")]
     | Annotated[NoSandbox, Tag("none")]
+    | Annotated[LocalSandbox, Tag("local")]
     | Annotated[CustomSandbox, Tag("custom")],
     Discriminator(_sandbox_discriminator),
 ]
--- a/src/nemo_evaluator/engine/exporters/mlflow_export.py
+++ b/src/nemo_evaluator/engine/exporters/mlflow_export.py
@@ -313,8 +313,29 @@ class MLflowExporter:
         for metric, val in scores.items():
             if isinstance(val, dict) and "value" in val:
                 val = val["value"]
-            if isinstance(val, (int, float)):
+            # bool is an int subclass; exclude so flags aren't logged as 0/1 metrics.
+            if isinstance(val, (int, float)) and not isinstance(val, bool):
                 safe_metrics[mlflow_sanitize(metric, "metric")] = float(val)
+
+        # The `runtime` block (token/step/latency counters) is a nested dict with no
+        # top-level "value", so the loop above drops it. Flatten its numeric children
+        # into flat `runtime_*` metrics and derive per-sample token averages, so
+        # downstream dashboards can surface tokens-per-sample for a run.
+        runtime = scores.get("runtime")
+        if isinstance(runtime, dict):
+            for rk, rv in runtime.items():
+                if isinstance(rv, (int, float)) and not isinstance(rv, bool):
+                    safe_metrics[mlflow_sanitize(f"runtime_{rk}", "metric")] = float(rv)
+            steps = runtime.get("total_steps")
+            if isinstance(steps, (int, float)) and not isinstance(steps, bool) and steps:
+                for src, dst in (
+                    ("total_completion_tokens", "runtime_avg_completion_tokens"),
+                    ("total_tokens", "runtime_avg_total_tokens"),
+                    ("total_reasoning_tokens", "runtime_avg_reasoning_tokens"),
+                ):
+                    tot = runtime.get(src)
+                    if isinstance(tot, (int, float)) and not isinstance(tot, bool):
+                        safe_metrics[dst] = float(tot) / float(steps)
         return safe_metrics
 
     def _export_one_bundle(self, bundle: dict[str, Any], output_dir: Path) -> str:
--- a/src/nemo_evaluator/environments/gym.py
+++ b/src/nemo_evaluator/environments/gym.py
@@ -207,7 +207,7 @@ class GymEnvironment(EvalEnvironment):
         return SeedResult(
             prompt=d.get("prompt", ""),
             expected_answer=d.get("expected_answer", ""),
-            metadata=d.get("metadata", {}),
+            metadata=(d.get("metadata") or {}),
             messages=d.get("messages"),
             system=d.get("system"),
             sandbox_spec=sandbox_spec,
@@ -232,7 +232,7 @@ class GymEnvironment(EvalEnvironment):
             reward=float(d.get("reward", 0.0)),
             extracted_answer=d.get("extracted_answer"),
             scoring_details=d.get("scoring_details", {}),
-            metadata=d.get("metadata", {}),
+            metadata=(d.get("metadata") or {}),
         )
 
     async def _verify_native(self, response: str, expected: str, **meta: Any) -> VerifyResult:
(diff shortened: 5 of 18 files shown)


hychiang-git:
Thanks @coderabbitai — addressed all five actionable comments (+ the `local.py` cwd nit) in the latest push:

- **`manager.py`** — the local-image import skip now covers `.sqsh` **and** absolute-path (`/`-prefixed) images, matching the `slurm_gen.py` pre-pull gate.
- **`pattern.py`** — `multichoice_regex` now tolerates `\boxed{B}` (added `(?:\\boxed\s*\{\s*)?` before the wrapper class), and stays generic via the `letters` arg.
- **`react.py`** — added `TestReActExtractLastText` covering the reasoning-only final-answer path (empty `content`, populated `reasoning_content`) plus the plain-content case.
- **`test_sandbox_backends.py`** — renamed the unused stacked-patch arg to `_mock_getpgid` (ARG002).
- **`test_benchmark_definitions.py`** — collapsed the multichoice cases into one `@pytest.mark.parametrize` with descriptive `ids` (incl. the new `boxed` case).
- **`local.py`** — cwd fallback now triggers on `not .exists()` (missing path) rather than `not .is_dir()`, so a real-but-non-dir path still errors.

Also threaded the service `request_timeout` into the eval `ModelClient` (it only governed the proxy→server hop; the 120s client default was timing out long reasoning outputs → empty responses).

`ruff` clean; scoring/solver/sandbox suites pass (243 tests).

hychiang-git:
Thanks for the thorough review. Addressed in `7940e423`:

- **`multichoice_regex` multi-group override** — switched `re.findall` → `re.finditer` + `group(1)`, so an explicit `pattern=` with auxiliary capture groups no longer crashes on tuples. Added a `single-group`/`multi-group` parametrized regression.
- **`ChatSolver` model_response alignment** — `model_response.content` is now set to the scored text, matching `ReActSolver`, so reasoning-only models don't persist an empty answer artifact. Asserted in the fold-in test.
- **`request_timeout` fallback duplication** — extracted `_resolve_request_timeout()` and used it in both client paths.
- **Test conventions** — collapsed the individual `multichoice_regex` cases into one `@pytest.mark.parametrize` table with descriptive `ids`; hoisted `SimpleNamespace`/`ChatSolver` imports to module scope.

Already in place on the prior commit (`94e92934`):

- `\boxed{B}` is matched by the default pattern (`(?:\\boxed\s*\{\s*)?`).
- `manager.py` skips import for both `.sqsh` and absolute-path images.
- `local.py` cwd fallback uses `.exists()` (non-dir paths still error).
- `test_exec_timeout` uses `_mock_getpgid`; ReAct reasoning-only path is covered by `TestReActExtractLastText`.

`ruff check`/`format` clean and the scoring/solver/sandbox suites pass locally.

hychiang-git:
Follow-up review addressed in `86fb7006`:

- **`_resolve_request_timeout` typed `Any` (ANN401)** — changed to `service: object | None`; behavior unchanged (still handles `None`).
- **Assert `extracted` in scorer tests** — replaced the boolean `expect_none` flag with an `expected_extracted` column and assert it for every case (including the wrong-target and the override-pattern cases), so extraction regressions can't slip through behind a passing `correct`.

Re: the `chat.py` model_response note — that's already in this branch: `resp.content = scored` is set before constructing the `SolveResult` (chat.py, just above the `build_single_turn_atif` call), and `test_folds_reasoning_content_into_scored_response` asserts `result.model_response.content` carries the scored text. Looks like the thread was re-anchored onto the new commit.

`ruff check`/`format` clean; scoring suite passes locally.

hychiang-git:
Addressed in `d0736655`:

- **Ruff PT006** — the multichoice parametrize now passes its arg names as a tuple instead of a comma-delimited string (`ruff check --select PT006` clean).

The other two threads re-surfaced on this commit are already satisfied in-branch:

- **`chat.py` model_response alignment** — `resp.content = scored` is set just before `build_single_turn_atif(...)`; `test_folds_reasoning_content_into_scored_response` asserts `result.model_response.content` carries the scored text.
- **ReAct reasoning-only regression test** — covered by `TestReActExtractLastText` in `tests/test_solvers/test_react_solver.py` (`test_folds_reasoning_only_final_answer` exercises a reasoning-only `ToolCallingResponse` through `_extract_last_text`, and `test_plain_content_unchanged` guards the normal path).

`ruff check`/`format` clean; scoring + solver suites green.

PR #1099, opened 2026-07-06
fix: harbor failure attribution, timeout flush, and crash-safe checkpoint resume

AWarno:
One PR, five related reliability fixes for Harbor benchmark runs (SWE-bench, Terminal-Bench). It absorbs stacked PRs #1114 and #1116 (squash-merged into this branch after independent review) and has `main` merged in (conflicts resolved; the diff below is purely this PR's delta). Closes FEP-1131, FEP-1169, FEP-1132, FEP-1295.

## What it does

**1. Failure attribution is now trustworthy (FEP-1131).** A shared anchored classifier replaces the substring heuristics: named LiteLLM/OpenAI exception classes (28 covered, incl. `ContextWindowExceededError`, `BudgetExceededError`, gateway/connection classes) map to stable categories; HTTP-status matching is anchored to real status contexts, so digits inside UUIDs/commit hashes can no longer fabricate `server_error`/`model_timeout`; turn-budget exits classify as `turn_budget_exhausted`.

**2. The scoring asymmetry is gone (FEP-1131).** Rollouts that crash or exhaust the turn budget *with workspace changes* proceed to verification instead of being forced to reward 0 — the model gets credit for a valid patch even when the agent died on the last mile.

**3. Timed-out work is preserved and scored (FEP-1169).** On `run_timeout`, OpenHands gets a SIGTERM flush plus a short post-timeout persistence grace, so the partial trajectory and workspace diff survive; diff-carrying timeouts go through verification labeled `solve_timeout`. A persisted retry-marker protocol (schema v2 with a retry-window validity check) stops hours-old recovered transient errors from forcing reward 0.

**4. `empty_response` is an observation, not a verdict (FEP-1132, was #1114).** An empty final assistant message no longer terminates classification with a failure category (it had fired on ordinary verified failures and even a passing rollout). Emptiness is reported as a per-trial `empty_final_response` flag plus an aggregate; the never-firing harbor suppression machinery is deleted.

**5. Walltime kills no longer eat trajectories or labels (FEP-1295, was #1116).** The resume checkpoint used to silently null any trajectory over 5 MB — after a SLURM walltime kill (the checkpoint being the only surviving copy) exactly the biggest, kill-adjacent timeout rollouts resumed as zero-step rows. Now: oversized trajectories spill to gzip sidecars (`trajectory_overflow/`, cap tunable via `NEL_MAX_TRAJECTORY_BYTES`) and resume dereferences them; `error`/`error_kind`/solver scoring details are persisted so verify-only resume replays failures with labels intact (`solve_timeout`, `turn_budget_exhausted`); unverified infra failures re-solve on resume; capture markers make a kill-during-workspace-capture detectable (`resume_workspace_capture="missing"`).

Supporting: `request_hash` persisted in `model_traffic.jsonl` for duplicate/retry visibility; failure-category docs updated; stale CHANGELOG rule removed from AGENTS.md.

Diff vs main: 24 files, +3039/−88 (about half is new tests: classifier matrix, timeout/flush, checkpoint kill/resume integration suites).

## Verification

- Full pre-merge campaign (Linear FEA-244): score-reproduction gates on SWE-bench Verified 500×5, Terminal-Bench 2.1 89×8, and SWE-bench Multilingual 300×5 against validated baselines, plus forced-timeout and forced-turn-budget mini-runs asserting the fix signatures from run artifacts.
- FEP-1295: local real-SIGKILL drill (pre-fix control reproduces the bug) + a cluster mini-run with 4 real 30-min walltime kills — 0/100 trajectories lost (originating incident: 11/11 lost), labels and capture detection verified; two-round independent review; evidence on #1116.
- Post-merge-resolution tree: full offline suite green (2253 passed), mirror CI green, and the confirmatory kill/resume mini-run on the exact merged tree (image `dev-20260717-143853-a422daac-harbor`, run `20260717_132222_c4e5cba6`) returned **MERGE-SAFE**: FEP-1295 battery 7/7 across 5 real walltime kills (0/100 trajectories lost), both conflict-resolution patch blocks execute 144/144, and the score delta vs the pre-merge run is not significant (paired McNemar p=0.108) — full decomposition in the verdict comment below.


src/nemo_evaluator/solvers/harbor.py
```diff
@@ -2110,7 +2275,9 @@ async def solve(
                     latency_ms=round(latency_ms, 2),
                 ),
                 trajectory=trajectory,
-                error=str(exc),
+                error=error,
```
piojanu, line 2671: **H1. [HIGH — blocker: bug + hard standards violation] Turn-budget reclassification sits in the wrong exception handlers: dead code in `except InfraError`, guaranteed UnboundLocalError in `except GracefulError`**

`src/nemo_evaluator/solvers/harbor.py:2213-2223, 2225-2237, 2278-2280`

The block at 2213-2223 (inside `except InfraError`) computes `error=None`, `error_kind=ErrorKind.SOLVE_TIMEOUT`, `scoring_details={"error_category": "turn_budget_exhausted"}` and logs "graceful turn-budget exit with workspace changes — submitting for verification", but the handler's return (2225-2237) still hardcodes `error=str(exc), error_kind=ErrorKind.INFRA` with no `scoring_details` — everything the block computes is discarded and the log line lies. So an InfraError carrying turn-budget text plus a workspace diff is still graded 0.0 as infra_error (FEP-1131's scoring-asymmetry requirement, unimplemented on this path).

Meanwhile `except GracefulError`'s return (2278-2280) was rewired to `error/error_kind/scoring_details` — names bound only at try-body lines 2101-2103, which no GracefulError path ever reaches → any GracefulError entering solve() raises UnboundLocalError, which eval_loop treats as a system error (expensive full-rollout retries or run abort) instead of main's graceful reward-0 return. The two edits were evidently meant for one handler. Neither handler has any test (`grep GracefulError tests/test_solvers/test_harbor_solver.py` → no hits) — hard breach of AGENTS.md Phase 4 "changed behavior → updated tests".

**Related requirement gap (FEP-1131, same fix site):** the ticket asks for "turn-budget exhaustion AND other graceful agent-crash exits" to take the verify-if-workspace-diff path. The gates at 2109/2119 match only turn-budget text (`_is_turn_budget_exhausted_error`), so e.g. a ContextWindowExceededError crash with a modified workspace (the ticket's own evidence #2) still discards the patch and grades 0.0 — now correctly labeled, still wrongly scored.

**Fix:** move the reclassification into the GracefulError handler (or a shared helper returning `(error, error_kind, scoring_details)`), make the InfraError handler return what it computes or delete the block, widen the gate beyond turn-budget text to other graceful crash markers, and add regression tests for BOTH handlers. Note the same 5-line block is triplicated at 2109/2119/2215 (see M2) — the helper extraction is the fix vehicle.

src/nemo_evaluator/solvers/harbor.py
```diff
@@ -1656,7 +1771,16 @@ async def _wait_for_agent(
                 jitter,
                 self._timeout_strategy,
             )
-            if sandbox.is_running:
+            response_grace = _agent_timeout_response_grace(effective_timeout - jitter)
```
piojanu, line 2141: **H2. [HIGH] Post-timeout grace budgets do not compose — OpenHands flush can be killed mid-write (the exact FEP-1169 symptom)**

`src/nemo_evaluator/solvers/harbor.py:43-45, 95-112, 1774-1807, 1941`; `src/nemo_evaluator/sandbox/docker.py:161-168`

The sandbox agent-command deadline is `effective_timeout + 30s` (`_SANDBOX_AGENT_TIMEOUT_GRACE_SECONDS`), but after NEL timeout at T the code now spends up to 15s response-persistence grace (`min(15s, 5%)`, 1774-1782) plus ≤10s SIGTERM exec (1783-1806) BEFORE the 30s flush wait (1807). SIGTERM lands ~T+16-25s while the sandbox deadline fires at T+30: DockerSandbox.exec kills the docker-exec client (docker.py:166 — client only, not the in-container process), agent_task returns, the flush wait is cut to ~5-14s, and solve() proceeds to `_download_agent_logs` while run_agent.py may still be writing → truncated/absent partial trajectory. The grace must be added INTO the inner deadline, not carved out of it.

**Fix:** `agent_exec_timeout = effective_timeout + response_grace + sigterm_budget + flush_wait + margin`; add a test asserting the composition.

**Reviewer note on FEP-1169 status:** the PR otherwise resolves the ticket's actual requirement — the ordering fix (cc057999) plus response grace (70ff6d5b), validated by the Kimi run's zero wire-vs-final mismatches. This composition defect is the one thing to fix/comment. Condensation-call work now lives in FEP-1221 and is NOT related to this PR — do not raise it here.

src/nemo_evaluator/observability/collector.py
```diff
@@ -39,6 +40,16 @@
 ]
 
 
+def _completed_verification_without_error(scoring_details: object) -> bool:
```
piojanu: **H3. [HIGH] empty_response suppression drops the ticket's patch/diff guard — the label can now never fire on any scored Harbor rollout**

`src/nemo_evaluator/observability/collector.py:43-50` (predicate), `:86-89` (call site); duplicate predicate at `src/nemo_evaluator/reports/trajectories.py:166-176`

FEP-1132's suggested fix had two conditions: suppress "when scoring_details.method == 'harbor' AND the SWE-bench report.patch_exists == true (or whenever a workspace diff exists)". The implemented predicate `_completed_verification_without_error` checks method=="harbor" + test_exit_code/test_summary present + no error — the patch/diff condition was dropped. Since `environments/harbor.py:962-968` writes test_exit_code and test_summary on EVERY scored rollout, the predicate is true for all of them, patch or no patch.

**Consequence:** a rollout with an empty final message and NO diff — which the FEP-1132 comment thread (quoting Ingo) flags as "might be proper model capability failures… thorough investigation required" — now gets no label at all. The over-fire is fixed by making the label unfireable, which erases exactly the signal that must stay visible and silently forecloses the ticket's pending investigation.

**Fix (cheap — the data is already plumbed):** `scoring_details["report"]` carries the full verifier report (`reward_details["report"] = report` in environments/harbor.py, from `/logs/verifier/report.json`), so require `report.patch_exists` (SWE-bench) or workspace-diff evidence (the `"[workspace modified]"` sentinel) as a benchmark-agnostic fallback, at BOTH suppression sites — which per M1 should become one shared helper anyway. Add a test: empty final message + no patch + verification ran → label kept.

src/nemo_evaluator/reports/trajectories.py
```diff
@@ -162,6 +163,62 @@ def _failure_error(row: dict[str, Any]) -> str:
...
+    category = _wire_failure_category(last_wire)
+    if not category:
+        return "", ""
+    return category, _truncate_text(_wire_failure_text(last_wire))
+
+
+def _trial_failure_category(row: dict[str, Any], all_wire_rows: list[dict[str, Any]]) -> str:
+    category = _failure_category(row)
+    wire_category, _ = _trial_wire_failure(row, all_wire_rows)
+    if wire_category and category in {"", "empty_response", "model_error"}:
```
piojanu: **H4. [HIGH] Wire-failure override ignores reward/verification — verified-passing trials get failure labels from one trailing non-2xx wire row**

`src/nemo_evaluator/reports/trajectories.py:206-213` (`_trial_failure_category`; persisted into enriched rows via `_enrich_bench` at 945-951)

The wire check runs FIRST and consults neither reward nor verification: `if wire_category and category in {"", "empty_response", "model_error"}: return wire_category` — it even overrides category `""` (no failure at all). Repro'd against head (reward=1.0, method=harbor, test_exit_code=0):

| Wire rows (same request_hash on both) | Result |
|---|---|
| [429, 200] — classic retry, success logs last | `''` — clean ✅ |
| [200, 429] — failed twin completes last | `'rate_limit'` ❌ |
| same, on a row labeled empty_response + verified | `'rate_limit'` ❌ (override runs before the harbor suppression) |
| [200] only | `''` — suppression works ✅ |

**Scoping:** the mislabel needs a FAILED row to complete LAST in the final session — a hung duplicate that 504s/429s at the proxy after the successful retry already finished (the PR's own fake-endpoint test produced exactly this shape: "1x200 + 6x504, duplicate request_hash detected"), or a genuinely failed trailing auxiliary call on a rollout that still passed. Within that window the impact is severe: the wrong label is written into `trajectories_enriched.jsonl` (`rows_reclassified_from_wire_failure`) and counted in `failures_by_category` — the same outcome-independence defect FEP-1132 reports, reintroduced by the PR that fixes it. The PR's own 1500-rollout validation had 26 trials with a non-2xx last wire row. All new tests only exercise reward=0.0 rows.

**Fix:** in `_trial_wire_failure`, ignore a failed last row whose request_hash also has a successful twin (`_wire_dedup_key` already exists), and/or skip the override when reward > 0 / `_completed_harbor_verification_without_error(row)`; never override category `""` for successful trials. Add reward=1.0 test rows.

**Optional empirical check:** in the Kimi run's trajectories_report.json (`/lustre/fsw/portfolios/coreai/users/awarno/nv-eval-rundirs/leaderboard/v1/moonshotai/Kimi-K2.6/swebench-verified-pr1099-e8a4f8b0/20260706_220337_09550176`), inspect the 26 "last wire non-200" trials for any with reward > 0.

src/nemo_evaluator/observability/collector.py
```diff
@@ -63,19 +74,18 @@ def _classify_failure(self, step: StepRecord) -> None:
...
-                step.failure_category = "rate_limit"
-            elif any(code in err for code in ("500", "502", "503", "504")):
-                step.failure_category = "server_error"
-            else:
-                step.failure_category = "model_error"
+        if isinstance(sd, dict) and sd.get("error_category") in MODEL_FAILURE_CATEGORIES:
+            step.failure_category = sd["error_category"]
+            return
+
+        scoring_error = str(sd.get("error") or "") if isinstance(sd, dict) else ""
```
piojanu, line 77: **H5. [HIGH — scoped] System (harness) errors now classified with the model-failure classifier — infra failures blamed on the model**

`src/nemo_evaluator/observability/collector.py:71-83`

Rows with `scoring_details.error_category == "system"` (not caught by the early returns at 71-79) fall through to `classify_model_failure(step.model_error or scoring_error) or "model_error"`. Repro: a Docker daemon failure during verify that exhausts eval_loop retries (error="500 Server Error for http+docker://localhost/v1.41/containers/create", standard docker-py text) → leading "500" → failure_category="server_error" (a MODEL_FAILURE_CATEGORIES member); "container exited with status 137" → "model_error". Pre-PR these rows were never model-classified — this regression corrupts model-vs-harness attribution in leaderboard audits.

**SCOPE:** the fall-through for error_category=="graceful" is largely INTENDED (FEP-1131 wants turn-budget/context-window crash text classified) — do not revert that half.

**Fix:** early-return for "system" rows (or route them to a non-model category such as "system"/"infra_error"); never let harness error text land in MODEL_FAILURE_CATEGORIES. Add the docker-py 500 repro as a test.

src/nemo_evaluator/solvers/harbor.py
```diff
@@ -1967,14 +2100,33 @@ async def solve(
...
+                if workspace_diff and _is_turn_budget_exhausted_error(crash_error):
+                    logger.info(
+                        "HarborSolver: agent hit turn budget with workspace changes — submitting for verification"
+                    )
+                    error_kind = ErrorKind.SOLVE_TIMEOUT
+                    scoring_details = {"error": crash_error, "error_category": "turn_budget_exhausted"}
+                else:
+                    error = crash_error
+                    logger.warning("HarborSolver: %s", error)
+            elif agent_timed_out and not response and last_llm_error:
```
piojanu, line 2516: **H6. [HIGH] Stale `last_llm_error.json` marker corrupts rewards: written on every recovered retry, never cleared**

`src/nemo_evaluator/solvers/harbor.py:403-436` (written), `~2086` (read), `2128-2130` (timeout branch)

The retry_mixin patch writes the marker from tenacity's log_retry_attempt — i.e. predominantly for TRANSIENT errors whose retry then succeeds — and never clears it. At NEL timeout, `elif agent_timed_out and not response and last_llm_error:` sets `error`, so eval_loop skips verification and grades 0. Two consequences, same root cause:

(a) a rollout that hit one transient 429 early, ran clean for hours, then timed out with empty final text and no git-visible workspace diff is forced to reward 0 with "timed out after model API error: \<hours-old error\>" — on main this took the SOLVE_TIMEOUT verify branch, and on state-verified benchmarks (terminal-bench: success is container state, not a /testbed git diff) the verifier could still have awarded reward;

(b) conversely (CodeRabbit's complement), a genuine dead-model zero-progress timeout with a stale marker gets ErrorKind.NONE (graded as model failure) instead of raising InfraError and taking the infra retry path — a rollout retry would have salvaged is permanently lost.

**Fix:** make the marker evidence-of-current-failure — timestamp it and honor it only if recent relative to the timeout / last model-traffic activity, or clear it whenever a subsequent LLM call succeeds. Test both consequences.

src/nemo_evaluator/reports/trajectories.py
```diff
@@ -162,6 +163,62 @@ def _failure_error(row: dict[str, Any]) -> str:
     return _truncate_text(scoring_details.get("error") or row.get("model_error") or row.get("error") or "")
 
 
+def _completed_harbor_verification_without_error(row: dict[str, Any]) -> bool:
```
piojanu: **M1. [MED] Duplicated predicate: "completed harbor verification without error"**

`collector.py:43-50` vs `trajectories.py:166-176` — copy-pasted, and both hardcode harbor's scoring method + detail keys inside environment-agnostic layers (tension with AGENTS.md "environments are interchangeable"). Extract ONE shared helper — natural home `observability/failure_classification.py` — and implement the H3 patch/diff guard there once.

src/nemo_evaluator/solvers/harbor.py
```diff
@@ -1967,14 +2100,33 @@ async def solve(
...
             error_kind = ErrorKind.NONE
+            scoring_details: dict[str, Any] = {}
             if agent_error is not None:
                 etype = type(agent_error).__name__
                 if etype in _INFRA_ERROR_NAMES:
                     raise InfraError(f"Agent infrastructure failure: {etype}: {agent_error}") from agent_error
-                error = f"Agent crashed: {etype}: {agent_error}"
-                logger.warning("HarborSolver: %s", error)
+                crash_error = f"Agent crashed: {etype}: {agent_error}"
+                if workspace_diff and _is_turn_budget_exhausted_error(crash_error):
```
piojanu: **M2. [MED] Triplicated turn-budget-with-diff block**

`harbor.py:2109/2119/2215` — the same 5-line shape three times; extracting a helper would have made the H1 bug harder to write. Fold into the H1 fix.

src/nemo_evaluator/observability/failure_classification.py
```diff
@@ -0,0 +1,158 @@
...
+        "context_window_exceeded",
+        "budget_exceeded",
+        "model_timeout",
+        "rate_limit",
+        "server_error",
+        "model_error",
+    }
+)
+
+_STATUS_RE = re.compile(
```
piojanu, line 21: **L1. [LOW] Timeout-vs-server-error taxonomy + status-regex edges**

`src/nemo_evaluator/observability/failure_classification.py`

- **REVIEWER DIRECTIVE (behavior change, not just observation):** classify 504 ALWAYS as `server_error`; reserve `model_timeout` for timeouts detected on the client side. Currently the category splits by message form: status 504 → model_timeout, but BadGatewayError / "gateway timeout" text → server_error.
- `_LEADING_STATUS_RE` matches ANY message starting with a 3-digit 100-599 number — an ID or count at line start is misread as an HTTP status (note: an earlier claim that leading "504 Gateway Time-out" would be MISSED was wrong; it is caught — the defect is the false-positive direction).
- "HTTP/1.1 504" is still missed (`http\s+` doesn't match "http/").

src/nemo_evaluator/observability/failure_classification.py
```diff
@@ -0,0 +1,158 @@
...
+        "budget has been exceeded",
+    ):
+        return "budget_exceeded"
+    if status == 429 or _has_any(
+        text,
+        "ratelimiterror",
+        "routerratelimiterror",
+        "routerratelimiterrorbasic",
+        "ratelimittype",
+        "rate_limit",
```
piojanu, line 96: **L2. [LOW] Spaced rate-limit text falls through**

`failure_classification.py:76-83` — markers include "ratelimiterror", "routerratelimiterror", "rate_limit", but not "rate limit" with a space. Common provider text like "rate limit exceeded" misclassifies unless a 429 digit happens to be present. Add the spaced markers.

tests/test_engine/test_observability.py
```diff
@@ -47,24 +52,136 @@ def test_to_dict_records_error(self):
...
+            ("litellm.APITimeoutError: Request timed out", "model_timeout"),
+            ("litellm.Timeout: request timed out", "model_timeout"),
+            ("litellm.BadGatewayError: MidStreamFallbackError: APIConnectionError", "server_error"),
+            ("litellm.MidStreamFallbackError: No fallback model group found", "server_error"),
+            ("litellm.APIConnectionError: connection reset by peer", "server_error"),
+            ("litellm.APIError: provider returned an unexpected server response", "server_error"),
+            ("litellm.ServiceUnavailableError: service unavailable", "server_error"),
+            ("litellm.InternalServerError: internal server error", "server_error"),
+            ("ServerDisconnectedError: upstream closed", "server_error"),
+            ("litellm.APIResponseValidationError: response was missing choices", "model_error"),
```
piojanu, line 86: **L3. [LOW] Missing fallback-branch test**

No test exercises `classify_model_failure(...) or "model_error"` with UNCLASSIFIABLE text — every model_error case in the suite matches a known pattern. Add e.g. ("Agent crashed: ValueError" → "model_error").

piojanu (commented):
# Verified review findings — PR #1099 (head `aadd5cf0`)

Consolidated review at head `aadd5cf0d1bfc52ba5fff75bb3722e41726db7b4`, base `main` (merge-base `f7136f79`), diff `git diff f7136f79...aadd5cf0`. Tickets: FEP-1131, FEP-1169, FEP-1122, FEP-1132 (Linear).

Each finding is posted as its own inline comment so it can be resolved independently: **H1–H6** (high), **M1–M2** (medium, also the fix vehicles for H1/H3), **L1–L3** (low/minor). Findings were adversarially verified against the head code; severities reflect that this repo's failure-category taxonomy is a first-class deliverable (leaderboard audits consume it) and that reward corruption outranks label corruption.

## Do NOT act on (retracted after verification)

- **"FEP-1122 not resolved / missing interior steps not captured": retracted.** The missing interior steps were wire-side over-counting (dead-sandbox races + timeout-abort retries), addressed by this PR's grace + request_hash persistence, pre-existing session filtering (PR #1088), and the FEP-1095 timeout-propagation fix; the 1500-rollout Kimi rerun shows exact wire = per-step = final-metrics token equality with zero mismatches.
- **Condensation-call accounting:** tracked in FEP-1221, unrelated to this PR.

*Note: CodeRabbit threads whose content is incorporated into these findings (stale marker → H6, leading-status regex → L1, spaced rate-limit → L2, fallback-branch test → L3) are being resolved with a pointer to the superseding finding.*

src/nemo_evaluator/solvers/harbor.py
```diff
@@ -1951,9 +2019,10 @@ async def solve(
             completion_tokens = context.n_output_tokens or recovered["completion_tokens"]
 
             latency_ms = (time.monotonic() - t0) * 1000
+            last_llm_error = _last_llm_error_from_marker(agent_logs_dir) if agent_timed_out else None
 
             # Timeout with zero progress → model is likely dead.
-            if agent_timed_out and not workspace_diff and prompt_tokens + completion_tokens == 0:
+            if agent_timed_out and not workspace_diff and prompt_tokens + completion_tokens == 0 and not last_llm_error:
```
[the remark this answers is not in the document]
piojanu: Incorporated into finding **H6** of the consolidated review (stale `last_llm_error.json` marker — covers both this dead-model/ErrorKind.NONE misroute and the forced-reward-0 consequence, with a recency/clearing fix). Resolving in favor of H6.

src/nemo_evaluator/observability/failure_classification.py
```diff
@@ -0,0 +1,73 @@
...
+    status = _status_int(status_code)
+    if status is None:
+        status = _status_from_text(text)
+    if not text and status is None:
+        return None
+
+    if any(marker in text for marker in ("turn budget", "token budget", "budget exceeded", "budget exhausted")):
+        return "budget_exceeded"
+    if status == 429 or _has_code(text, 429) or "rate_limit" in text or "too many requests" in text:
+        return "rate_limit"
```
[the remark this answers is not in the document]
piojanu: Incorporated into finding **L2** of the consolidated review (spaced "rate limit" markers missing from `failure_classification.py:76-83`). Resolving in favor of L2.

src/nemo_evaluator/observability/failure_classification.py
```diff
@@ -0,0 +1,158 @@
...
+
+    if _has_any(text, "turn budget exhausted", "turn budget exceeded"):
+        return "turn_budget_exhausted"
+    if _has_any(
+        text,
+        "contextwindowexceedederror",
+        "context window exceeded",
+        "context length exceeded",
+        "context length is exceeded",
+        "maximum context length",
```
piojanu, line 76: **M3. [MED] Classifier doesn't recognize the vLLM context-length phrasing that #1101 just taught the recovery layer — self-hosted vLLM context overflows label as `model_error`**

*(Extends the consolidated review H1–H6 / M1–M2 / L1–L3; found while rebasing onto main.)*

`src/nemo_evaluator/observability/failure_classification.py:54-64` (context-window marker tuple)

PR #1101 (on main, merged into this branch by the rebase) converts vLLM's `VLLMValidationError` body — "You passed N input tokens and requested 0 output tokens. However, **the model's context length** is only M tokens" — to `ContextLengthExceededError` so Terminus-2's three-tier summarization fallback recovers in-flight. That fix is recovery-layer and `is_terminus2`-gated; the labeling layer here doesn't know the phrasing. Verified at head `3e28b8f4`:

- `classify_model_failure("litellm.BadRequestError: You passed 171337 input tokens and requested 0 output tokens. However, the model's context length is only 163840 tokens. ...")` → **`model_error`** ❌
- `classify_model_failure("This model maximum context length is 262144 tokens")` → `context_window_exceeded` ✅

LiteLLM wraps the vLLM error as a plain `BadRequestError` (the very reason #1101 exists), so no exception class name rescues it, and status 400 routes to `model_error`.

**When it bites:** anywhere #1101's recovery doesn't apply — OpenHands/SWE-bench agents (the #1101 patch fires only for Terminus-2), or Terminus-2 rollouts where the summarization fallback still fails. Those crashes get labeled `model_error` instead of `context_window_exceeded` — the same mislabel class as FEP-1131's evidence #2, for the deployment shape (self-hosted vLLM) most Nemotron leaderboard runs use. It also feeds H1's scoring note: even once correctly labeled, a context-window crash with a workspace diff is still graded 0.

**Fix (one line + test):** add `"model's context length"` (vLLM 0.16 phrasing; Harbor upstream ≥ v0.18 ships this exact substring per #1101's description) to the context-window marker tuple, and add a test case with the literal vLLM body. Consider also auditing the tuple against LiteLLM's own `is_error_str_context_window_exceeded` list for other provider phrasings.

src/nemo_evaluator/solvers/harbor.py
piojanu: Resolved by 8061b24 + 6a9884c (tests in e06d2e9/6083c39). Both handlers now route through a single `_classify_workspace_failure` helper (src/nemo_evaluator/solvers/harbor.py:1181-1204): the `except InfraError` handler returns what it computes (2525-2529 → 2541-2543), so a turn-budget InfraError with a workspace diff comes back as `error=None` / `SOLVE_TIMEOUT` / `{"error": ..., "error_category": "turn_budget_exhausted"}` instead of the old hardcoded `INFRA`, and the `except GracefulError` handler binds all three names locally (2575-2580) before the return at 2592-2594, eliminating the UnboundLocalError. The gate is also widened beyond turn-budget text via the `is_agent_crash` branch (1200-1202, envelope check at 1176-1178), applied to both try-body crash sources (2428-2434) and to GracefulError crash envelopes (2579) — covering FEP-1131's ContextWindowExceededError-with-diff case — and the M2 triplication is folded into the same helper. `TestSolveFailureRecovery` (tests/test_solvers/test_harbor_solver.py:1313-1416) locks both handlers in end-to-end through `solve()` with a real workspace diff: the previously dead InfraError path (1343-1358), the previously crashing GracefulError path (1328-1340), and the widened crash gate from all three sources (1361-1416). One non-blocking nit: `test_graceful_agent_crash_with_workspace_diff_is_submitted_for_verification` (1361-1371) only asserts `error is None`; also asserting `error_kind`/`scoring_details` there would be cheap.

src/nemo_evaluator/solvers/harbor.py
piojanu: Resolved by 4ee8a5b: `_sandbox_agent_exec_timeout` (src/nemo_evaluator/solvers/harbor.py:97-111) now composes the sandbox deadline as `agent_timeout + response_grace + 10s SIGTERM exec + 30s flush + 1s margin` instead of the flat +30s, and each runtime phase draws from the same constants (grace at harbor.py:2062-2068, SIGTERM `timeout_sec` at 2093, flush wait at 2095-2096). The composed value is plumbed through `agent_exec_timeout` at harbor.py:2232 into the adapter's `default_timeout` (harbor.py:2244, applied in harbor_adapter.py:122), so the sandbox kill now lands after the full flush window rather than cutting it to ~5-14s. The budget is even slightly conservative — it computes grace on the jittered timeout while runtime uses the un-jittered one. b7a0049/478881b lock it in: `_assert_command_deadline_covers_timeout_lifecycle` (tests/test_solvers/test_harbor_solver.py:1421-1436) asserts the full composition through real `solve()` plumbing at both 60s (line 1530) and 10920s-with-jitter (line 1622), and both would fail against the old flat grace. Note `openclaw_helpers.py:227-228` still uses a flat +30s, but OpenClaw has no host-side post-timeout phases carved from that budget, so it's not affected.

src/nemo_evaluator/observability/collector.py
piojanu: Resolved by b798859 (+ tests in d0bd1ec). The predicate is now the single shared helper `completed_harbor_verification_with_workspace_change` in `src/nemo_evaluator/observability/failure_classification.py:57-85`, which keeps the old gates and re-adds the dropped guard: suppression fires only on positive workspace-change evidence — `report.patch_exists is True` (:73-75, plumbed from the verifier report at `environments/harbor.py:972-974`), the `"[workspace modified]"` sentinel (:76-77), or a non-empty `final_metrics.workspace_diff_preview` in the trajectory (:78-85) as the benchmark-agnostic fallback; otherwise `empty_response` stays visible. Both suppression sites delegate to it (`collector.py:84-90`, `trajectories.py:169-180` via :212/:218/:958), which also closes M1's one-shared-helper directive — no other copy of the predicate remains. The exact regression test I asked for (empty final message + no patch + verification ran → label kept and counted) exists at both layers: `tests/test_engine/test_observability.py:287-307` and `tests/test_reports/test_trajectories_report.py:850-873`, with suppression-positive counterparts at `test_observability.py:272-285` and `test_trajectories_report.py:827-847` exercising the fire paths.

src/nemo_evaluator/reports/trajectories.py
piojanu: Resolved. fd15a41 first added the twin-dedup in `_trial_wire_failure`, then ea61246/b798859 replaced it with an outcome gate: `_trial_failure_category` now computes `successful_verified_outcome` from `reward > 0` and skips the wire override entirely for passing trials, including category `""` (src/nemo_evaluator/reports/trajectories.py:213-217). Because `failures_by_category` (trajectories.py:726-728), the per-row `failure_category` (:391), and the enriched rows / `rows_reclassified_from_wire_failure` counter (:952-957) all route through `_trial_failure_category`, every persistence site named here inherits the fix — and the reward gate is broader than twin-dedup for this scope, also covering a genuinely failed trailing call with a distinct `request_hash`. Tests lock it in: `test_verified_harbor_trial_ignores_failed_duplicate_wire_twin` (tests/test_reports/test_trajectories_report.py:741; reward=1.0, parametrized [200,429]/[429,200] with shared `request_hash` — the failed-twin-last variant returned `rate_limit` under the old code), `test_verified_harbor_trial_ignores_distinct_trailing_wire_failure` (:775), and `test_failed_duplicate_wire_twin_reclassifies_unsuccessful_trial` (:801) pinning that reward=0.0 still reclassifies. The assertions cover exactly the outputs I flagged: `failures_by_category`, the enriched jsonl row, and the reclassification counter.

src/nemo_evaluator/observability/collector.py
piojanu: Resolved by 23b096e: `_classify_failure` now early-returns for `scoring_details.error_category == "system"` rows, setting `failure_category = "system"` (src/nemo_evaluator/observability/collector.py:65-67) before the `classify_model_failure` fall-through at collector.py:78-80 — and `"system"` is not a `MODEL_FAILURE_CATEGORIES` member, so harness error text can no longer land in a model bucket. This matches the sole producer of that category (system-retry exhaustion at src/nemo_evaluator/engine/eval_loop.py:588), and the `"graceful"` carve-out is preserved: those rows match no early return and still reach the model classifier. Attribution also survives the report layer — src/nemo_evaluator/reports/trajectories.py:161 surfaces `error_category` directly and the wire override at trajectories.py:216 only fires for `{"", "empty_response", "model_error"}`. a46114e locks it in with `test_system_errors_remain_non_model_failures` (tests/test_engine/test_observability.py:219-235), parametrized over the docker-py 500 repro, `container exited with status 137`, and an unclassifiable message — each would classify as a model category without the fix — plus `test_graceful_turn_budget_error_remains_model_classified` (tests/test_engine/test_observability.py:237-247) guarding the FEP-1131 scope.

src/nemo_evaluator/solvers/harbor.py
piojanu: Resolved by a17b36a, dac3ebe, and e76dcaa. The marker is now evidence-of-current-failure: the retry_mixin patch writes a v2 payload with `written_at`, `request_timeout_seconds`, `retry_after_seconds`, and `successful_tokens` (src/nemo_evaluator/solvers/harbor.py:438), and `_last_llm_error_from_marker` rejects legacy markers (harbor.py:1130-1133), markers superseded by later successful token traffic (harbor.py:1143-1151), and markers older than their `request_timeout + retry_after + 15s` evidence window (harbor.py:1153-1166); solve also clears any persisted marker before agent setup (harbor.py:2251), and the patch's `schema_version: 2` idempotency key (harbor.py:417, 444) upgrades containers persisted with the v1 producer. With a stale marker ignored, a no-diff timeout with token traffic falls through to the SOLVE_TIMEOUT verify branch again (harbor.py:2463-2476) — consequence (a) — and a zero-progress timeout raises InfraError for the rollout-retry path (harbor.py:2403) — consequence (b). Both consequences are locked in end-to-end in tests/test_integration/test_harbor_stale_llm_error.py: a 120s-old marker no longer skips state verification (:210), a stale marker no longer blocks the InfraError retry that salvages the rollout (:275), later successful traffic supersedes a fresh marker (:231), and the negative control confirms a genuinely current marker still classifies as a model failure (:252); unit coverage includes the v2 producer, legacy-marker rejection, the persisted-v1 upgrade path, and the per-solve clear (tests/test_solvers/test_harbor_solver.py:231, 249, 370, 401, 1315). Only residual nit, fine to leave: the age check compares container-written `written_at` against host `time.time()` (harbor.py:1159), so extreme clock skew could mis-age a marker — supersession and the per-solve clear backstop that case.

src/nemo_evaluator/reports/trajectories.py
piojanu: Resolved by b798859 (with tests in d0bd1ec). The two copy-pasted predicates are gone — `git grep` finds no trace of `_completed_verification_without_error` / `_completed_harbor_verification_without_error` — replaced by a single shared helper `completed_harbor_verification_with_workspace_change` in `src/nemo_evaluator/observability/failure_classification.py:57-85`, exactly where I suggested, and it implements the H3 patch/diff guard once (`report.patch_exists` at :73-74, the `[workspace modified]` sentinel at :76, `final_metrics.workspace_diff_preview` at :78-85). Both former sites now delegate: `src/nemo_evaluator/observability/collector.py:84-89` calls it directly, and `src/nemo_evaluator/reports/trajectories.py:169-180` is just a thin row-to-args adapter with no predicate logic of its own. The harbor-specific method/detail keys are now confined to that one module. Locked in from both directions at both call sites: `tests/test_engine/test_observability.py:272,287` (collector) and `tests/test_reports/test_trajectories_report.py:827,850` (trajectories report).

src/nemo_evaluator/solvers/harbor.py
piojanu: Resolved by 8061b24 (refined in 6a9884c). The three copies are folded into a single `_classify_workspace_failure` helper (src/nemo_evaluator/solvers/harbor.py:1181-1204), which now holds the only `turn_budget_exhausted` block (1192-1198). The two try-body copies collapsed into one unified `crash_error` computation feeding a single call (2419-2434), and the InfraError and GracefulError handlers route through the same helper (2525-2529, 2575-2580) — with the InfraError return now actually consuming the computed triple (2541-2543), closing the H1 dead code this duplication enabled. Grep confirms exactly three call sites and no residual copies. Each site is locked in by `TestSolveFailureRecovery`: tests/test_solvers/test_harbor_solver.py:1343 drives an InfraError turn-budget failure with a diff to `SOLVE_TIMEOUT` + `turn_budget_exhausted` (the formerly dead path), 1374/1391 cover both merged crash branches, and 1328/1361 cover the GracefulError site.

src/nemo_evaluator/observability/failure_classification.py
piojanu: Resolved. The taxonomy directive is complete: `status == 504` short-circuits to `server_error` before any text-marker check (src/nemo_evaluator/observability/failure_classification.py:95-96), and commits 1619bed/df3456a/91524d8 hoisted "gateway timeout" / "upstream timeout" / "upstream timed out" into a dedicated `server_error` check at :132-133 that runs before the `model_timeout` block (:134-146), which now holds only client-side timeout markers — the ordering matters because "gateway timeout error" contains the `model_timeout` marker "timeout error". Both regex edges hold at HEAD: `_STATUS_RE` matches `HTTP/1.1 504` via `http(?:/\d+(?:\.\d+)?)?` (:22-23), and `_LEADING_STATUS_RE` only accepts a leading 3-digit code when the trailing text starts with the canonical `HTTPStatus` reason phrase (:41-50), so "429 model calls completed successfully" no longer parses as a status. Tests lock all three in: tests/test_engine/test_observability.py:131-141 (504 as explicit status, `HTTP 504`, leading `504 Gateway Time-out`, and `HTTP/1.1 504`), :143-155 (server timeout phrase family including the ordering-sensitive prefix/suffix forms), :157-165 (leading counts → None), while client-detected timeouts still assert `model_timeout` (:58, :77-78). Commit 73c07fa aligned the report layer end-to-end: the 504 wire row with "Upstream timed out after 5s" now expects `server_error` (tests/test_reports/test_trajectories_report.py:665).

src/nemo_evaluator/observability/failure_classification.py
piojanu: Resolved — turns out this finding was stale against the review baseline: the spaced `"rate limit"` marker was added in ad296e0 (an ancestor of the reviewed commit) and is present at HEAD in the 429/rate-limit branch (src/nemo_evaluator/observability/failure_classification.py:128, branch at lines 121-131), so `"rate limit exceeded"` classifies as `rate_limit` even without a 429 digit. That text matches none of the earlier branches (turn-budget/context-window/budget markers don't fire on it), and the fix-range diff left the rate-limit marker tuple untouched. Locked in by the `spaced-rate-limit` case in `test_provider_failure_phrases_are_classified` (tests/test_engine/test_observability.py:170, added in f8f1f80), green in CI.

tests/test_engine/test_observability.py
piojanu: Resolved by b6d19ab: `test_unclassified_model_error_falls_back_to_model_error` (tests/test_engine/test_observability.py:204-208) records `StepRecord(model_error="Agent crashed: ValueError")` through `ArtifactCollector.record` and asserts `failure_category == "model_error"` — the exact case I asked for. I traced the string through `classify_model_failure` at HEAD: no digits, so neither `_STATUS_RE` nor `_LEADING_STATUS_RE` extracts a status, and it contains none of the marker substrings, so the classifier hits the final `return None` (src/nemo_evaluator/observability/failure_classification.py:205) and the asserted category comes from the `or "model_error"` fallback at src/nemo_evaluator/observability/collector.py:80. Bonus: testing through the collector rather than the classifier directly covers the actual fallback expression site.

src/nemo_evaluator/observability/failure_classification.py
piojanu: Resolved. The one-line fix is in place: `"model's context length"` sits in the context-window marker tuple at `src/nemo_evaluator/observability/failure_classification.py:106`, which is checked before the `badrequesterror`/status-400 → `model_error` routing (lines 171-202), and neither status regex extracts a code from the vLLM body, so the exact repro string now classifies as `context_window_exceeded`. The fix landed in ad296e0 (marker) with the test in f8f1f80 — both already ancestors of 2c4fbce, so this was fixed before the review anchor (the finding was written against the earlier head 3e28b8f4), and the post-review commits didn't touch the tuple. `tests/test_engine/test_observability.py:171-179` (`vllm-context-length`) locks it in with the literal vLLM `BadRequestError` body. Non-blocking follow-up: the suggested audit against LiteLLM's `is_error_str_context_window_exceeded` is incomplete — phrasings like "prompt is too long" / "input is too long" are still absent from the tuple; adding them with matching test params would be a trivial hardening.

piojanu (approved):
Amazing team work! Thanks Claude 😄

piojanu:
**Post-merge confirmation run: MERGE-SAFE.** The kill/resume mini run reran on the exact merged tree (`a422daac`, image `dev-20260717-143853-a422daac-harbor`) with a config identical to the pre-merge run — only the image differs. Run `20260717_132222_c4e5cba6` ([MLflow](https://mlflow-nemo-evaluator.nvidia.com/#/experiments/1550/runs/2b3e66b2ec704733bcbe571620fa7659); artifacts `oci-hsg:/lustre/fsw/portfolios/llmservice/users/pjanuszewski/nemo-evaluator-rundirs/nano_v35/fep1295-mini-c2-merge-verify/20260717_132222_c4e5cba6`).

**FEP-1295 battery: 7/7 PASS, now across 5 real walltime kills** (6-job chain): 100/100 oversized trajectories spilled to sidecars and survived with **0 zero-step rows**; 11 rollouts replayed across kill boundaries; all 5 kill-crossing `solve_timeout` labels retained; 100/100 capture markers with 0 false flags.

**Conflict-resolution / #1103: no regression.** Both patch blocks kept in the merge resolution execute on every rollout — `tool_concurrency_limit: patched` and `retry_error_capture=patched` each fire 144/144 (the pre-merge run shows 0 tool-concurrency lines — clean differential). No duplicated agent steps, no step/token inflation (+0.3% steps, +2.4% tokens). Honest caveat: the model issued **no parallel tool calls in 200 rollouts**, so #1103's merging path itself was never exercised end-to-end here — its coverage rests on the patch-application evidence and unit tests.

**Score: 0.40 vs 0.49 pre-merge — not significant** (paired per-problem, 17 vs 8 discordant, exact McNemar p=0.108). The flips decompose mechanically:
- 9 flips (both directions, net +3 *toward* the post-merge run) trace to a **pre-existing** defect present in both runs: on verify-only resume, the EFS workspace transfer creates a fresh session dir, so the pre-kill `workspace.tar` is never reattached and the replayed verify scores a pristine workspace — replayed rollouts pass 1/20 vs ~45% base rate (p≈2×10⁻⁴). Smoking gun: a replay's `/input/workspace.tar` symlink timestamps to the resume second, not the capture. Root cause posted on the existing ticket FEP-888 (this is its zero-stub mechanism); not introduced by this PR, and the new capture markers are what made it measurable.
- ~6 flips: more 600 s timeouts (43 vs 31) at +25% walltime with unchanged tokens/steps — environment ran slower that day; watch item, not attributable to the merge.
- ~10 flips: ordinary SWE-bench single-repeat variance.

PR #1101, opened 2026-07-08
fix(harbor): recognize vLLM "model's context length" phrasing

iguehring-nvidia:
## What

Extend `harbor.llms.lite_llm.LiteLLM._is_context_length_error` at `HarborSolver._create_agent`-time so that the `VLLMValidationError` body raised by self-hosted vLLM deployments — literally:

> `"You passed N input tokens and requested 0 output tokens. However, the model's context length is only M tokens..."`

— is classified as a context-length overflow, which lets Harbor's existing `_handle_litellm_error` translate the wire error to `harbor.llms.base.ContextLengthExceededError` instead of leaving it as a plain `LiteLLMBadRequestError`. Terminus-2's `_query_llm::except ContextLengthExceededError` block then fires its three-tier reactive-summarization fallback and the agent continues, instead of hard-crashing with `Agent crashed`.

## Why

Harbor 0.3.x's `_is_context_length_error` phrases tuple lists `"maximum context length"` but not `"model's context length"`, so the vLLM body currently falls through unrecognized. Chain:

- vLLM (self-hosted) raises `VLLMValidationError` with the body above.
- LiteLLM classifies as a plain `BadRequestError` (its own `is_error_str_context_window_exceeded` substring list does not include this phrasing either).
- Harbor's own catch-net misses it too.
- `terminus_2.py::_query_llm` only catches `ContextLengthExceededError`; the `BadRequestError` retries 3× and then propagates → OpenHands reports `Agent crashed` → task hard-zeros.

The three-tier summarization fallback in Terminus-2 (`_summarize` → short summary → screen-only prompt at `terminus_2.py:1017-1094`) already exists and is correct — it just never triggers because the exception type never converts.

## How

- New module-level tuple `_VLLM_CTXLIM_PHRASES = ("model's context length",)`.
- New `_patch_harbor_lite_llm_context_length_matcher()` runtime patch, structured like the existing `_patch_terminus_cle_reset` / `_patch_terminus_unwind_min_pairs` / `_patch_terminus_api_token_anchor` functions. Wraps Harbor's original matcher additively (`if original(self, error): return True`) so existing coverage is unchanged.
- Wired into `HarborSolver._create_agent` alongside the other Terminus-2 patches, gated on `is_terminus2`. Idempotent via `_HARBOR_LITELLM_CTXLIM_PATCHED`.
- No new dependencies. No harbor pin bump. No behavior change for non-Terminus-2 agents or for non-vLLM overflows.

## Relationship to a harbor pin bump

Harbor v0.18.0 already ships this exact substring in its own phrases tuple upstream (with the comment `# vllm 0.16.0 error message`). This runtime patch is a **bridge** that becomes a strict no-op the moment the `harbor>=0.3.0,<0.4.0` pin in `pyproject.toml` is bumped to a release containing it. That pin bump is a separate change and can delete this patch cleanly once it lands.

## Testing

```bash
uv run pytest tests/test_solvers/test_harbor_terminus_patches.py -q
# 36 passed
```

New `TestHarborLiteLLMContextLengthMatcher` covers the actual vLLM body attribute layout (message / body / plain str), continued support of the original OpenAI-style `"context length exceeded"` phrasing, non-matches on unrelated errors and near-miss phrasings, and patch idempotency. `TestCreateAgentGating` extended to assert the new patch fires on Terminus-2 and does not fire on non-Terminus-2 agents.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved “context length exceeded” error classification for Terminus-2 across different vLLM/LiteLLM error shapes and message fields.
  * Applied the enhancement safely during agent setup with idempotent one-time installation.
  * Added warning behavior when a related token-length drift is detected without a matching classifier phrase.
* **Tests**
  * Strengthened Harbor/Terminus-2 patch gating coverage to verify correct conditional installation.
  * Added matcher tests for multiple error representations, log assertions, and idempotency.

### DIFF
--- a/src/nemo_evaluator/solvers/harbor.py
+++ b/src/nemo_evaluator/solvers/harbor.py
@@ -1295,6 +1295,7 @@ def _patch_terminus_unwind_min_pairs() -> None:
 
 _TERMINUS_API_ANCHOR_PATCHED = False
 _CHAT_TOKEN_ANCHOR_PATCHED = False
+_HARBOR_LITELLM_CTXLIM_PATCHED = False
 
 
 def _terminus2_count_total_tokens(self, chat) -> int:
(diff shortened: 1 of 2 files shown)


mlazuka (approved):
lgtm

src/nemo_evaluator/solvers/harbor.py
```diff
@@ -1374,6 +1375,54 @@ def _patch_terminus_api_token_anchor() -> None:
...
+# Harbor 0.3.x's ``LiteLLM._is_context_length_error`` phrases tuple lists
+# ``"maximum context length"`` but NOT ``"model's context length"``, so this
+# text falls through to a plain ``LiteLLMBadRequestError``, Terminus-2's
+# ``except ContextLengthExceededError`` block never fires, and the agent
+# crashes instead of running its reactive-summarization fallback.
+#
+# Harbor v0.18.0 added exactly this substring upstream (see
+# ``harbor/llms/lite_llm.py`` with the comment ``# vllm 0.16.0 error
+# message``), so this patch becomes redundant the moment the ``harbor`` pin
+# in ``pyproject.toml`` is bumped to a release containing it.
```
Glorf, line 1394: so why won't we just update harbor?

Glorf (approved):
lgtm, but consider just updating harbor and running a short repro afterwards

src/nemo_evaluator/solvers/harbor.py
iguehring-nvidia: Long term I think updating harbor is the right move. But it would require thorough testing for all impacted benchmarks. This fix is needed for today.

src/nemo_evaluator/solvers/harbor.py
```diff
@@ -1374,6 +1375,75 @@ def _patch_terminus_api_token_anchor() -> None:
...
+            str(getattr(error, "body", "") or ""),
+            str(getattr(error, "message", "") or ""),
+            str(getattr(error, "error", "") or ""),
+        ]
+        combined = " ".join(p for p in parts if p).lower()
+        if any(phrase in combined for phrase in _VLLM_CTXLIM_PHRASES):
+            return True
+        if _VLLM_CTXLIM_DRIFT_MARKER in combined:
+            logger.warning(
+                "BadRequestError body contains vLLM's stable ctx-overflow marker "
```
iguehring-nvidia, line 1433: @agronskiy thanks for the suggestion! Added here.

agronskiy: Cool!

PR #1102, opened 2026-07-09
feat: add sandbox.scrub_git_history anti-cheat for SWE stateless evals

iguehring-nvidia:
## Summary
SWE-bench-style task images ship the repository with the gold fix and hidden tests present in the git history (as commits descending from the base commit). An agent can read them with `git log --all` / `git show <sha>` and copy the solution instead of solving the issue. This adds a config-toggled anti-cheat that scrubs the repo's future git history in the agent sandbox before the agent starts.

## How it works
- New sandbox config field `scrub_git_history` (default off).
- New module `sandbox/git_scrub.py`: `build_git_history_scrub_cmd(workdir)` returns a self-enforcing shell snippet that makes every non-ancestor commit unreachable, physically prunes the objects, emits a single `NEL_GIT_CLEANUP` audit line, and exits non-zero if any non-ancestor commit survived.
- Generic `SeedResult.pre_agent_cmd` / `LifecycleContext.pre_agent_cmd` seam: `StatelessSandbox` runs it in the agent container after acquire, best-effort (a broken step never aborts the rollout). Mirrors the existing `capture_cmd` / `apply_cmd` threading.
- `pick_lifecycle` builds the scrub command from the resolved agent workdir when the flag is set; otherwise honors an env-supplied `seed.pre_agent_cmd`.
- **StatelessSandbox only.** In stateful mode solve and verify share one container, so scrubbing could corrupt scoring; setting `stateful: true` together with `scrub_git_history: true` emits a warning and skips the scrub. The working tree is left untouched and verification runs in a fresh container, so scoring is unaffected.

## Usage
```yaml
benchmarks:
  - name: harbor://swebench_multilingual@1.0   # any harbor SWE benchmark
    sandbox:
      type: ecs_fargate
      scrub_git_history: true   # StatelessSandbox only
```
Note: `scrub_git_history` is validated at submit time (the sandbox config model is `extra="forbid"`), so the `nel` that submits the run must include this change in addition to the eval image.

## Testing
- `pytest tests/ -m "not network"` — new `test_git_scrub.py` runs the scrub against throwaway git repos and asserts the future commit becomes unreachable (`git log --all`) and is physically pruned (`git cat-file -e` fails), and that skip-cases (missing workdir / not a git repo / no HEAD) exit 0; lifecycle tests assert the command runs when flagged, is absent otherwise, and a non-zero scrub does not abort the rollout; a config test asserts that `stateful` + `scrub_git_history` warns.
- Validated end-to-end on a SWE-bench Multilingual oracle run: every rollout logged the scrub audit marker with `non_ancestor_left=0`, and grading was unaffected.

## Summary by CodeRabbit

* **New Features**
  * Added an optional sandbox setting to scrub Git “future history” before agent execution for supported stateless tasks.
  * Added support for a pre-agent command in sandbox lifecycles, including automatic Git history cleanup when enabled.
* **Bug Fixes**
  * Emit a warning when history cleanup is enabled in a mode where it’s ignored.
  * Improved failure behavior so pre-agent cleanup commands fail safely without aborting execution.
* **Tests**
  * Added integration-style coverage for Git history scrubbing, including skip behavior and paths with spaces.

### DIFF
--- a/src/nemo_evaluator/config/sandboxes.py
+++ b/src/nemo_evaluator/config/sandboxes.py
@@ -47,6 +47,10 @@ class _SandboxBase(BaseModel):
     capture_cmd: str | None = None
     verify_timeout: float = 600.0
     stateful: bool = False
+    scrub_git_history: bool = False
+    """Run the git future-history scrub in the agent sandbox before the agent
+    starts (anti-cheat for SWE-style tasks whose gold fix / hidden tests live in
+    the repo's future commits). Only applies to StatelessSandbox tasks."""
 
     @model_validator(mode="after")
     def _warn_stateful_with_capture_cmd(self) -> "_SandboxBase":
@@ -54,6 +58,18 @@ class _SandboxBase(BaseModel):
             warnings.warn(
                 "sandbox.stateful=True ignores capture_cmd — the capture/transfer "
                 "workflow is skipped when stateful mode is enabled.",
+                UserWarning,
+                stacklevel=2,
+            )
+        return self
+
+    @model_validator(mode="after")
+    def _warn_stateful_with_scrub_git_history(self) -> "_SandboxBase":
+        if self.stateful and self.scrub_git_history:
+            warnings.warn(
+                "sandbox.stateful=True ignores scrub_git_history — the git-history "
+                "scrub only runs for StatelessSandbox tasks (solve and verify share "
+                "one container in stateful mode, so scrubbing could corrupt scoring).",
                 UserWarning,
                 stacklevel=2,
             )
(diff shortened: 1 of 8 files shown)


Glorf (approved):
It's a little bit too much swe-centric imo, but I think it's ok for now.

iguehring-nvidia:
> It's a little bit too much swe-centric imo, but I think it's ok for now.

@Glorf Another option I was considering: Just introducing the `pre_agent_cmd` hook and let the bash command that handles the swe specific live in the configurations instead of in the evaluator package. That would introduce the general mechanism of enabling some command to be run before agent gets the task and the benchmark specific logic would be factored out into the config. WDYT? Would that address your concern?

PR #1107, opened 2026-07-09
fix(xstest): update CSV URL and rename ID field for consistency

cris96spa:
## Problem
The upstream dataset changed in two ways:
- File renamed from `xstest_v2_prompts.csv` to `xstest_prompts.csv`.
- Column renamed from `id_v2` to `id`.

Fixing only the file path still leaves a parser failure on the missing `id_v2` column.

## Change
- Replace `xstest_v2_prompts.csv` with `xstest_prompts.csv`.
- Replace `int(row["id_v2"])` with `int(row["id"])`.

## Impact
- Restores full XSTest ingestion.
- Preserves prompt content and safe/unsafe label handling.

## Validation
- Verified CSV loads successfully from the new path.
- Verified parser completes with the updated `id` column.
- Run standard checks:
  - `make test`
  - `make lint`

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the default benchmark prompt source to a newer CSV, improving the accuracy of loaded prompts.
  * Fixed row parsing to use the correct prompt ID field, helping benchmark records normalize correctly.

### DIFF
--- a/src/nemo_evaluator/benchmarks/xstest.py
+++ b/src/nemo_evaluator/benchmarks/xstest.py
@@ -34,7 +34,7 @@ import urllib.request
 from nemo_evaluator.environments.custom import benchmark, scorer
 from nemo_evaluator.scoring import ScorerInput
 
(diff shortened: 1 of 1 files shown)


cris96spa:
Closes #1110

PR #1108, opened 2026-07-09
fix(healthbench): update dataset loading URL to correct path

cris96spa:
## Problem
HealthBench loading fails because:
- The configured `split="test"` path does not resolve for this dataset layout.
- The fallback URL points to `data/test.jsonl`, which is not present in the repo.

## Change
- Update fallback path to `resolve/main/2025-05-07-06-14-12_oss_eval.jsonl`.

## Impact
- Restores HealthBench fallback ingestion.
- Maintains existing seeding expectations (`prompt` messages and `rubrics`) without changing scoring behavior.

## Validation
- Verified the referenced JSONL exists and is readable.
- Verified row shape matches current seeder expectations.
- Run standard checks:
  - `make test`
  - `make lint`

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the HealthBench test dataset source to use the latest available test content, ensuring benchmark results reflect the intended evaluation data.

PR #1112, opened 2026-07-10
fix: use available HealthBench fallback dataset

### LINKED ISSUE #1109: HealthBench fallback URL 404
**Describe the bug**

HealthBench loading fails because the configured dataset path does not resolve for the current dataset layout.

Specifically, the configured `split="test"` path is unavailable, and the fallback URL currently points to `data/test.jsonl`, which is not present in the repository.

**Steps/Code to reproduce bug**

```python
from nemo_evaluator.benchmarks.healthbench import _load_healthbench

if __name__ == "__main__":
    _load_healthbench()  # raises urllib.error.HTTPError: HTTP Error 404: Not Found

```

That file is not present in the HealthBench repository, so fallback ingestion fails.

**Expected behavior**

HealthBench should load successfully using a valid fallback JSONL source.

The fallback source should point to the available evaluation file:

```text
resolve/main/2025-05-07-06-14-12_oss_eval.jsonl
```

**Additional context**

A proposed fix updates the fallback path to `resolve/main/2025-05-07-06-14-12_oss_eval.jsonl`.

Cryptoteep:
## What changed

- update the HealthBench fallback URL to the available `2025-05-07-06-14-12_oss_eval.jsonl` dataset file
- add an offline regression test that forces the primary loader to fail and verifies the fallback URL, timeout, and JSONL parsing

## Why

The configured `test` split can be unavailable, and the previous fallback pointed to `data/test.jsonl`, which does not exist and returned HTTP 404. The fallback now resolves to the published evaluation file.

## Validation

- focused fallback test passes
- `ruff check` passes for both modified files
- `ruff format --check` passes for both modified files
- the required full offline suite collected 415 tests but is blocked on Windows during collection by the existing Unix-only `fcntl` import in `engine/checkpoint.py`

Closes #1109

### DIFF
--- a/src/nemo_evaluator/benchmarks/healthbench.py
+++ b/src/nemo_evaluator/benchmarks/healthbench.py
@@ -31,7 +31,7 @@ def _load_healthbench():
     except Exception:
         pass
 
-    url = "https://huggingface.co/datasets/openai/HealthBench/resolve/main/data/test.jsonl"
+    url = "https://huggingface.co/datasets/openai/HealthBench/resolve/main/2025-05-07-06-14-12_oss_eval.jsonl"
     with urllib.request.urlopen(url, timeout=60) as resp:
         text = resp.read().decode()
     rows = [_json.loads(line) for line in text.strip().splitlines() if line.strip()]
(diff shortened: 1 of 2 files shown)


cris96spa:
Replicated #1108

Cryptoteep:
Closing this because PR #1108 already implements the same HealthBench fallback URL fix. It did not reference #1109, so my initial issue-number search missed the overlap. Sorry for the duplicate.

PR #1113, opened 2026-07-13
fix: native skills aime26

trantrieuvy:
## Summary by CodeRabbit

* **Bug Fixes**
  * Updated math and multiple-choice scoring to use the current grading interface.
  * Improved numeric answer matching for signed values, decimals, formatted numbers, and units.
  * Resolved service request timeouts more consistently, including tool-calling workflows.

* **New Features**
  * Added support for custom `scorer:` metrics in gate validation.

* **Tests**
  * Added coverage for updated grading behavior, numeric scoring, custom scorers, and timeout-related evaluation paths.

### DIFF
--- a/examples/configs/24_aime26_skills_sglang.yaml
+++ b/examples/configs/24_aime26_skills_sglang.yaml
@@ -1,0 +1,53 @@
+# Flavor: Native NeMo Skills AIME 2026 against an external SGLang endpoint.
+# This is an integration smoke configuration, not a model-card reproduction:
+# Qwen thinking is disabled and the solver uses a concise boxed-answer prompt.
+#
+# Prerequisites:
+#   - pip install nemo-evaluator[skills]
+#   - An OpenAI-compatible endpoint serving the configured model
+#   - Network access on first run to prepare MathArena/aime_2026
+#
+# Defaults target a local port-forward:
+#   kubectl port-forward service/qwen36-native-sglang 30000:30000
+#
+# Run the two-problem smoke test:
+#   nel eval run examples/configs/24_aime26_skills_sglang.yaml
+#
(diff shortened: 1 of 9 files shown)


shaurya416:
One interaction to handle before this merges: `src/nemo_evaluator/benchmarks/mgsm.py:45` (not in this PR's file list) strips the target at load time.

```python
answer = str(row["answer_number"]).rstrip("0").rstrip(".")
```

`answer_number` is `int32` in `juletxara/mgsm`, so this removes real integer digits: 20 becomes `"2"`, 180 becomes `"18"`. At `887fa17` that is masked, because the current `numeric_match` strips both sides the same way. With this PR's `float()` comparison it stops being masked. I loaded the real `mgsm.py` and `pattern.py` (decorators, `SeedResult` and `load_dataset` stubbed) with this PR's `pattern.py` applied, for an item whose answer is 20:

| response | HEAD | with #1113 | expected |
|---|---|---|---|
| `Answer: 20` | True | **False** | True |
| `Answer: 2` | True | **True** | False |
| `Answer: 200` | True | False | False |

In the English test split, 99 of 250 answers end in 0 (the other language configs share the same `answer_number` values). So merging this alone would mark every correct answer on those items wrong and credit the answer that is 10x too small, and MGSM accuracy would drop by up to about 40 points.

Fix, here or in a follow-up that lands with it: `answer = str(row["answer_number"])` at `mgsm.py:45`, plus a test that runs the MGSM loader with a stubbed row `answer_number=20` and checks that the scorer accepts `Answer: 20` and rejects `Answer: 2` and `Answer: 200`. The existing `TestNumericMatch` cases (`5.00` vs `5`, `15` vs `15`) pass under both implementations, so they would not catch it.

PR #1116, opened 2026-07-16
fix(engine): spill oversized checkpoint trajectories to sidecar and preserve error labels across resume

piojanu:
Stacked on #1114 (base: `janusz/fep-1132-empty-response-observational-flag`), which stacks on #1099 — review only the top commit here. Intended merge order: this PR → #1114 → #1099.

Fixes [FEP-1295](https://linear.app/nvidia/issue/FEP-1295/resume-checkpoint-silently-drops-5mb-trajectories-steplog-cap-and).

## Problem

Found during PR #1099 pre-merge verification (FEA-244): 11 SWE-bench Multilingual rollouts ended with zero agent steps despite the FEP-1169 timeout flush demonstrably succeeding.

1. `StepLog.append` silently rewrote any checkpoint record whose trajectory JSON exceeded 5 MB as `{"trajectory": null, "_truncated": true}`. Harmless while the shard lives (final outputs come from memory) — but after a SLURM walltime kill the checkpoint is the **only** copy, and it's systematically the 3-hour-timeout rollouts (biggest trajectories, kill-adjacent by construction) that lose up to 3 h of model work.
2. The inference record stored no `error`/`error_kind`/solver `scoring_details`, so verify-only resume re-verified solve failures and structurally could not re-derive labels: `failure_category=solve_timeout` (live-path-only flag) and `turn_budget_exhausted` (rides in solver scoring_details) vanished across kills.
3. A kill landing before/during workspace capture made resume silently verify an unmodified workspace — the one non-reward-neutral window, previously undetectable.

## Change

- **Sidecar spill instead of dropping**: oversized trajectories go to `trajectory_overflow/p{i}_r{r}.json.gz` (atomic tmp+fsync+rename) with a WARNING; the record keeps a `trajectory_ref` pointer. Cap parametrizable per `StepLog` and via `NEL_MAX_TRAJECTORY_BYTES`. Spill failure degrades to the legacy `_truncated` behavior with an ERROR, never crashes the run.
- **Resume dereferences sidecars** on both the fully-cached and verify-only paths (`StepLog.resolve_trajectory`); legacy `_truncated` records warn instead of passing silently.
- **Error labels survive resume**: inference records persist `error`, `error_kind`, `sandbox_used`, and solver `scoring_details`; verify-only resume replays solve failures (no bogus re-verify) and restores `solve_timeout`/infra classification and solver-attached labels.
- **Unverified infra-kind records are dropped at resume load** so the step re-solves — mirrors the existing verified-cache infra retry policy.
- **Kill-during-capture is detectable**: a capture marker is written after `transition_to_verify`; verify-only resume of a `sandbox_used` record without a marker logs a WARNING and stamps `scoring_details.resume_workspace_capture="missing"`.

Old checkpoints stay readable (all new keys optional); records written by pre-fix code get the trajectory-loss warning but not capture detection.

## Verification

- Full offline suite: 2241 passed, 35 skipped (was 2219 at base — +22 new tests, zero regressions); `ruff check`/`ruff format --check` clean.
- New unit tests (`tests/test_engine/test_step_log.py`) cover spill round-trip, env/ctor cap precedence, spill-failure fallback, legacy `_truncated`, compaction, markers.
- New integration tests drive `run_evaluation` through kill-shaped resume: oversized trajectories survive fully-cached and verify-only resume; `solve_timeout` / `turn_budget_exhausted` labels survive; solve failures replay without re-verify; unverified infra records re-solve; capture-marker flagging.
- 8/9 integration tests verified red on pre-fix code with the exact FEP-1295 symptoms.
- FEA-244-style forced-failure mini run (tiny `NEL_MAX_TRAJECTORY_BYTES` + `run_timeout` 600 s + short-walltime kill/resume chain on a dev image built from this branch) to follow; evidence will be posted here and on FEP-1295.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -86,7 +86,6 @@ Before committing, always:
 1. **Lint**: Run linter checks on all modified files. Fix any errors you introduced. Do not fix pre-existing lints unless they are in code you modified.
 2. **Test**: Run `python -m pytest tests/ -v --tb=short -m "not network" -x`. All tests must pass. If a test fails because of your change, fix it. If a test fails for an unrelated reason, note it but do not silently skip it.
 3. **Verify imports**: If you moved or renamed a module, grep for the old import path across the entire repo. Update all references.
-4. **CHANGELOG**: For user-visible changes, add an entry under the current version in `CHANGELOG.md`.
 
 Only after all checks pass, create the commit.
 
--- a/src/nemo_evaluator/engine/eval_loop.py
+++ b/src/nemo_evaluator/engine/eval_loop.py
@@ -33,6 +33,9 @@ from nemo_evaluator.engine.step_log import (
     VERIFIED_LOG,
     StepLog,
     config_hash,
+    has_capture_marker,
+    reset_checkpoint_sidecars,
+    write_capture_marker,
 )
 from nemo_evaluator.environments.base import EvalEnvironment, VerifyResult
 from nemo_evaluator.errors import GracefulError, InfraError
@@ -190,6 +193,7 @@ async def run_evaluation(
                 if verified_cache:
                     verified_log.compact(verified_cache)
                 verified_log.open()
+                reset_checkpoint_sidecars(step_log_dir)
                 inference_log.open(truncate=True)
                 model_traffic_log.open(truncate=True)
                 inference_log.write_meta({"config_hash": cfg_hash})
@@ -201,6 +205,16 @@ async def run_evaluation(
                 inferred_cache = {k: v for k, v in inferred_cache_raw.items() if k not in infra_keys}
                 if infra_keys:
                     logger.info("resume: %d infra-error entries will be retried", len(infra_keys))
+                infra_inferred = {
+                    k
+                    for k, v in inferred_cache.items()
+                    if k not in verified_cache and v.get("error_kind") == ErrorKind.INFRA.value
+                }
+                inferred_cache = {k: v for k, v in inferred_cache.items() if k not in infra_inferred}
+                if infra_inferred:
+                    logger.info(
+                        "resume: %d unverified infra-error inference entries will be re-solved", len(infra_inferred)
+                    )
                 if inferred_cache:
                     meta = old_meta or {"config_hash": cfg_hash}
                     inference_log.compact(inferred_cache, meta=meta)
@@ -215,6 +229,7 @@ async def run_evaluation(
             if n_from_cache or n_verify_only:
                 logger.info("resume: %d fully cached, %d verify-only, rest from scratch", n_from_cache, n_verify_only)
         else:
+            reset_checkpoint_sidecars(step_log_dir)
             inference_log.open(truncate=True)
             inference_log.write_meta({"config_hash": cfg_hash})
             verified_log.open(truncate=True)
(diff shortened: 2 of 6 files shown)


piojanu:
**Local kill→resume drill (commit 21605188): fix verified end-to-end; pre-fix control reproduces the bug.**

Method: real `SIGKILL -9` of a live `run_evaluation` process mid-run, then a real `resume=True` rerun — identical kill profile on the fixed tree and on pre-fix `0e0bbd19` (30 problems, ~200 KB trajectories, 50 KB cap, kill at t+13.7 s with 13 rollouts inference-checkpointed / 9 verified / 4 in-flight, incl. one killed inside `transition_to_verify` before the capture marker).

| Assertion | Fix | Pre-fix control |
|---|---|---|
| Sidecars + `trajectory_ref` in checkpoint | 13/13 rows, 0 `_truncated` | 13/13 `_truncated`, 0 sidecars |
| Zero-step rows after resume | **0/30**; 13 sidecar snapshots byte-equal final trajectories (7 fully-cached + 6 verify-only) | **13** (exactly the pre-kill rollouts) |
| `solve_timeout` across kill (unverified at kill) | preserved (incl. `solve_scoring_details` rider) | `null` |
| Solve failures replayed, not re-run | errors verbatim, reward 0.0, `graceful`/`turn_budget_exhausted` kept; 0 re-solves of pre-kill rollouts; no re-verify of failed solves | errors lost, `env.verify` re-called |
| Kill-before-capture detection | WARNING + `resume_workspace_capture="missing"` on the affected rollout only | n/a (silent) |

Caveat: sandbox path driven via a `pick_lifecycle` stub (repo-test pattern), not a real Docker sandbox — the real-sandbox leg comes with the FEA-244-style cluster mini run on the dev image. Independent review returned one confirmed major on the capture-marker lifecycle (stale-marker invalidation) — revision incoming before the mini run.

piojanu:
**Independent review: APPROVE** (two rounds, adversarial).

Round 1 on `21605188` returned REQUEST-CHANGES with one confirmed major: capture markers were never invalidated when a (problem, repeat) was re-solved (infra-drop resume / in-run verify retry), falsely suppressing the new `resume_workspace_capture` flag in a subset of its target scenarios. Round 2 on `5d3424df` confirmed all findings resolved:

- **M1** — `clear_capture_marker()` before every live inference append; reviewer independently red-verified the regression test and traced both failure scenarios. The only residual window (kill between clear and append) produces a fail-safe *false positive* flag, never a false negative.
- **m1** — markers now written only for workspace-capturing lifecycles (`captures_workspace` attr: StatelessSandbox `True`, StatefulSandbox/NoSandbox `False`); stateful verify-only resume now always flags `missing`, which is truthful (verify runs in a fresh sandbox there).
- **m2/n2/n4** — spill `compresslevel=1`; byte-accurate `trajectory_bytes`; honest marker docstring. **m4** — tests now pin both `reset_checkpoint_sidecars` call sites and sidecar overwrite on re-append.

Suite on head: 2247 passed / 35 skipped locally and on the GitLab mirror (pipeline 58307901, which also built the verification image `dev-20260716-180604-5d3424df-harbor`).

Cluster mini run (FEA-244 style) now in flight: SWE-bench Verified 100×1 on the dev image with `NEL_MAX_TRAJECTORY_BYTES=65536`, `run_timeout=600`/`override`, single shard, 30-min walltime → real SLURM TIMEOUT kill/resume chain. 1-problem smoke submitted (HSG job 5434196); full run follows on smoke pass. Evidence lands here + FEP-1295.

PR #1123, opened 2026-07-21
fix: make model traffic flushing OOM-safe

AWarno:
## Summary
- make completed model-traffic records leave process memory by spooling drained session data to disk
- keep evaluator-loop semantics and public helper names unchanged
- stream final `model_traffic.jsonl` appends from the drained session

## Validation
- Output parity against `origin/main`: deterministic 4-row case covering 200 JSON, 400 JSON error, client-side timeout/error, and 200 SSE produced byte-identical `model_traffic.jsonl`. SHA-256: `acd48628e6700f555962b81453ed32cecd4d6d23cb6271a16c0bfd1a5db0fb41`. Artifacts: `/tmp/model-traffic-before-after-28FzCr`.
- Capture+flush stress: 5,000 completed calls, 20,000 captured request chars/call; output row count/stats/size stayed identical (`5,000` rows, `99.35 MB`, `60,000` total tokens). Retained RSS after capture dropped from `168.55 MB` to `63.41 MB`; total measured Python time changed from `0.537s` to `0.880s`. Artifacts: `/tmp/model-traffic-mem-wall-91zgfb`.
- Capture-only OOM path: 10,000 completed calls, 20,000 captured request chars/call. `origin/main` retained RSS grew by `200.25 MB` (`68.36 -> 268.61 MB`); this PR stayed flat (`62.50 -> 62.50 MB`) and moved `196.47 MB` to disk. Capture wall time changed from `0.623s` to `0.965s`. Artifacts: `/tmp/model-traffic-capture-mem-UcWUQl`.
- `uv run --python 3.12 --extra dev python -m pytest tests/test_engine/test_model_traffic_capture.py tests/test_adapters/test_interceptors_extended.py -q` (`38 passed`)
- `ruff check` and `ruff format --check` passed for changed files

GitLab MR: https://gitlab-master.nvidia.com/dl/JoC/competitive_evaluation/nemo-evaluator-next/-/merge_requests/232

## Summary by CodeRabbit

* **New Features**
  * Model traffic is now spooled to disk as append-only JSONL and drained for processing, enabling better handling of large sessions.
  * Added disk-backed drained sessions with lazy iteration, on-demand stats, and automatic cleanup.

* **Bug Fixes**
  * Updated traffic-log formatting and aggregation to accept streaming/iterable inputs while preserving correct per-row fields (status, usage, tool-call details), including error cases.

* **Tests**
  * Added end-to-end coverage for spooling/flush, drained lifecycle and reuse/rejection behavior, permission semantics, cleanup without iteration, and updated formatting assertions.

### DIFF
--- a/src/nemo_evaluator/engine/model_traffic_log.py
+++ b/src/nemo_evaluator/engine/model_traffic_log.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 import time
 from datetime import datetime, timezone
-from typing import Any
+from typing import Any, Iterable, Iterator
 
 from nemo_evaluator.engine.step_log import StepLog
 from nemo_evaluator.observability.model_traffic import ModelTrafficStore
@@ -37,7 +37,7 @@ def _timestamp(value: Any) -> str:
 def drain_model_traffic_session(
     model_traffic_store: ModelTrafficStore | None,
     session_id: str | None,
-) -> list[dict[str, Any]]:
+) -> Iterable[dict[str, Any]]:
     """Return and clear traffic records for one Adapter Session."""
     if model_traffic_store is None or not session_id:
         return []
@@ -46,7 +46,7 @@ def drain_model_traffic_session(
 
 async def append_model_traffic_records(
     model_traffic_log: StepLog | None,
-    records: list[dict[str, Any]],
+    records: Iterable[dict[str, Any]],
     *,
     benchmark: str,
     problem_idx: int,
@@ -55,7 +55,7 @@ async def append_model_traffic_records(
     """Append captured model traffic as attempt-scoped step-log rows."""
     if model_traffic_log is None or not records:
         return
-    for traffic_record in format_model_traffic_log_records(
+    for traffic_record in _iter_model_traffic_log_records(
         records,
         benchmark=benchmark,
         problem_idx=problem_idx,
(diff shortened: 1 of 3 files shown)


piojanu:
Reviewed — the disk-spooling design looks solid and reads cleanly. I pushed one small **test-only** commit to the branch (`83b250fa`) closing two gaps in the spool-cleanup coverage:

1. **The only post-drain cleanup assertion could never fail.** `test_model_traffic_spools_and_flushes_session` asserted `not list(spool_dir.rglob("*.jsonl"))`, but `drain_session` renames the spool file to `*.drain` before returning — so that glob never matches the artifact that actually has to be unlinked, and it stayed green even if `discard()` never ran. It now asserts on `*.drain` and that no spool files remain.

2. **The `__del__ -> discard()` cleanup path was untested.** The eval-loop graceful/system-error branches drain the session and drop the return value *without iterating* it, so only `__del__` unlinks the `.drain` file on those paths. Added `test_dropped_drained_session_unlinks_spool_without_iterating` to cover it.

Both assertions were confirmed to fail if `discard()` regresses to a no-op (mutation-checked). `ruff check`/`format --check` clean; `test_model_traffic_capture.py` + `test_interceptors_extended.py` pass.

Otherwise LGTM — I'll approve once this lands.

piojanu (approved):

piojanu:
Pushed one more commit (`98ec6a44`) hardening `DrainedModelTrafficSession` so it fails loud instead of silently returning incomplete data.

A file-backed drained session streams its spool file once and deletes it. Previously, re-iterating an already-consumed session (or accessing one after an interrupted iteration) silently yielded an empty list — and `model_stats()` could return partial counts — because a released `_path` was treated as an empty session. That turns a caller bug into silent, missing traffic in `model_traffic.jsonl`.

Now the session tracks consumption state and raises `DrainedSessionError` from `__iter__`/`_load`/`__len__`/`__getitem__`/`model_stats` when a file-backed session is reused or was interrupted. `__iter__` is kept lazy so a caller-triggered materialization (`list()`/`len()` consulting `__len__` for a length hint) is still observed before falling back to streaming — so a single `list()`/`len()` keeps working, and an empty session is unchanged. The normal eval-loop path (stream once, then read stats) is unaffected.

Tests: added coverage for reuse-after-stream, access-after-interrupted-iteration, and the empty-session case; suites pass.

Still LGTM from me with this in.

PR #1125, opened 2026-07-22
fix(harbor): record Terminus length-error ATIF steps

AWarno:
## Summary
- Preserve successful LiteLLM `finish_reason="length"` call usage/model/reasoning on the raised output-length error.
- Queue the failed Terminus-2 model-call response and inject it as an ATIF agent step before the normal retry step.
- Keep Terminus behavior unchanged: the truncated response is still rejected and retried with the existing output-length prompt.

## Validation
- `uvx ruff check src/nemo_evaluator/solvers/harbor.py tests/test_solvers/test_harbor_terminus_patches.py`
- `python -m py_compile src/nemo_evaluator/solvers/harbor.py tests/test_solvers/test_harbor_terminus_patches.py`
- `git diff --check -- src/nemo_evaluator/solvers/harbor.py tests/test_solvers/test_harbor_terminus_patches.py`
- `uv run pytest tests/test_solvers/test_harbor_terminus_patches.py -q` (`41 passed`)
- `uv run pytest tests/test_solvers/test_harbor_solver.py -q` (`110 passed`)
- Forced small-`max_tokens` smoke: wire total `19`, ATIF per-step total `19`, final metrics total `19`, ordered steps `user -> failed agent -> retry agent`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Evaluation trajectories now consistently include a “failure step” contribution when LLM calls fail during context recovery, summarization, or length/context errors, followed by the eventual retry step.
  * Improved preservation of failure metadata (including available model/reasoning info) and more reliable token/cost accounting to avoid double-counting.
  * Hardened runtime patching so evaluator accounting changes only apply when expected source patterns are present, preventing partial/incorrect instrumentation.
* **Tests**
  * Expanded CLE/Terminus patch coverage for truncation and retry flows, including ordering, metadata handling, and failure-vs-retry accounting accuracy.

### DIFF
--- a/src/nemo_evaluator/solvers/harbor.py
+++ b/src/nemo_evaluator/solvers/harbor.py
@@ -1555,23 +1555,37 @@ def _terminus2_build_local_fallback_llm_content(self) -> str:
     )
 
 
-def _terminus2_save_failed_llm_response(self, chat: Any, error_msg: str, content: str, error: Any) -> None:
+def _terminus2_apply_failed_llm_usage(self, chat: Any, error: Any) -> dict[str, Any] | None:
     usage = getattr(error, "llm_usage", None)
-    metrics = None
-    if usage is not None:
-        chat._cumulative_input_tokens += usage.prompt_tokens
-        chat._cumulative_output_tokens += usage.completion_tokens
-        chat._cumulative_cache_tokens += usage.cache_tokens
-        chat._cumulative_cost += usage.cost_usd
-        totals = (chat.total_input_tokens, chat.total_output_tokens, chat.total_cache_tokens, chat.total_cost)
-        chat._api_token_anchor = (len(chat.messages), totals[0] + totals[1])
+    if usage is None:
+        return None
+
+    chat._cumulative_input_tokens += usage.prompt_tokens
+    chat._cumulative_output_tokens += usage.completion_tokens
+    chat._cumulative_cache_tokens += usage.cache_tokens
+    chat._cumulative_cost += usage.cost_usd
+    chat._api_token_anchor = (
+        len(chat.messages),
+        chat.total_input_tokens + chat.total_output_tokens,
+    )
+    return {
+        "prompt_tokens": usage.prompt_tokens,
+        "completion_tokens": usage.completion_tokens,
+        "cached_tokens": usage.cache_tokens or None,
+        "cost_usd": usage.cost_usd or None,
+    }
+
+
+def _terminus2_save_failed_llm_response(self, chat: Any, error_msg: str, content: str, error: Any) -> None:
+    metrics = self._apply_failed_llm_usage(chat, error)
+    if metrics is not None:
+        totals = (
+            chat.total_input_tokens,
+            chat.total_output_tokens,
+            chat.total_cache_tokens,
+            chat.total_cost,
+        )
         self._failed_llm_step_metric_anchor = totals
-        metrics = {
-            "prompt_tokens": usage.prompt_tokens,
-            "completion_tokens": usage.completion_tokens,
-            "cached_tokens": usage.cache_tokens or None,
-            "cost_usd": usage.cost_usd or None,
-        }
 
     pending = getattr(self, "_pending_failed_llm_response_steps", [])
     pending.append(
@@ -1585,6 +1599,21 @@ def _terminus2_save_failed_llm_response(self, chat: Any, error_msg: str, content
         }
     )
     self._pending_failed_llm_response_steps = pending
+
+
+_TERMINUS_LLM_ERROR_SALVAGE_ANCHOR = (
+    "                if response_path is not None:\n"
+    "                    response_path.write_text(salvaged_response)\n"
+    "\n"
+    "                return salvaged_response\n"
+)
+_TERMINUS_LLM_ERROR_SALVAGE_REPLACEMENT = (
+    "                if response_path is not None:\n"
+    "                    response_path.write_text(salvaged_response)\n"
+    "\n"
+    "                self._apply_failed_llm_usage(chat, e)\n"
+    "                return salvaged_response\n"
+)
 
 
 def _terminus2_append_pending_failed_llm_response_steps(
@@ -1628,8 +1657,47 @@ def _patch_terminus_cle_reset() -> None:
     from harbor.agents.terminus_2 import terminus_2 as terminus2_mod
     from harbor.llms import lite_llm as harbor_litellm
 
-    litellm_src = inspect.getsource(inspect.unwrap(harbor_litellm.LiteLLM.call))
+    def get_source(obj: Any, target: str) -> str:
+        try:
+            return inspect.getsource(inspect.unwrap(obj))
+        except OSError as exc:
+            raise RuntimeError(f"Cannot patch {target}: source is unavailable.") from exc
+
+    def require_patterns(src: str, patterns: tuple[str, ...], target: str) -> None:
+        missing = [pattern for pattern in patterns if pattern not in src]
+        if missing:
+            names = ", ".join(repr(pattern.strip().splitlines()[0]) for pattern in missing)
+            raise RuntimeError(f"Cannot patch {target}: required source pattern(s) missing: {names}.")
+
+    def apply_replacements(src: str, replacements: list[tuple[str, str]], target: str) -> str:
+        for anchor, replacement in replacements:
+            occurrences = src.count(anchor)
+            if occurrences != 1:
+                raise RuntimeError(
+                    f"Cannot patch {target}: expected exactly one match for an anchor "
+                    f"but found {occurrences}. Harbor's source has diverged from the expected source."
+                )
+            src = src.replace(anchor, replacement, 1)
+        return src
+
+    litellm_src = get_source(harbor_litellm.LiteLLM.call, "Harbor LiteLLM.call")
+    responses_src = get_source(harbor_litellm.LiteLLM._call_responses, "Harbor LiteLLM._call_responses")
+    query_src = get_source(terminus2_mod.Terminus2._query_llm, "Terminus2._query_llm")
+    run_loop_src = get_source(terminus2_mod.Terminus2._run_agent_loop, "Terminus2._run_agent_loop")
+
+    litellm_call_fn = None
     if "_EVALUATOR_LENGTH_ERROR_DETAILS" not in litellm_src:
+        require_patterns(
+            litellm_src,
+            (
+                "response = await litellm.acompletion(**completion_kwargs)",
+                "usage_info = self._extract_usage_info(response)",
+                'message = choice["message"]',
+                'content = message.get("content") or ""',
+                'reasoning_content = message.get("reasoning_content")',
+            ),
+            "Harbor LiteLLM.call length-error details",
+        )
         anchor = "                truncated_response=content,\n            )\n            raise exc\n"
         replacement = anchor.replace(
             "            raise exc\n",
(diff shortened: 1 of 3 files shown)


src/nemo_evaluator/solvers/harbor.py
```diff
@@ -1564,33 +1626,74 @@ def _patch_terminus_cle_reset() -> None:
...
-    for anchor, replacement in _TERMINUS_CLE_RESET_REPLACEMENTS:
-        occurrences = src.count(anchor)
-        if occurrences != 1:
-            raise RuntimeError(
-                "Cannot patch Terminus2._query_llm for CLE chat reset: expected exactly "
-                f"one match for an anchor but found {occurrences}. Harbor's terminus_2.py "
-                "has diverged from the expected source."
-            )
-        src = src.replace(anchor, replacement, 1)
+    litellm_src = inspect.getsource(inspect.unwrap(harbor_litellm.LiteLLM.call))
```
piojanu: The `LiteLLM.call` source-patch here only rewrites the **chat-completions** `finish_reason == "length"` branch — the anchor (`truncated_response=content,` / `)` / `raise exc`) matches a single site in `lite_llm.py`. Harbor has a **second** truncation branch on the Responses-API path (`_call_responses`, the `reason == "max_output_tokens"` case) that raises the same `OutputLengthExceededError` but never attaches `llm_usage` / `llm_model_name` / `llm_reasoning_content`.

Both branches funnel into the same `_query_llm` `except OutputLengthExceededError` handler, which now calls `_save_failed_llm_response`. On the Responses-API path `error.llm_usage` is `None`, so the handler skips the cumulative token/cost bump and queues a step with `metrics=None` — silently re-introducing the wire-vs-recorded under-count this change fixes. Unlike the other patch points, nothing raises to flag the gap, and no test covers it.

`use_responses_api` defaults to `False` and isn't set on the current path, so this is latent rather than an active regression. Still worth closing so it can't silently degrade if the Responses API is ever enabled: either apply the same usage attachment to the `_call_responses` truncation branch, or add a fail-loud guard on that path consistent with the `RuntimeError` anchor checks used elsewhere in this patch. A test that raises `OutputLengthExceededError` with no `llm_usage` attached would pin the expected behavior.

src/nemo_evaluator/solvers/harbor.py
```diff
@@ -1564,33 +1626,74 @@ def _patch_terminus_cle_reset() -> None:
...
+            "            # _EVALUATOR_LENGTH_ERROR_DETAILS\n"
+            "            exc.llm_usage = usage_info\n"
+            '            exc.llm_model_name = response.get("model")\n'
+            '            exc.llm_reasoning_content = reasoning_content or message.get("reasoning")\n'
+            "            raise exc\n",
+        )
+        if litellm_src.count(anchor) != 1:
+            raise RuntimeError("Cannot patch Harbor LiteLLM.call length-error details; lite_llm.py has diverged.")
+        namespace: dict[str, Any] = {}
+        exec(textwrap.dedent(litellm_src.replace(anchor, replacement, 1)), harbor_litellm.__dict__, namespace)
```
piojanu: **Partial global-state mutation before the divergence checks makes re-entry fail opaquely**

`_patch_terminus_cle_reset` replaces `harbor_litellm.LiteLLM.call` with an `exec`'d function (lines 1645-1646) and installs the three `Terminus2` helper methods (1648-1655) *before* the `Terminus2._query_llm` / `_run_agent_loop` anchor checks run — and those checks can `raise RuntimeError` via `apply_replacements` (1678, 1686). The `_TERMINUS_CLE_RESET_PATCHED` guard is only set at the very end (1692).

So if Harbor's `terminus_2.py` drifts and an anchor count is not exactly 1:

1. The first `_create_agent` raises the intended clear `"...has diverged"` `RuntimeError` — but `LiteLLM.call` is already swapped and the guard is still `False`.
2. On the next `_create_agent`, line 1631 runs `inspect.getsource(inspect.unwrap(LiteLLM.call))` against the `exec`'d function, whose code object has no backing source file, so it raises `OSError: could not get source code`.

Every task after the first then fails with that opaque `OSError` instead of the clear divergence message — which defeats the fail-loud design exactly when someone is debugging a Harbor version bump.

Suggested fix: do all `inspect.getsource` reads and run every anchor-count/divergence check up front, and only mutate global state (`LiteLLM.call`, the `Terminus2` methods, the `exec`'d `_query_llm`/`_run_agent_loop`) once all checks have passed. Alternatively, wrap the body so any failure restores the originals, or gate re-entry so a partially-applied patch can't be re-run. That keeps a real divergence surfacing the intended `RuntimeError` on every call.

src/nemo_evaluator/solvers/harbor.py
```diff
@@ -1564,33 +1626,74 @@ def _patch_terminus_cle_reset() -> None:
...
+                )
+            src = src.replace(anchor, replacement, 1)
+        return src
+
+    query_src = inspect.getsource(inspect.unwrap(terminus2_mod.Terminus2._query_llm))
+    query_replacements = []
+    if "full_summarize_failed_with_cle" not in query_src:
+        query_replacements.extend(_TERMINUS_CLE_RESET_REPLACEMENTS)
+    if "EVALUATOR_LLM_ERROR_TRAJECTORY_STEP" not in query_src:
+        query_replacements.append((_TERMINUS_LLM_ERROR_STEP_QUERY_ANCHOR, _TERMINUS_LLM_ERROR_STEP_QUERY_REPLACEMENT))
```
piojanu, line 1771: The injected usage-recording hook only covers the retry branch, so salvaged truncated responses go uncounted.

`_TERMINUS_LLM_ERROR_STEP_QUERY_REPLACEMENT` appends `self._save_failed_llm_response(...)` right after the `chat.reset_response_chain()` anchor. In harbor's `_query_llm`, that anchor lives *below* the salvage early-return inside the `OutputLengthExceededError` handler:

```python
if hasattr(self._parser, "salvage_truncated_response"):
    salvaged_response, _ = self._parser.salvage_truncated_response(truncated_response)
if salvaged_response:
    ...
    return salvaged_response          # early return -- hook never reached
# only past here:
chat.messages.append(...); chat.reset_response_chain()
self._save_failed_llm_response(...)  # injected here
```

Because `LiteLLM.call` raises the length exception before returning `usage`, the truncated wire call never updates `chat.total_input_tokens/total_output_tokens`, and the run loop derives per-step tokens from those cumulative deltas. So when `salvage_truncated_response` succeeds, that call's tokens are recorded nowhere — the same under-count this change targets, resurfacing on the salvage path.

Scope: only the XML parser exposes `salvage_truncated_response`; the default `json` parser has no salvage method and always falls through to the instrumented retry branch, so the default configuration is unaffected. If the XML parser is intentionally out of scope, an explicit note would help; otherwise the salvaged call's usage should be recorded too.

One caveat on the fix: on the salvage path the truncated call is the *accepted* response (no retry follows), so reusing `_save_failed_llm_response` as-is is wrong — it queues a separate rejected step and sets an anchor to subtract these tokens from the *next* retry step, which doesn't exist here. Instead, count the call exactly once as the normal agent step, e.g. bump the chat cumulative token/cost counters on the salvage branch (before `return salvaged_response`) so the standard step accounting picks up the delta.

piojanu (commented):
Reviewed the Terminus length-error ATIF accounting — the core fix and tests look correct. Three inline comments on in-scope gaps: two sibling length-error paths the patch doesn't cover (the Responses-API truncation branch and the XML-parser salvage branch), plus a fail-loud/re-entry robustness issue when Harbor's source drifts. Smaller coverage and naming nits left out to keep this focused.

piojanu (changes requested):

src/nemo_evaluator/solvers/harbor.py
```diff
@@ -1564,33 +1626,74 @@ def _patch_terminus_cle_reset() -> None:
...
+        terminus2_mod.Terminus2._query_llm = namespace["_query_llm"]
+    if run_loop_patched:
+        exec(textwrap.dedent(run_loop_src), terminus2_mod.__dict__, namespace)
+        terminus2_mod.Terminus2._run_agent_loop = namespace["_run_agent_loop"]
     _TERMINUS_CLE_RESET_PATCHED = True
-    logger.info("Patched Terminus2._query_llm: reset chat on full-summary CLE + parseable local fallback")
+    logger.info(
+        "Patched Terminus2._query_llm: reset chat on full-summary CLE, "
+        "parseable local fallback, and failed-LLM response ATIF accounting"
+    )
```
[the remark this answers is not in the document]
AWarno: Addressed. `_patch_terminus_cle_reset` now validates the required local-name patterns before rewriting the `LiteLLM.call` length branch: `usage_info`, `response`, `message`, and `reasoning_content`. It also applies the same style of guard to `_call_responses`. The replacements are compiled into local function objects before any Harbor class attribute is mutated, so source drift fails with a clear `RuntimeError` instead of leaving a partially patched runtime.

Coverage: `test_litellm_length_patch_requires_expected_locals_before_mutating`, `test_responses_length_patch_requires_expected_locals_before_mutating`, and `test_terminus_run_loop_diverged_source_raises`.

src/nemo_evaluator/solvers/harbor.py
AWarno: Fixed. The Responses API `reason == "max_output_tokens"` branch now creates the `OutputLengthExceededError`, attaches `llm_usage`, `llm_model_name`, and `llm_reasoning_content`, then raises it. That keeps the shared Terminus2 `OutputLengthExceededError` handler from queuing a metrics-less failed step when `use_responses_api=True`.

Coverage: `test_responses_api_length_error_keeps_usage_details` exercises `_call_responses` directly, and `test_responses_api_length_error_keeps_details_via_call` verifies the public `LiteLLM.call()` path.

Scope note: this is intentionally limited to the affected Evaluator runtime path: Harbor Terminus2 using Harbor `LiteLLM`. It covers both Harbor `LiteLLM` chat-completions and Responses-API length exits. It does not claim to patch unrelated length-error producers such as Harbor `Tinker` or standalone Terminal Bench wrappers; OpenHands uses a separate Harbor adapter path based on OpenHands metrics/raw completion usage rather than this Terminus2 retry handler.

src/nemo_evaluator/solvers/harbor.py
AWarno: Fixed. The patch function now reads all target source first, performs all required pattern checks/replacements, and `exec`s replacement functions into local variables before assigning `LiteLLM` or `Terminus2` methods. The global mutations happen only after those checks pass, so a drift in `_query_llm` or `_run_agent_loop` does not leave `LiteLLM.call` swapped with the guard still unset.

Coverage: the local-name drift tests assert `LiteLLM.call`, `_call_responses`, and `_query_llm` stay unchanged on failure; `test_terminus_run_loop_diverged_source_raises` covers the run-loop drift case.

src/nemo_evaluator/solvers/harbor.py
AWarno: Fixed. The XML salvage early-return path now calls `_apply_failed_llm_usage(chat, e)` before returning the salvaged `LLMResponse`, preserving the successful wire call usage without creating an extra failed-step placeholder. The returned `LLMResponse` also carries the salvaged model/reasoning metadata, so `_run_agent_loop` records it as the normal agent step.

Coverage: `test_salvaged_length_response_counts_usage_without_failed_step` checks the cumulative usage bump, and `test_salvaged_length_response_is_recorded_as_normal_step` checks the ATIF step metrics/model/reasoning path.

AWarno:
Non-inline bot comments are covered too:

- CodeRabbit parametrization nit: I kept the two reasoning tests split deliberately because they assert two distinct contract points: preferred `message.reasoning_content` and fallback to `message.reasoning`. Both are small and directly map to the review questions.
- Codecov: the added behavior is covered by the focused Terminus/report tests; local validation is green: `tests/test_solvers/test_harbor_terminus_patches.py`, `tests/test_reports/test_trajectories_report.py`, ruff, and the full non-network/non-slow/non-e2e suite (`2273 passed, 35 skipped`).

The current PR head is `884399c`, and GitHub checks are green: 34 success, 6 skipped, 0 non-success; CodeRabbit status is success.

piojanu (approved):
Re-reviewed the latest revision — all three points from the earlier review are addressed cleanly, and the suite passes locally (159 passed).

- **Responses-API length branch:** now source-patched in `_call_responses`, so the `max_output_tokens` truncation attaches `llm_usage`/`model`/`reasoning` before raising and flows through the same accounting. Covered by the direct `_call_responses` test and the via-`call` test.
- **Salvage path:** the new `_apply_failed_llm_usage` helper counts the salvaged truncated call once as a normal accepted step (bumps the chat cumulative counters, no rejected-step queue, no metric anchor) and returns a real `LLMResponse` — the right accepted-response semantics. Covered by the `test_salvaged_length_response_*` tests.
- **Partial-patch re-entry:** all `getsource`/divergence checks now run and build the new callables into locals before any global state is mutated, and `get_source` raises a clear `RuntimeError` instead of an opaque `OSError`. The two `*_requires_expected_locals_before_mutating` tests and the diverged-source test assert nothing is mutated on a pattern miss.

Also nice: the end-to-end `generate_trajectories_report` assertions (`all_sources_match`, `problems_with_more_wire_than_steps == 0`) pin the actual reconciliation.

One non-blocking nit (your call): `_apply_failed_llm_usage` still writes the cumulative `total_input_tokens + total_output_tokens` into `chat._api_token_anchor` rather than the single call's `prompt_tokens + completion_tokens` — it only feeds the internal context estimate and is overwritten by the next response, so no metric impact, but the salvage path now exercises it too.

Approving — thanks for the thorough follow-up.

PR #1130, opened 2026-07-27
fix(harbor): harden OpenHands trajectory flush

AWarno:
Hardens OpenHands trajectory preservation for interrupted SWE/Harbor runs.

Changes:
- Convert catchable SIGTERM/SIGINT interruptions in the OpenHands runner into the existing final trajectory serialization path, with nested signals ignored while the file is being written.
- Add periodic structured trajectory checkpoints during `conversation.run()` so SIGKILL/OOM-style hard exits can still leave ATIF instead of only a text log fallback.
- Reuse the runner's existing SDK-event-to-ATIF conversion helper for both final save and checkpoints; no separate ad hoc event serializer.
- Save trajectories atomically through `trajectory.json.partial` -> `trajectory.json` and prefer valid partial JSON before falling back to `openhands_sdk.txt`.

Why:
- In long SWE runs, agent processes can be hard-killed after producing tool/model activity and a workspace diff but before OpenHands writes final ATIF. That leaves rewards/verification output and model traffic present, while the trajectory has zero useful agent steps.
- Catchable signals are handled by final flush. Hard exits are handled by the new checkpoint file already on disk.

Before/after validation:
- Before the checkpoint path, SIGKILL cannot run final cleanup, so no structured trajectory is written if the process is hard-killed between model/tool activity and final save.
- After the checkpoint path, the regression test starts a patched runner, waits for a checkpoint containing an agent step, then sends SIGKILL; recovered `trajectory.json` still contains agent steps and `extra.partial_trajectory.checkpoint=true`.
- SIGTERM/SIGINT still use the final flush path and mark `extra.partial_trajectory.reason="timeout"`.

Validation:
- `.venv/bin/python -m pytest -q tests/test_solvers/test_harbor_solver.py` -> 116 passed.
- `.venv/bin/python -m ruff check src/nemo_evaluator/solvers/harbor.py tests/test_solvers/test_harbor_solver.py` -> passed.
- `.venv/bin/python -m ruff format --check src/nemo_evaluator/solvers/harbor.py tests/test_solvers/test_harbor_solver.py` -> passed.
- GitHub checks for `786a5819` -> 32 success, 6 skipped.
- Internal CI for the matching GitLab MR passed, including the Harbor image build.

### DIFF
--- a/src/nemo_evaluator/solvers/harbor.py
+++ b/src/nemo_evaluator/solvers/harbor.py
@@ -923,13 +923,12 @@ print(f'cmd_timeout_{{_MAX}}s={{ok}} at {{p}}')
 
     # -- Patch 4: flush trajectory when the runner is interrupted ----------
     # Anchor 1: wrap `conversation.send_message()` + `conversation.run()` in
-    # try/except BaseException so main() continues to the existing
+    # try/except BaseException so main() continues to the runner's existing
     # event-reconstruction + build_trajectory + save code on timeout/crash.
-    # SIGTERM/SIGINT are converted to a catchable BaseException so evaluator
-    # timeout cancellation gets the same flush path.  On interruption, first
-    # write a cheap partial trajectory from already-received model events.  The
-    # normal full writer below overwrites it if post-run serialization completes.
-    # Anchor 2: after the final print in main(), exit(0) cleanly so Harbor
+    # SIGTERM/SIGINT are converted to a catchable BaseException, then ignored
+    # while the already-existing post-run trajectory serialization completes.
+    # Anchor 2: add partial-run metadata and make the existing save atomic.
+    # Anchor 3: after the final print in main(), exit(0) cleanly so Harbor
     # treats the run as a normal completion and downloads the trajectory.
     _budget_flush_script = """\
 import sys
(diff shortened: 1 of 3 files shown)


src/nemo_evaluator/solvers/harbor.py
```diff
@@ -990,17 +999,36 @@ async def _patch_openhands_sdk(sandbox: Sandbox, *, cmd_timeout: float | None =
...
+        ind + '            for _event in (getattr(getattr(conversation, "state", None), "events", []) or []):\\n' +
+        ind + '                try:\\n' +
+        ind + '                    if isinstance(_event, MessageEvent):\\n' +
+        ind + '                        if _event.source in ("user", "agent"):\\n' +
+        ind + '                            _entry_type = "assistant_message" if _event.source == "agent" else "user_message"\\n' +
+        ind + '                            _entry = {"type": _entry_type, "content": _trajectory_text_from_event(_event), "timestamp": _event.timestamp}\\n' +
+        ind + '                            _events_list.append(_entry)\\n' +
+        ind + '                    elif isinstance(_event, ActionEvent):\\n' +
+        ind + '                        _events_list.append({"type": "assistant_message", "content": "", "timestamp": _event.timestamp, "llm_response_id": getattr(_event, "llm_response_id", None), "tool_calls": [{"id": _event.tool_call_id, "name": _event.tool_name, "arguments": _trajectory_tool_args(_event)}]})\\n' +
+        ind + '                except BaseException as _event_e:\\n' +
```
AWarno: why thi is added

AWarno: This is defensive per-event handling in the emergency flush path. Before this change, one bad OpenHands event while converting text/tool args could abort the whole partial trajectory save and lose already-converted steps.

With this catch, we keep the usable events, record the conversion problem in `extra.partial_trajectory.flush_error`, and still write a degraded partial ATIF instead of ending with an empty/missing trajectory. `BaseException` is used because this code runs while handling interruption/cancellation paths; the outer save path is also intentionally best-effort.

Covered by `test_flush_patch_saves_degraded_partial_when_event_replay_fails` and `test_flush_patch_ignores_reentrant_signal_while_writing_trajectory`.

src/nemo_evaluator/solvers/harbor.py
```diff
@@ -990,17 +999,36 @@ async def _patch_openhands_sdk(sandbox: Sandbox, *, cmd_timeout: float | None =
...
+        ind + '        except BaseException as _build_e:\\n' +
+        ind + '            _flush_error = f"trajectory build failed: {type(_build_e).__name__}: {_build_e}"\\n' +
+        ind + '            _steps = []\\n' +
+        ind + '            for _idx, _entry in enumerate(_events_list, start=1):\\n' +
+        ind + '                _source = "agent" if _entry.get("type") == "assistant_message" else "user"\\n' +
+        ind + '                _step = {"step_id": _idx, "source": _source, "message": _entry.get("content", "")}\\n' +
+        ind + '                if _entry.get("tool_calls"):\\n' +
+        ind + '                    _step["tool_calls"] = _entry["tool_calls"]\\n' +
+        ind + '                _steps.append(_step)\\n' +
+        ind + '            _trajectory = {"schema_version": "ATIF-v1.6", "session_id": "partial", "agent": {"name": "openhands-sdk", "model_name": model}, "steps": _steps, "final_metrics": {"total_steps": len(_steps), "total_prompt_tokens": _metrics["prompt_tokens"], "total_completion_tokens": _metrics["completion_tokens"]}}\\n' +
```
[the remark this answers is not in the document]
AWarno: Addressed in `2c32fad`: the degraded fallback trajectory was removed, so there is no second schema producer anymore. The recovery fixture was also updated to `ATIF-v1.7`.

src/nemo_evaluator/solvers/harbor.py
AWarno: Updated in `2c32fad`: I removed this per-event partial writer entirely. The patch now lets the runner continue into its existing post-run event conversion / `build_trajectory` / save path, then only annotates that canonical trajectory with `extra.partial_trajectory` when the run was interrupted. This avoids maintaining a second event-to-ATIF conversion path in the patch.

PR #1139, opened 2026-08-10
feat: mirror prebuilt task images into ECR instead of rebuilding them

iguehring-nvidia:
## Problem

On the ECS Fargate sandbox, NEL ignores the prebuilt image a Harbor task declares in `task.toml [environment] docker_image` and rebuilds the task's `environment/Dockerfile` instead. Those Dockerfiles typically install from apt, pip and `curl` without pinning, so a rebuild resolves them against whatever upstream looks like that day — two builds of the same pinned task commit do not produce the same image.

The ECR tag is a content hash of the build context, so the *first* build becomes an accidental pin, stable until the repository's lifecycle policy evicts it. When it does, the next run rebuilds against drifted upstream state and re-caches the broken result under the same hash, so it never recovers.

This is not hypothetical. On a Terminal-Bench 2.1 oracle nightly, three tasks broke in a single night when their images were rebuilt: two had a `curl` of a source file that had 404'd weeks earlier (no `-f`, so `curl` wrote the error page and exited 0), and one pinned an apt version that had since been superseded.

Note the local Docker backend already prefers `docker_image` (`environments/harbor.py::_resolve_image`) — the two backends disagreed.

## Change

Adds `image_mirror_mode` to the ECS Fargate sandbox config:

- **`off`** (default) — build the Dockerfile. Existing behaviour, unchanged.
- **`codebuild`** — copy the declared upstream image into ECR instead, tagged by its **upstream manifest digest**.

Tagging by digest is what makes this worth doing: a cache miss re-mirrors byte-identical content, so eviction costs latency rather than changing what runs. Rebuilding cannot offer that.

An existing mirror is recognised from its tag prefix alone, so the upstream registry is contacted only on a miss, and resolutions are memoised per ref with bounded retry. Sandboxes start once per task, so a naive implementation would issue one registry call per task start from a single egress IP — enough to trip anonymous registry rate limits at realistic concurrency.

Digests resolve over plain HTTPS against the Registry v2 API, so no Docker daemon is required.

Benchmarks that construct images programmatically (SWE-bench and friends) declare no `docker_image`, expose no `source_image`, and keep the build path unchanged.

## Testing

- 25 new unit tests covering tag derivation, digest/canonical ref parsing, cache-hit and cache-miss paths, memoisation under 200 simulated task starts, retry and give-up behaviour, and the mirror-vs-build branch.
- Validated end-to-end on a real 89-task Terminal-Bench 2.1 oracle run: warm images served with zero registry calls, cold images mirrored correctly, and **zero Dockerfile builds**. Two of the regressions described above are fixed by this change; the remainder install from the network during the task's own solve script and need upstream pins instead.

## Rollout

Defaults to `off`, so merging changes nothing. Benchmarks opt in per-config first; playbook defaults follow once proven.

## Summary by CodeRabbit

* **New Features**
  * Added optional image mirroring for ECS Fargate sandboxes, using immutable upstream image digests.
  * Added support for Docker Hub authentication during image mirroring.
  * Pre-built task images can now be used directly without local Dockerfile builds.
  * Added configuration to enable or disable image mirroring.

* **Bug Fixes**
  * Improved image reuse through caching and retry handling.
  * Preserved Dockerfile builds for tasks without pre-built images.

### DIFF
--- a/docs/architecture/aws-ecs-sandbox.mdx
+++ b/docs/architecture/aws-ecs-sandbox.mdx
@@ -92,6 +92,37 @@ Once apply completes:
 
    See `EcsFargateSandbox` in [`src/nemo_evaluator/config/sandboxes.py`](https://github.com/NVIDIA-NeMo/Evaluator/blob/main/src/nemo_evaluator/config/sandboxes.py) for the complete schema.
 
+## How task images reach ECR
+
+Each task needs its image in ECR before Fargate can run it. There are two routes,
+selected by `image_mirror_mode`:
+
+```yaml
+sandbox:
+  type: ecs_fargate
+  region: us-west-2
+  image_mirror_mode: codebuild   # default: off
+```
+
+- **`off`** (default) — build the task's `environment/Dockerfile` via CodeBuild and push
+  the result. The ECR tag is a content hash of the build context.
+- **`codebuild`** — when the benchmark declares an already-published image (Harbor reads
+  `task.toml [environment] docker_image`), copy that image into ECR instead of rebuilding
+  it. The ECR tag is derived from the upstream manifest digest.
+
+Prefer mirroring where it applies. A task Dockerfile typically installs from apt, pip or
+`curl` without pinning, so rebuilding it resolves those references against whatever
+upstream looks like at build time — two builds of the same commit are not guaranteed to
+produce the same image. Because the repository carries a **500-image lifecycle policy
+shared across every benchmark**, a high-volume benchmark can evict another's cached
+images and silently trigger such a rebuild.
+
+Copying is idempotent: the digest-derived tag means an evicted image is restored
+byte-identical, so a cache miss costs latency rather than changing what the task runs.
+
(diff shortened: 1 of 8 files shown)


iguehring-nvidia:
Closing: the benchmarks this targets are being deprecated.

The Harbor-based execution path for the datasets that benefit from this change is being retired in favour of a NeMo Gym-based implementation, so there is no consumer left for the mechanism.

Worth recording for anyone revisiting image handling on the ECS Fargate sandbox:

- The Fargate backend ignores `task.toml [environment] docker_image` and rebuilds `environment/Dockerfile`, while the local Docker backend prefers the declared image — the two backends disagree.
- The ECR tag is a content hash of the build context, so the first build becomes an accidental pin. It survives only until the repository's lifecycle policy evicts it, after which a rebuild silently resolves the Dockerfile's unpinned references against current upstream state and re-caches the result under the same hash.
- Only datasets whose tasks declare `docker_image` were ever affected; others already take the build path unchanged.

PR #1148, opened 2026-08-15
fix: fail loudly instead of silently falling back to exact match in Skills env

### LINKED ISSUE #1141: SkillsEnvironment silently degrades to exact string match when the NeMo Skills evaluator cannot be constructed or raises
**Describe the bug**

`SkillsEnvironment` silently degrades to whole-response exact string matching when the NeMo Skills evaluator is unavailable or raises. Benchmarks whose scoring requires a real evaluator (code execution, unit tests, LLM judge, task-specific graders) therefore produce a complete run of `0.0` rewards rather than an error, and the run still reports as successful.

There are two ways to reach the fallback, and one of them is completely silent.

**Path A — the evaluator is never constructed.** `src/nemo_evaluator/environments/skills.py:101`:

```python
def _get_evaluator(eval_type: str):
    try:
        from nemo_skills.evaluation.evaluator import get_evaluator_class

        return get_evaluator_class(eval_type, config={})
    except (ImportError, KeyError, ValueError, TypeError):
        return None
```

The evaluator is constructed with an empty `config={}`, so any evaluator that requires configuration (judge endpoint, sandbox, credentials) fails to build. The exception is swallowed and `None` is returned. `__init__` (`skills.py:145-148`) logs the benchmark and `eval_type` but not whether an evaluator was obtained, so this leaves no trace at all.

**Path B — the evaluator raises at scoring time.** `skills.py:209`:

```python
def _score_via_evaluator(self, response, expected, meta):
    try:
        sample = {"predicted_answer": response, "expected_answer": expected}
        result = self._evaluator.eval_single(sample)
        return float(result.get("is_correct", 0)), {...}
    except Exception as e:
        logger.warning("Skills evaluator failed: %s", e)
        return self._score_exact(response, expected)
```

Note also that `predicted_answer` receives the raw, unextracted model response, which may itself be what causes `eval_single` to raise for evaluators that expect an extracted answer field.

Both paths land here (`skills.py:217`):

```python
def _score_exact(self, response, expected):
    correct = response.strip().lower() == expected.strip().lower()
    return (1.0 if correct else 0.0, {"method": "exact_match_fallback", "match": correct})
```

Only `math` and `multichoice` have dedicated handlers (`_score_handlers`, `skills.py:222`); every other `eval_type` depends on the evaluator and is therefore exposed to this path.

**Steps/Code to reproduce bug**

No model or GPU required.

1. Select a `skills://` benchmark whose `METRICS_TYPE` is neither `math` nor `multichoice` and whose evaluator requires configuration — for example a code-execution or judge-scored benchmark.
2. `_get_evaluator` returns `None` because `config={}` cannot satisfy the evaluator's requirements, and nothing is logged.
3. Every sample is then scored by `response.strip().lower() == expected.strip().lower()`.

For a code-generation benchmark this means a functionally correct solution scores `0.0` unless the response text happens to be byte-identical to the reference, which it essentially never is. The run completes successfully and reports a near-zero score.

**Expected behavior**

When the configured scoring protocol is unavailable, the environment should fail explicitly rather than substitute a different scoring method. Exact string matching is not a degraded form of code execution or judge-based grading — it measures something else entirely, and silently substituting it turns an infrastructure failure into a plausible-looking result.

Concretely:

* `_get_evaluator` returning `None` for an `eval_type` that has no dedicated handler should be a hard error at environment construction, not a silent `None`.
* An evaluator raising during scoring should fail the benchmark (or at minimum mark the affected samples as unscored), rather than being rescored by a different method under `logger.warning`.
* If a lenient mode is wanted, it should be opt-in and clearly reflected in the reported result, not the default.

The fallback is recorded per record as `scoring_details["method"] == "exact_match_fallback"`, so the substitution is technically observable — but nothing consumes that field, it does not affect the aggregate, and the run's exit status is unchanged.

**Additional context**

For context on the use case: we are an NVIDIA-internal team running `nemo-evaluator` in an automated accuracy-regression pipeline that compares quantized checkpoints against their unquantized baselines.

This failure mode is particularly damaging for paired comparisons. If both sides of an A/B comparison degrade to exact match, both score approximately `0.0`, the delta is `0.0`, and the comparison reports "no regression" — when in fact the benchmark was never evaluated. A hard failure would be strictly preferable to a result that is indistinguishable from a passing one.

This is the reason we currently cannot qualify judge-scored or execution-scored `skills://` benchmarks for that pipeline.

Related but filed separately: NVIDIA-NeMo/Evaluator#1140 covers `vlmevalkit://` MCQ scoring not preserving VLMEvalKit's judge-based answer extraction.

**Version**

`main` @ `4d081325` (`version = "0.4.0"`). Also present in `0.3.0`.

arthi-arumugam-git:
Fixes #1141, per the Expected behavior in the issue.

Both silent-degradation paths now fail loudly by default:

- **Construction (path A):** `_get_evaluator` logs the previously swallowed exception, and `__init__` raises `SkillsEvaluatorError` when an evaluator-dependent `eval_type` (no dedicated handler) gets `None` back. The startup log line now also states whether an evaluator was constructed.
- **Scoring time (path B):** an evaluator exception raises `SkillsEvaluatorError` with the original exception chained, instead of rescoring by a different method under a warning.
- **Lenient mode is explicit opt-in:** `allow_exact_match_fallback=True` restores the old behavior, is plumbed through the `skills://` factory, logs loudly when active, and the substituted records still carry `method == "exact_match_fallback"`.

`SkillsEvaluatorError` deliberately subclasses `RuntimeError` rather than `GracefulError`/`InfraError`: both of those end in `reward=0.0`, which is exactly the indistinguishable-from-a-real-score outcome the issue is about.

Tests (`tests/test_environments/test_skills_evaluator_fallback.py`) run with no GPU, network, or nemo_skills install: construction helpers are patched for the gate tests, scoring-time behavior is tested on a directly assembled instance, and the working-evaluator path is covered to show no behavior change there. 7 pass with the fix; the construction gate and raise-on-failure cases fail on current main. Pre-existing environment-dependent failures in `test_environments_extended.py` are identical with and without this change.

Not addressed here, flagged in the issue: `predicted_answer` receives the raw unextracted response, which may itself be why some evaluators raise. That looks like a separate decision about extraction, happy to follow up if wanted.

## Summary by CodeRabbit

* **Bug Fixes**
  * Skills evaluation now reports clear errors when the configured evaluator is unavailable or fails.
  * Exact-match fallback is no longer used automatically, preventing potentially misleading evaluation results.

* **New Features**
  * Exact-match fallback can be explicitly enabled when desired.
  * Evaluator status and construction issues now provide warning details for easier diagnosis.
  * Evaluation can use configured handlers and successfully report evaluator-based scores.

### DIFF
--- a/src/nemo_evaluator/environments/registry.py
+++ b/src/nemo_evaluator/environments/registry.py
@@ -131,6 +131,7 @@ def _make_skills(rest: str, **kwargs: Any) -> "EvalEnvironment":
         data_dir=kwargs.get("data_dir"),
         prompt_template=kwargs.get("prompt_template"),
         eval_type=kwargs.get("eval_type"),
+        allow_exact_match_fallback=kwargs.get("allow_exact_match_fallback", False),
     )
 
 
--- a/src/nemo_evaluator/environments/skills.py
+++ b/src/nemo_evaluator/environments/skills.py
@@ -98,12 +98,28 @@ def _load_dataset(benchmark: str, split: str | None, data_dir: str | None) -> li
     return samples
 
 
+class SkillsEvaluatorError(RuntimeError):
+    """The configured NeMo Skills evaluator is unavailable or failed.
+
+    Raised instead of silently falling back to whole-response exact string
+    matching, which measures something else entirely and turns an
+    infrastructure failure into a plausible-looking score. Pass
+    ``allow_exact_match_fallback=True`` to opt in to the old behavior.
+    """
+
+
 def _get_evaluator(eval_type: str):
     try:
         from nemo_skills.evaluation.evaluator import get_evaluator_class
 
         return get_evaluator_class(eval_type, config={})
-    except (ImportError, KeyError, ValueError, TypeError):
+    except (ImportError, KeyError, ValueError, TypeError) as e:
+        logger.warning(
+            "NeMo Skills evaluator for eval_type=%r could not be constructed: %s: %s",
+            eval_type,
+            type(e).__name__,
+            e,
+        )
         return None
 
 
@@ -131,6 +147,7 @@ class SkillsEnvironment(EvalEnvironment):
         data_dir: str | None = None,
         prompt_template: str | None = None,
         eval_type: str | None = None,
+        allow_exact_match_fallback: bool = False,
     ) -> None:
         super().__init__()
         self._benchmark = benchmark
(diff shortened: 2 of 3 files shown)


arthi-arumugam-git:
Thanks, and agreed on the reasoning: tests that build `SkillsEnvironment` directly would not catch a regression in the `skills://` factory default or in argument forwarding, which is exactly where a silent fallback would come back.

That coverage went in as e49fcca, a few minutes after this review ran against bd1d697. `TestFactoryForwarding` calls `registry._make_skills("gsm8k", ...)` with the constructor patched and asserts on the captured kwargs, offline and parametrized the way you suggested:

- `test_default_is_false` pins the default at `False`, so the fallback cannot become opt-out by accident
- `test_explicit_optin_is_forwarded` pins that an explicit `True` actually reaches the constructor rather than being dropped on the way through

Happy to extend it to the other factories in the registry if the same forwarding risk applies to them.

arthi-arumugam-git:
@chtruong814 @ko3n1g would one of you be able to take a look at this one when you have a moment? It has been sitting since 15 Aug with the copy-pr-bot gate still closed, so no workflow has run on it yet.

Small change, one behaviour: when the configured scorer cannot be resolved the Skills environment now raises instead of quietly scoring with exact match, and the test covers both the factory default and an explicit argument. Happy to rebase or adjust if anything about it does not fit how you want failures surfaced.

PR #1161, opened 2026-09-22
fix(slurm): capture the evaluation exit code before server teardown

wprazuch:
## Problem

The generated Slurm batch script captures the evaluation client's exit status with
`EVAL_EXIT_CODE=$?`, but that statement is emitted *after* the deployment teardown,
not after the evaluation.

Order in the generated `run.sub`, when a self-hosted deployment is configured:

```sh
srun ... bash -c '<evaluation client>'          # the thing we want the status of
for _pid in "${SERVER_PIDS[@]}"; do kill "$_pid" 2>/dev/null || true; done
EVAL_EXIT_CODE=$?                                # reads the kill loop
```

`$?` reflects only the most recent command. The teardown loop ends in `|| true`, so it
always exits 0. `EVAL_EXIT_CODE` is therefore deterministically 0 whenever a deployment
is present, regardless of how the evaluation actually ended.

**Impact:** every Slurm evaluation with a self-hosted deployment reports success, and
auto-export runs on failed results. Observed on a real run where the evaluation step
exited non-zero while the script printed "Evaluation completed successfully. Starting
auto-export...".

This is long-standing rather than a recent regression; the same shape is present in
older released launcher versions. The array teardown loop with `|| true` made it
deterministic, where a singular `kill` previously leaked a failure through occasionally.

## Fix

Capture the status immediately after the evaluation `srun`, and pass the variable name
into `_generate_auto_export_section`.

This cannot be fixed inside that function: the capture is the first statement it emits,
so it is always at the mercy of whatever the caller emitted last. The ordering
constraint belongs to the caller, so the capture moves there.

## Tests

The regression test asserts on the **fully generated script**, not on a section in
isolation: it requires that no statement appears between the evaluation `srun` and the
`$?` capture, that the teardown comes after the capture, and that exactly one `$?`
capture exists in the whole script. It is parametrised over the single-instance and
multi-instance (proxy) shapes.

The existing section-level test passes against the buggy script, because the caller's
ordering is structurally invisible to it — which is why this defect survived. A
companion test now also pins that the auto-export section must not re-read `$?` itself.

Verified that all three new tests fail against the unfixed source and pass with the fix.
Full suite: 172 passed.


packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
```diff
@@ -81,6 +81,11 @@
 )
 from nemo_evaluator_launcher.executors.registry import register_executor
 
+# Shell variable in the generated batch script that holds the evaluation
+# client's exit status. It is assigned immediately after the client `srun` and
+# read later on; nothing between the two may clobber `$?`.
```
iguehring-nvidia: nit: I believe all three lines of comments are unnecessary.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
```diff
@@ -1057,6 +1062,13 @@ def _create_slurm_sbatch_script(
     s += eval_factory_command
     s += "'\n\n"
 
+    # Capture the evaluation client's status here and nowhere else. `$?` reflects
+    # only the most recent command, and everything emitted below (server
+    # teardown, auxiliary teardown) would overwrite it -- the teardown ends in
+    # `|| true`, so a later read would deterministically see 0 and report a
+    # failed evaluation as a success.
+    s += f"{EVAL_EXIT_CODE_VAR}=$?\n\n"
```
iguehring-nvidia, line 1064: nit: maybe also less verbose comments here?

iguehring-nvidia (approved):

wprazuch:
CI notes for the first run:

- `unit-tests-launcher` (the job covering this change): **1111 passed, 2 skipped**, including the three new tests.
- `secrets-detector`: the new test shifted three already-baselined false positives in `test_slurm_executor.py` by 92 lines. Refreshed the baseline line numbers in a follow-up commit; no entries added or removed.
- `unit-tests`: fails on `test_vlm_endpoint_type_deprecation_removal_reminder` in the `nemo-evaluator` package. That is a deliberate time-bomb reminder asserting `datetime.now() < datetime(2026, 6, 6)`, so it has failed on every PR to this branch since that date. Unrelated to this change and not touched here.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
wprazuch: Done in 3584354b — dropped the constant's comment entirely and cut the capture note to the one fact that makes its position load-bearing.

packages/nemo-evaluator-launcher/src/nemo_evaluator_launcher/executors/slurm/executor.py
wprazuch: Done in 3584354b — dropped the constant's comment entirely and cut the capture note to the one fact that makes its position load-bearing.

Glorf:
oh boy, we fixed this one in NEL 1 yeat ago :D

Glorf (approved):

agronskiy (approved):

